PR #14, opened 2021-08-11
starter: correctly resolve system color mode

vojtaholik:
color-mode-toggle@2x


DrShpongle (approved):

apps/next-product-starter/src/components/color-mode-toggle.tsx
```diff
@@ -3,7 +3,7 @@ import {useTheme} from 'next-themes'
 
 const DarkModeToggle: React.FC = () => {
   const [mounted, setMounted] = React.useState(false)
-  const {theme, setTheme} = useTheme()
+  const {resolvedTheme: theme, setTheme} = useTheme()
```
vojtaholik, line 6: > If `enableSystem` is true and the active theme is "system", this returns whether the system preference resolved to "dark" or "light". Otherwise, identical to `theme`

PR #26, opened 2021-08-12
 import rust adventure

### LINKED ISSUE #22: 🦀 import rustadventure into the monorepo
Rust Adventure by Chris Biscardi is currently living in a private repo. We want to migrate the project into the monorepo so it can leverage the shared code environment.

joelhooks:
closes #22


apps/rustadventure/src/@types/mdx.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.mdx' {
```
joelhooks, line 1: this file was the source of many tears to narrow down 😭

PR #27, opened 2021-08-12
just a test

joelhooks:


package.json
```diff
@@ -44,6 +44,11 @@
       "build": {
         "dependsOn": [
           "^build"
+        ],
+        "outputs": [
+          "dist/**",
+          ".next/**",
+          "!.next/cache/**/*"
```
joelhooks, line 51: this was tricky to figure out, but if packages/apps have special output folders they need to be listed

PR #33, opened 2021-08-13
Jh/31 extract quiz related stuff

### LINKED ISSUE #31: extract email quiz functionality into its own package
### Discussed in https://github.com/skillrecordings/turbo-playground/discussions/15

<div type='discussions-op-text'>

<sup>Originally posted by **vojtaholik** August 11, 2021</sup>
Would it make sense to start new package `@skillrecordings/quiz` for all things quizzes/surveys?

I would probably start by extracting core email quiz mechanism first, specifically [useQuizQuestion](https://github.com/skillrecordings/turbo-playground/blob/main/apps/next-product-starter/src/hooks/use-quiz-question.ts) hook. 

Other bits are calling convertkit utils which could likely be extracted into `@skillrecordings/convertkit`. 

Not sure what's the best approach for handling UI components yet.

```
// related files

components/forms/quiz
  - essay-question.tsx
  - multiple-choice-question.tsx
  - completed-message.tsx
  - submit.tsx
pages/api
  - answer.ts
pages
  - answer.tsx
hooks
  - use-quiz-question.ts
utils
  - fetch-convertkit-subscriber.ts
  - axios-convertkit-api.ts
```
</div>

joelhooks:
closes #31


apps/next-product-starter/src/pages/answer.tsx
```diff
@@ -1,147 +1,62 @@
...
-    switch (currentQuestion.type as string) {
-      case 'multiple-choice':
-        return <MultipleChoiceQuestion question={currentQuestion as Question} />
-      default:
-        return <EssayQuestion question={currentQuestion as Question} />
-    }
+import {QuizAnswerPage} from '@skillrecordings/quiz'
+import {Questions} from '@skillrecordings/types'
+
+export async function getStaticProps() {
```
joelhooks, line 6: this is a "big change" in that the page component expects the quiz data as a prop instead of as a global/const or whatever. In this case I'm just dumping static data in, but we could back this with a data source which would be rad af.

PR #40, opened 2021-08-15
[#39] migrates composable commerce components to commerce package

joelhooks:
also adds an /api/prices endpoint to replace the direct call to the
egghead api which will let end users see a "local" call instead
of the call to egghead server

ran into an issue with the reachui dialog css after moving commerce
component to the package that required the css inport be moved
to the _app component of the app and out of the commerce component
package or a build error occurred.


apps/next-product-starter/src/pages/_app.tsx
```diff
@@ -8,6 +8,9 @@ import 'focus-visible'
 import {ViewerProvider} from '@skillrecordings/viewer'
 import {ConvertkitProvider} from '@skillrecordings/convertkit'
 
+//TODO: build error if this was in the component after moving it to commerce package
```
joelhooks, line 11: This is annoying.

apps/next-product-starter/src/pages/buy.tsx
```diff
@@ -17,7 +28,23 @@ const Buy: FunctionComponent<BuyProps> = ({bundles}) => {
   return (
     <Layout meta={{title: `Buy ${config.defaultTitle}`}}>
       <ClaimCoupon sellable={sellable} />
-      <Commerce bundles={bundles} />
```
joelhooks, line 20: instead of a single `Commerce` component here I broke it down into part, similar to the work we've poked at with `Card` for egghead next.

The result is more composable, so technically we can hot swap aspects of the checkout widget more readily and should make conversations around customizations easier to have.

This was a quick take on this, and I think it could use more thought and attention.

packages/commerce/src/api/prices.ts
```diff
@@ -0,0 +1,45 @@
+import {NextApiRequest, NextApiResponse} from 'next'
```
joelhooks, line 1: new common endpoint for loading prices from the egghead api

packages/commerce/src/components/purchase-bundle.tsx
```diff
@@ -228,6 +228,18 @@ const PurchaseBundle = ({
...
+    if (isString(error)) {
+      return error
+    } else if (error?.message) {
+      return error?.message
+    } else if (error) {
+      return 'An error has occurred'
+    }
+  }
+
+  console.log(state.context?.error)
```
joelhooks, line 241:
```suggestion

```

packages/commerce/src/machines/commerce/index.ts
```diff
@@ -299,7 +296,7 @@ const commerceMachine = createMachine<CommerceMachineContext, CommerceEvent>(
         appliedCoupon: (context, _) => {
           const quantity = get(context, 'quantity')
           if (quantity !== 1) {
-            return null
+            return
```
joelhooks, line 299: returning `null` was causing much more complicated types downstream because `null` is an object/value so setting a type like `mything?: string` doesn't work (the `?` says it can be `undefined` but doesn't cover `null`) so I went through and eliminated the null from the typing and returns in the machine

packages/commerce/src/machines/commerce/utils.ts
```diff
@@ -126,24 +101,28 @@ export const getPriceParams = (machineContext: CommerceMachineContext) => {
   }
   const {site, id: sellable_id, type} = sellable
 
+  const upgradeParams = {
+    upgrade_from_sellable_id: upgradeFromSellable?.slug,
+    upgrade_from_sellable: upgradeFromSellable?.type,
+  }
+
+  const sellableStuff: EggheadSellableParam = {
```
joelhooks, line 109: added some types here and left out the `pickBy` in favor of "assembling" the params with spreads.

packages/commerce/src/machines/commerce/utils.ts
```diff
@@ -126,24 +101,28 @@ export const getPriceParams = (machineContext: CommerceMachineContext) => {
...
+        sellables: [sellableStuff],
         site: process.env.NEXT_PUBLIC_SITE_NAME || 'TEST_PRODUCT',
         code: appliedCoupon,
-      })
+      }
     : {id: stripePriceId}
 }
 
-export const eggheadPriceCheckUrl = `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/api/v1/sellable_purchases/prices`
+export const eggheadPriceCheckUrl = `/api/prices`
```
joelhooks, line 125: local call for prices that is a proxy for this endpoint

generally, I want to migrate to this pattern for these projects where it makes sense.

PR #43, opened 2021-08-16
compile css to import

joelhooks:
compile css uncompressed and unminified in package projects that use tailwind and import that into apps that need to use that CSS to prevent purging


packages/commerce/package.json
```diff
@@ -8,12 +8,14 @@
...
-    "dist/**"
+    "dist/**",
+    "dist/css/commerce.css"
   ],
   "scripts": {
     "test": "jest",
     "dev": "etsc",
-    "build": "tsc --jsx react-jsx --outDir dist",
+    "build": "tsc --jsx react-jsx --outDir dist && pnpm build:css",
+    "build:css": "NODE_ENV=production postcss ./src/styles/commerce.css -o ./dist/css/commerce.css",
```
joelhooks, line 18: straight forward build step here to get the css

packages/commerce/tailwind.config.js
```diff
@@ -0,0 +1,48 @@
...
+        ...colors,
+      },
+      screens: {
+        print: {raw: 'print'},
+      },
+      typography: (theme) => ({
+        DEFAULT: {
+          css: {
+            a: {
+              color: theme('colors.rose.600'),
```
joelhooks, line 22: I cut and pasted this config but it feels like we could have a shared config where the theme colors are the only thing that really change for us.

apps/next-product-starter/src/pages/_app.tsx
```diff
@@ -10,6 +10,7 @@ import {ConvertkitProvider} from '@skillrecordings/convertkit'
 
 //TODO: build error if this was in the component after moving it to commerce package
 import '@reach/dialog/styles.css'
+import '@skillrecordings/commerce/dist/css/commerce.css'
```
joelhooks, line 13: curious if this will cause subtle cascade issues 🤔

joelhooks: I think this might end up being a theming disaster 😭

apps/next-product-starter/tailwind.config.js
```diff
@@ -5,7 +5,12 @@ module.exports = {
   mode: 'jit',
   darkMode: 'class',
   purge: {
-    content: ['./src/**/*.tsx', './src/**/*.mdx'],
+    content: [
+      './src/**/*.tsx',
+      './src/**/*.mdx',
+      '../../packages/commerce/src/**/*.tsx',
+      '../../packages/react/src/**/*.tsx',
```
joelhooks, line 12: technically peaking we can just point to these packages and call it a day?

packages/commerce/package.json
jaredpalmer: why do you want distribute all of tailwind here? Isn't commerce only going to be effectively used by next.js projects (i.e. apps/*)? There are no custom styles so this is unecessary I think

joelhooks: it's purged so it only includes what is used, but I have no idea what I'm doing and there are no examples so I'm just guessing 

the bit where I just point purge at these source trees is probably better for our use case, but won't work if we distribute these libraries

PR #47, opened 2021-08-18
sets up testing for commerce and react packages

joelhooks:
Adds testing with jest and RTL to `commerce` and `react` components with a couple initial tests. Also went through and pulled out a little common config from the tailwind configuration. Left the stuff that was variable, but the purge format is the same.

Just used the next testing stuff since it works and the packages were importing next anyway.


apps/engmanagement/src/__tests__/testing-library.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
```
joelhooks, line 1: these are just nonsense "does it work" tests

PR #48, opened 2021-08-19
manipulate the configs

joelhooks:


apps/next-product-starter/src/pages/buy.tsx
```diff
@@ -33,10 +37,19 @@ const Buy: FunctionComponent<BuyProps> = ({bundles}) => {
         <CommerceBundles>
           {bundles.map((bundle) => {
             return (
-              <CommerceBundleDetails bundle={bundle}>
-                <CommerceBundleImage bundle={bundle} />
+              <CommerceBundleDetails
+                getBundleStyles={getBundleStyles}
```
joelhooks, line 41: these "getter" functions are specific to the product, but they all need them so we will pass them in as callbacks

packages/scripts/next.config.js
```diff
@@ -0,0 +1,46 @@
+const withPlugins = require('next-compose-plugins')
```
joelhooks, line 1: this config was basically the same for all apps

PR #49, opened 2021-08-19
add tests to auth flow

joelhooks:
This was a challenge and required a lot of trial and error to get the oauth flow functioning in a sensible way.

homer spins a rubiks cube


packages/auth/package.json
```diff
@@ -22,19 +22,33 @@
...
+    "resetMocks": false,
+    "setupFiles": [
+      "jest-localstorage-mock"
+    ],
+    "passWithNoTests": true,
+    "testEnvironment": "jsdom",
+    "setupFilesAfterEnv": [
+      "<rootDir>/setupTests.js"
+    ],
+    "testURL": "http://localhost:3000"
```
joelhooks, line 51: the `tesURL` **absolutely needs to be set** or you can't change the URL later. Once it is set, you can use `window.history.pushState` to change the current `href`. This was a whole thing as far as I can tell, with a lot of sad times and verbose solutions.

here's the comment that did the trick:

https://github.com/facebook/jest/issues/890#issuecomment-506926841

if you see a better solution, please let me know 😅

packages/auth/setupTests.js
```diff
@@ -0,0 +1,29 @@
...
+const listeners = window._virtualConsole.listeners('jsdomError')
+const originalListener = listeners && listeners[0]
+
+window._virtualConsole.removeAllListeners('jsdomError')
+
+// Add a new listener to swallow JSDOM errors that orginate from clicks on anchor tags.
+window._virtualConsole.addListener('jsdomError', (error) => {
+  if (
+    error.type !== 'not implemented' &&
+    error.message !== 'Not implemented: navigation (except hash changes)' &&
```
joelhooks, line 22: this was a pretty common issue and just noise in the console

jsdom is very much NOT trying to be a browser which is probably healthy. I haven't scratched below the surface on what jsdom is actually doing so this is all a guess.

packages/auth/src/__tests__/auth.test.ts
```diff
@@ -1,6 +1,91 @@
...
-test('can create an instance', () => {
+import Auth, {getAuthorizationHeader, getLocalUser} from '../index'
+
+export const cornDogUser = {
+  name: 'Corn Dog',
+}
+
+export const server = setupServer(
+  // we can fake the actual user tat gets loaded with msw 🎉
+  rest.get(`/api/users/current`, (req, res, ctx) => {
```
joelhooks, line 14: [msw](https://mswjs.io/docs/) is rad and makes mocking end points a breeze

amaze!

packages/auth/src/__tests__/auth.test.ts
```diff
@@ -1,6 +1,91 @@
...
+test('can create an instance of Auth', () => {
   const auth = new Auth()
   expect(auth).not.toBeNull()
 })
+
+test('read the token from auth instance', async () => {
+  const auth = new Auth()
+  const accessToken = '0f95fab36e7a'
+
+  const token = await auth.eggheadAuth.token.getToken(
```
joelhooks, line 36: this is exercising the [client-oauth2](https://www.npmjs.com/package/client-oauth2) token a bit while verifying that the `Auth` instance is initializing the client properly and it can be accessed

packages/auth/src/__tests__/auth.test.ts
```diff
@@ -1,6 +1,91 @@
...
+  const accessToken = '0f95fab36e7a'
+  const expiresIn = 5184000
+  const url = `${process.env.NEXT_PUBLIC_REDIRECT_URI}#access_token=${accessToken}&token_type=bearer&expires_in=${expiresIn}`
+
+  // set the testURL in package.json to the redirect DOMAIN, then we can
+  // push state and get a new URL and simulate a valid login
+  //https://github.com/facebook/jest/issues/890#issuecomment-506926841
+  window.history.pushState({}, '', url)
+
+  const viewer = await auth.handleAuthentication()
```
joelhooks, line 54: `handleAuthentication` is a cascade that does quite a lot internally and respond with a `Viewer` (user) instance. As long as what it does is return a `Viewer` we are ignoring all that internal complexity. Since it digs in and makes the API call to return the user, I'm satisfied.

packages/auth/src/__tests__/auth.test.ts
```diff
@@ -1,6 +1,98 @@
...
+
+  expect(viewer).toEqual(cornDogUser)
+})
+
+test('sends the proper Authorization header', () => {
```
joelhooks, line 59: A failing test for tomorrow ❌

packages/auth/src/__tests__/auth.test.ts
```diff
@@ -1,6 +1,98 @@
...
+  // since this test is using a mock service, we want to verify that the
+  // request is structured as we expect
+  // TODO: Figure out how to intercept and analyze the mock api request with msw
+  fail()
+})
```
joelhooks, line 64: a failing test for tomorrow ❌

PR #51, opened 2021-08-23
subscribe form: separate styles

vojtaholik:

### DIFF
--- a/apps/compilersforhumans/src/pages/index.tsx
+++ b/apps/compilersforhumans/src/pages/index.tsx
@@ -36,7 +36,19 @@ export default function Home() {
           <h3 className="font-bold lg:text-7xl sm:text-5xl text-4xl pb-8 text-center">
             Learn about Compilers
           </h3>
-          <ConvertkitSubscribeForm />
+          <ConvertkitSubscribeForm
+            classNames={{
+              input:
+                'autofill:text-fill-black autofill:caret-black dark:bg-black w-full text-lg py-2 px-4 leading-7 border-gray-200 rounded-md focus:ring-violet-400 focus:ring-2 focus:outline-none focus:border-transparent placeholder-gray-400',
+              form: 'max-w-sm w-full mx-auto py-8 space-y-5',
+              label: 'block text-base pb-1',
+              asterisk: 'dark:text-fuchsia-400',
+              button:
+                'font-medium text-lg rounded-full focus:bg-gray-500 focus:ring-2 focus:scale-90 focus:ring-black hover:scale-105 transition-all hover:shadow-xl focus:outline-none mt-4 sm:px-16 px-14 py-4 bg-gradient-to-r dark:from-pink-500 dark:to-purple-500 from-violet-500 to-pink-500 text-white ease-in-out',
+              comment: 'text-center pt-4 text-base opacity-70',
+              buttonContainer: 'w-full flex items-center justify-center',
+            }}
+          />
         </div>
         <div
           aria-hidden="true"
--- a/apps/compilersforhumans/src/pages/index.tsx
+++ b/apps/compilersforhumans/src/pages/index.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react'
-//TODO: Check form styling
 import ConvertkitSubscribeForm from '@skillrecordings/convertkit/dist/forms'
 import Layout from 'layouts'
 import LandingCopy from 'components/landing-copy.mdx'
--- a/apps/engmanagement/src/pages/index.tsx
+++ b/apps/engmanagement/src/pages/index.tsx
@@ -78,7 +78,19 @@ export default function Home() {
             </div>
             <div className="uppercase">Subscribe to get notified</div>
           </h2>
-          <ConvertkitSubscribeForm />
+          <ConvertkitSubscribeForm
+            classNames={{
+              input:
+                'w-full bg-gray-900 sm:text-xl text-lg px-5 py-4 rounded-lg text-white placeholder-gray-300 border-gray-700 shadow-inner focus:ring-0 focus:border-orange-300',
+              label: 'sm:text-lg text-base block pb-2',
+              form: 'space-y-6 w-full max-w-md mx-auto',
+              asterisk: 'text-orange-300',
+              button:
+                'sm:px-20 px-16 py-5 text-xl font-semibold bg-gradient-to-t from-orange-500 via-orange-400 to-orange-300 text-white rounded-lg transform hover:shadow-xl hover:scale-105 focus:scale-95 transition-all ease-in-out duration-200 focus:outline-none outline-none focus:ring-2 focus:ring-orange-200',
+              buttonContainer: 'w-full flex items-center justify-center py-8',
+              comment: 'text-lg text-center text-gray-200',
+            }}
+          />
         </div>
       </section>
     </Layout>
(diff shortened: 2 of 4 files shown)


apps/compilersforhumans/src/pages/index.tsx
```diff
@@ -36,7 +36,19 @@ export default function Home() {
...
+                'autofill:text-fill-black autofill:caret-black dark:bg-black w-full text-lg py-2 px-4 leading-7 border-gray-200 rounded-md focus:ring-violet-400 focus:ring-2 focus:outline-none focus:border-transparent placeholder-gray-400',
+              form: 'max-w-sm w-full mx-auto py-8 space-y-5',
+              label: 'block text-base pb-1',
+              asterisk: 'dark:text-fuchsia-400',
+              button:
+                'font-medium text-lg rounded-full focus:bg-gray-500 focus:ring-2 focus:scale-90 focus:ring-black hover:scale-105 transition-all hover:shadow-xl focus:outline-none mt-4 sm:px-16 px-14 py-4 bg-gradient-to-r dark:from-pink-500 dark:to-purple-500 from-violet-500 to-pink-500 text-white ease-in-out',
+              comment: 'text-center pt-4 text-base opacity-70',
+              buttonContainer: 'w-full flex items-center justify-center',
+            }}
+          />
```
vojtaholik, line 51: fixed styles for compilersforhumans product.

apps/engmanagement/src/pages/index.tsx
```diff
@@ -78,7 +78,19 @@ export default function Home() {
...
+                'w-full bg-gray-900 sm:text-xl text-lg px-5 py-4 rounded-lg text-white placeholder-gray-300 border-gray-700 shadow-inner focus:ring-0 focus:border-orange-300',
+              label: 'sm:text-lg text-base block pb-2',
+              form: 'space-y-6 w-full max-w-md mx-auto',
+              asterisk: 'text-orange-300',
+              button:
+                'sm:px-20 px-16 py-5 text-xl font-semibold bg-gradient-to-t from-orange-500 via-orange-400 to-orange-300 text-white rounded-lg transform hover:shadow-xl hover:scale-105 focus:scale-95 transition-all ease-in-out duration-200 focus:outline-none outline-none focus:ring-2 focus:ring-orange-200',
+              buttonContainer: 'w-full flex items-center justify-center py-8',
+              comment: 'text-lg text-center text-gray-200',
+            }}
+          />
```
vojtaholik, line 93: fixed styles for engmanagement product.

packages/convertkit/src/forms/index.tsx
```diff
@@ -1,69 +1,91 @@
...
+    /** classNames for text below submit button */
+    comment?: string
+    /** children container classNames */
+    children?: string
+    form?: string
+    /** required field indicator (*) classNames */
+    asterisk?: string
+    buttonContainer?: string
+  }
 }
```
vojtaholik, line 26: I went with this `classNames` object approach, but am open to other suggestions. Seems simple and flexible enough, plus type descriptions make it nice to work with. Button component is replaceable since it's required at some times.

packages/convertkit/src/forms/index.tsx
```diff
@@ -1,69 +1,91 @@
...
+      'w-full rounded-sm dark:bg-gray-900 dark:text-white dark:placeholder-gray-300',
+    label: 'text-sm font-semibold block',
+    button:
+      'w-full px-3 py-2 font-semibold dark:bg-white dark:text-black bg-black text-white rounded-sm',
+    comment: 'text-xs italic text-center',
+    form: 'space-y-4 w-full',
+    children: 'pb-4 font-bold sm:text-xl text-lg tracking-tight',
+    asterisk: 'opacity-50',
+    buttonContainer: '',
+  },
```
vojtaholik, line 44: these are the defaults.

apps/compilersforhumans/src/pages/index.tsx
```diff
@@ -36,7 +35,19 @@ export default function Home() {
...
```
vojtaholik, line 50: correct styles for compilersforhumans product.

apps/engmanagement/src/pages/index.tsx
vojtaholik, line 93: correct styles for engmanagement product.

apps/keyboardlegend/src/pages/index.tsx
```diff
@@ -50,7 +51,18 @@ export default function Home() {
...
+                classNames={{
+                  input:
+                    'autofill:caret-black autofill:text-fill-black autofill:bg-white w-full rounded-lg border border-gray-400 focus:ring-2 focus:ring-brand-pink-500 focus:outline-none focus:border-transparent',
+                  label: 'block pb-1',
+                  form: 'space-y-4 w-full',
+                  buttonContainer:
+                    'w-full flex items-center justify-center pt-4',
+                  comment: 'text-sm text-gray-500 text-center pt-6',
+                }}
+              />
```
vojtaholik, line 65: correct styles for keyboardlegend product.

vojtaholik: with tailwind intellisense plugin you won't get autocomplete in objects like this by default, but you can achieve that by adding this to your vscode settings json: 
```json
"tailwindCSS.experimental.classRegex": [":\\s*?[\"'`]([^\"'`]*).*?"]
```

packages/convertkit/src/forms/index.tsx
joelhooks: i was immediately confused by the. among because of the [classnames library](https://github.com/JedWatson/classnames). 

is it typed? that’d be handy for using it. 

also think you could merge the defaults so they’d be useful individually and not require the entire object. 

the other approach that springs to mind is like chakra or other similar system where style props are individually addressable, which may provide a “nicer” api

i’d be very interested in seeing other existing example systems that approach this problem if they are available.

packages/convertkit/src/forms/index.tsx
joelhooks: i think this should be merged with incoming values and not either/or

packages/convertkit/src/forms/index.tsx
vojtaholik: Yeah not set on the naming either and will look if classnames library if that can be any help here. I'll also look for some real world examples and post my findings here.

> also think you could merge the defaults so they’d be useful individually and not require the entire object.

Merging the defaults is tricky since you can't change the precedence in html with the order of classnames. What matters is the specificity of css selectors. Gets weird quickly. And if I understand your point correctly, you don't need to specify the entire object and can change styles granularly `classNames: {{ button: "text-rose-500" }}`

apps/compilersforhumans/src/pages/index.tsx
vojtaholik: This is also a fix for incorrectly styled subscribe form in production.

packages/convertkit/src/forms/index.tsx
joelhooks: you’d merge the defaults and incoming object in javascript, not html so an incomplete incoming style object wouldn’t be incomplete and would still have default styles applied for any missing incoming properties

i’d like to see some “prior art” if possible because this feels like a problem that probably gets solved quite often

apps/compilersforhumans/src/pages/index.tsx
joelhooks: i’d probably recommend applying that fix in a separate PR so it can be immediately applied

joelhooks (approved):

PR #55, opened 2021-08-30
engmanagement: load articles from sanity

vojtaholik:
[preview →](https://engmanagement-turbo-dsraz0cpr-skillrecordings.vercel.app/example-chapter)

screenshots

gif


apps/engmanagement/src/pages/[slug].tsx
```diff
@@ -0,0 +1,151 @@
...
+  title,
+  mainImage,
+  description,
+  publishedAt,
+  'slug': slug.current,
+  subscribersOnly,
+  ckTagId,
+  ogImage,
+  'body': preview
+  }`
```
Comment (vojtaholik) [inline apps/engmanagement/src/pages/[slug].tsx:106]: this initial query gives us everything we need to determine whether given article is meant to be accessible to everyone or just subscribers.

apps/engmanagement/src/pages/[slug].tsx
```diff
@@ -0,0 +1,151 @@
...
+
+const restQuery = groq`*[_type == "post" && slug.current == $slug][0]{
+  body,
+  }`
```
Comment (vojtaholik) [inline apps/engmanagement/src/pages/[slug].tsx:110]: query rest of article if conditions are met.

apps/engmanagement/src/pages/[slug].tsx
```diff
@@ -0,0 +1,151 @@
...
+
+export const getServerSideProps: GetServerSideProps = async (context: any) => {
+  const allPosts = await sanityClient.fetch(allPostsQuery)
+  const currentPost = find(allPosts, {slug: context.params.slug})
+
+  if (isEmpty(currentPost)) {
+    return {
+      notFound: true,
+    }
+  }
```
Comment (vojtaholik) [inline apps/engmanagement/src/pages/[slug].tsx:126]: get all posts and find the one we're currently viewing. return 404 if slugs don't match.

apps/engmanagement/src/pages/[slug].tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  const initialData = await sanityClient.fetch(initialQuery, {
+    slug: currentPost.slug,
+  })
+
+  const subscribed = initialData.subscribersOnly
+    ? await checkSubscriber(context, initialData.ckTagId)
+    : true
```
Comment (vojtaholik) [inline apps/engmanagement/src/pages/[slug].tsx:134]: subscriber check is in separate utility function.

apps/engmanagement/src/utils/check-subscriber.ts
```diff
@@ -0,0 +1,37 @@
...
+import get from 'lodash/get'
+import isEmpty from 'lodash/isEmpty'
+import find from 'lodash/find'
+import {CK_SUBSCRIBER_KEY, ACCESS_TOKEN_KEY} from '@skillrecordings/config'
+import {fetchConvertkitSubscriberFromServerCookie} from '@skillrecordings/convertkit'
+
+export default async function checkSubscriber(
+  context: GetServerSidePropsContext,
+  tagId?: number,
+) {
```
vojtaholik, line 11: this returns whether user is authorised to view the article or not

apps/engmanagement/src/utils/sanity-client.ts
```diff
@@ -0,0 +1,7 @@
+import client from '@sanity/client'
+
+export const sanityClient = client({
```
vojtaholik, line 3: setting up sanity client.

apps/engmanagement/src/components/convertkit-subscribe-and-tag.tsx
```diff
@@ -0,0 +1,118 @@
+import React from 'react'
+import {useConvertkit} from '@skillrecordings/convertkit'
+import axios from 'axios'
+import {useFormik} from 'formik'
+import Spinner from 'components/spinner'
+import * as Yup from 'yup'
+
+const ConvertkitSubscribeAndTagForm: React.FC<{tag: string; idNum?: string}> =
```
vojtaholik, line 8: I need to get this DRYed up.

packages/convertkit/src/api/subscribe.ts
```diff
@@ -0,0 +1,47 @@
+import {NextApiRequest, NextApiResponse} from 'next'
+import {convertkitAxios} from '@skillrecordings/axios'
+import serverCookie from 'cookie'
+
+const CK_SUBSCRIBER_KEY = process.env.NEXT_PUBLIC_CONVERTKIT_SUBSCRIBER_KEY
+const subscribe = async (req: NextApiRequest, res: NextApiResponse) => {
```
vojtaholik, line 6: subscribe endpoint for subscribing and tagging.

packages/convertkit/src/utils/fetch-convertkit-subscriber.ts
```diff
@@ -37,20 +41,24 @@ export default async function fetchConvertkitSubscriberFromServerCookie(
...
   }
 
   if (isEmpty(subscriber))
     throw new Error('no convertkit subscriber was loaded')
 
+  const tags = await convertkitAxios
+    .get(`/subscribers/${subscriber.id}/tags?api_key=${CONVERTKIT_TOKEN}`)
+    .then(({data}) => data.tags)
+
+  subscriber = {...subscriber, tags}
```
vojtaholik, line 60: fetch and include tags in subscriber object.

apps/engmanagement/src/components/convertkit-subscribe-and-tag.tsx
```diff
@@ -0,0 +1,118 @@
...
+      validationSchema: Yup.object().shape({
+        email_address: Yup.string()
+          .email('Invalid email address')
+          .required('Required'),
+        first_name: Yup.string(),
+      }),
+      enableReinitialize: true,
+      onSubmit: async (values) => {
+        setSubmitting(true)
+        axios
```
joelhooks, line 27: I'd consider putting this in a serverless function and moving the call to a "service" versus "raw axios" in a callback

note that you are using `async` here and then using more traditional promise API and not `await` - I'd generally try to do one or the other

apps/engmanagement/src/components/convertkit-subscribe-and-tag.tsx
```diff
@@ -0,0 +1,118 @@
...
+            {submitting ? (
+              <div className="px-14">
+                <span className="sr-only">Submitting</span>
+                <Spinner />
+              </div>
+            ) : (
+              <>
+                Subscribe {'&'} Continue Reading
+                {/* prettier-ignore */}
+                {/* <svg className="ml-2" width="20" height="20" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg"><g fill="none"><path fillRule="evenodd" clipRule="evenodd" d="M12.293 5.293a1 1 0 0 1 1.414 0l4 4a1 1 0 0 1 0 1.414l-4 4a1 1 0 0 1-1.414-1.414L14.586 11H3a1 1 0 1 1 0-2h11.586l-2.293-2.293a1 1 0 0 1 0-1.414z" fill="currentColor"/></g></svg> */}
```
joelhooks, line 106: can be removed? I'd also move these to a component and not worry about the prettier ignore and just let it be so it's not blowing out the margins

apps/engmanagement/src/pages/[slug].tsx
```diff
@@ -0,0 +1,153 @@
...
+import ConvertkitSubscribeAndTagForm from 'components/convertkit-subscribe-and-tag'
+import groq from 'groq'
+import {sanityClient} from 'utils/sanity-client'
+import checkSubscriber from 'utils/check-subscriber'
+import {useNextSanityImage} from 'next-sanity-image'
+import Image from 'next/image'
+import {useConvertkit} from '@skillrecordings/convertkit'
+import {TheFutureOfRemoteWorkBackground} from 'components/backgrounds'
+
+export default function ExampleChapter({post, source, authorized}: any) {
```
Comment (joelhooks) [inline apps/engmanagement/src/pages/[slug].tsx:17]: component name needs to be updated?

in almost every circumstance in our codebases we prefer a `const ComponentName = ()` approach over a named function like this. Are we introducing this pattern for a specific reason? I miss the typing of the props using this approach, but if it has an advantage I'm not aware of I'd like to hear ore.

apps/engmanagement/src/pages/[slug].tsx
```diff
@@ -0,0 +1,153 @@
...
+        className="md:object-contain object-cover object-top select-none pointer-events-none"
+        quality={100}
+        placeholder="blur"
+        loading="eager"
+        priority
+      />
+    )
+  }
+  return (
+    <div className="bg-[#111725]">
```
Comment (joelhooks) [inline apps/engmanagement/src/pages/[slug].tsx:49]: this hex color being hard coded like this might benefit from extracting the color into a named `const` so it isn't as confusing. I'd consider passing it as a prop too.

apps/engmanagement/src/pages/[slug].tsx
```diff
@@ -0,0 +1,153 @@
...
+  }`
+
+const allPostsQuery = groq`
+  *[_type == "post"]{
+    "slug": slug.current
+  }
+`
+
+export const getServerSideProps: GetServerSideProps = async (context: any) => {
+  context.res.setHeader('Cache-Control', 's-maxage=1, stale-while-revalidate')
```
Comment (joelhooks) [inline apps/engmanagement/src/pages/[slug].tsx:119]: you're not going to be able to use SWR like this if the content will vary between auth'd users and non-auth'd users because the `authorized` prop will be edge cached and delivered in oddball ways between requests.

joelhooks (commented):
Primarily style comments, but the SWR issue is a blocker and I don't think it will work because of the need to check "auth" on each request.

1. auth'd user makes request, sees appropriate content
2. un-auth'd user makes request, gets cached response for auth'd user, see's content
3. auth'd user makes request and receives now stale response for un-auth'd user's previous request
4. ♻️

apps/engmanagement/src/pages/[slug].tsx
Comment (vojtaholik) [reply apps/engmanagement/src/pages/[slug].tsx]: good catch — removed, thanks!

apps/engmanagement/src/pages/[slug].tsx
Comment (vojtaholik) [reply apps/engmanagement/src/pages/[slug].tsx]: donezo 👍

apps/engmanagement/src/pages/[slug].tsx
Comment (vojtaholik) [reply apps/engmanagement/src/pages/[slug].tsx]: nah I simply missed this — updated, thanks!

apps/engmanagement/src/components/convertkit-subscribe-and-tag.tsx
vojtaholik: I took this from rustadventure so I'm probably gonna update both of them in next pass if that's cool.
I think we should also consider merging this with default subscribe form from convertkit package.

joelhooks (approved):

PR #64, opened 2021-09-07
fixes some issues with the spinner on the player

joelhooks:


packages/player/src/components/video.tsx
```diff
@@ -87,6 +87,12 @@ export const Video: React.FC<VideoProps> = ({
       onWaiting={() => {
         videoService.send('WAITING')
       }}
+      onPlaying={() => {
+        videoService.send('DONE_WAITING')
```
joelhooks, line 91: the `playing` event means we have resumed and can stop `waiting`

DrShpongle (approved):

PR #65, opened 2021-09-07
more robust hls

joelhooks:
promotes robust health

### DIFF
--- a/packages/player/src/components/hls-source.tsx
+++ b/packages/player/src/components/hls-source.tsx
@@ -29,12 +29,12 @@ export const HLSSource: React.FC<HLSSourceProps> = ({
   const video = useSelector(videoService, selectVideo)
 
   React.useEffect(() => {
-    let hls: Hls | null = null
+    let hls: Hls
 
     if (!video) return
 
     function _initPlayer() {
-      if (hls != null) {
+      if (hls) {
         hls.destroy()
       }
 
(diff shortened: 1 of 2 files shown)


packages/player/src/components/hls-source.tsx
```diff
@@ -25,57 +25,58 @@ export const HLSSource: React.FC<HLSSourceProps> = ({
...
+    const shouldUseNative = canUseNative && !Hls.isSupported()
+
     // Check for Media Source support
-    if (Hls.isSupported()) {
-      sethlsSupported(true)
+    if (!shouldUseNative) {
       _initPlayer()
     } else {
-      sethlsSupported(false)
+      videoService.send('LOADED')
```
joelhooks, line 79: if we are using native player (ios safari) we can assume its loaded here?

packages/player/src/components/hls-source.tsx
```diff
@@ -25,66 +25,66 @@ export const HLSSource: React.FC<HLSSourceProps> = ({
...
+    const shouldUseNative = canUseNative && !Hls.isSupported()
+
     // Check for Media Source support
-    if (Hls.isSupported()) {
-      sethlsSupported(true)
+    if (!shouldUseNative) {
       _initPlayer()
     } else {
-      sethlsSupported(false)
+      video.load()
```
joelhooks, line 78: on native players (not using hls.js) we want to tell the element to load and this causes the dynamic sources to be recognized. 

assumed that hls.js is doing this under the hood when it does it's things on other browsers

PR #66, opened 2021-09-07
monitor ios fullscreen state and send pause on exit since the video pauses but does not send an event

joelhooks:
monitor the pulse

### DIFF
--- a/packages/player/src/components/hls-source.tsx
+++ b/packages/player/src/components/hls-source.tsx
@@ -29,12 +29,12 @@ export const HLSSource: React.FC<HLSSourceProps> = ({
   const video = useSelector(videoService, selectVideo)
 
   React.useEffect(() => {
-    let hls: Hls | null = null
+    let hls: Hls
 
     if (!video) return
 
     function _initPlayer() {
-      if (hls != null) {
+      if (hls) {
         hls.destroy()
       }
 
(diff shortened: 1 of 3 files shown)


packages/player/src/components/hls-source.tsx
```diff
@@ -71,19 +71,32 @@ export const HLSSource: React.FC<HLSSourceProps> = ({
...
+    function handleExitedFullscreenOnIphone() {
+      videoService.send('PAUSE')
+    }
+
     // Check for Media Source support
     if (!shouldUseNative) {
       _initPlayer()
     } else {
       video.load()
+      video.addEventListener(
```
joelhooks, line 83: we want to monitor for this event on iOS specifically because the video pauses coming out of fullscreen but doesn't explicitly set an event 🥴

packages/player/src/components/video.tsx
```diff
@@ -90,9 +90,6 @@ export const Video: React.FC<VideoProps> = ({
       onPlaying={() => {
         videoService.send('DONE_WAITING')
       }}
-      onStalled={() => {
```
joelhooks, line 93: `stalled` doesn't mean what I think it means

packages/player/src/context/video-context.tsx
```diff
@@ -34,6 +34,11 @@ export const VideoProvider: React.FC = (props) => {
     actions: defaultActions,
   })
 
+  // see all state changes
+  // videoService.onTransition((state) => {
```
joelhooks, line 38: this is handy if you need to see ALL the state transitions

PR #73, opened 2021-09-08
add cue-bar to player

joelhooks:
cue laugh track


apps/next-product-starter/src/styles/cueplayer.css
```diff
@@ -0,0 +1,158 @@
+
+
+/* cue note */
```
joelhooks, line 3: @DrShpongle I brought this over so it would render but it has a couple of issues:

* it's all tailwind specific and needs to be scss in the /packages/player package 
* the play-progress-bar styles completely change the play progress bar so it can't inline like normal. this probably needs to be an optional/configurable approach so the bar can be "floating" or "inline"

DrShpongle (approved):

apps/next-product-starter/src/styles/cueplayer.css
DrShpongle: What if we extract `ProgressControl` out of `ControlBar` component and put it just under the video holder? I tried it, seems work well.

PR #76, opened 2021-09-10
Player styles

DrShpongle:
- move styles from `cueplayer.css` to the package's scss files
- remove `cueplayer.scss`
- get rid of tailwind's classes from player's code

giphy


packages/player/src/styles/scss/components/cue-bar.scss
```diff
@@ -1,40 +1,58 @@
...
+    background-color: orange;
+    cursor: pointer;
+  }
+  .cueplayer-react-cue-note-inactive {
+    opacity: 0.5;
+  }
+  .cueplayer-react-cue-note-active {
+    opacity: 1;
+  }
+  .cue-popup-content {
```
joelhooks, line 30: for consistency we should consider the same prefix for these as all the other classes `cueplayer-react`

joelhooks (approved):

PR #82, opened 2021-09-15
C4H: subscribe to tag passed from query param

vojtaholik:
gif


tayiorbeii (approved):
😎

tayiorbeii:
We will make this shared in the near future!

Issue #79, opened 2021-09-11
Report of site not rendering with firefox 


laurieontech:
https://twitter.com/tburgess57/status/1436504901350219785?s=20

laurieontech:
https://twitter.com/tburgess57/status/1436507920896757765

laurieontech:
Also with ios safari it seems? But that's not a consistent report
https://twitter.com/naterealmusicnf/status/1436523530968047618

Issue #77, opened 2021-09-10
Poor contrast on text input in darkmode (Android?)


laurieontech:
Some people are reporting black text on a black background within the sign up form fields.

This is on chrome, in one case it's only autofill on a desktop browser. In the other it's on android.

Both are using some form of dark mode, desktop is an extension and the other is "android dark theme".

Screenshots show a webkit autofill color in the desktop use case.

joelhooks:
https://twitter.com/shashiwhocodes/status/1436447639319584773?s=20

![image](https://user-images.githubusercontent.com/86834/132925858-a45caa30-84b9-4728-9cf6-123c89c58fd1.png)

https://twitter.com/shashiwhocodes/status/1436451935561953286?s=21

![image](https://user-images.githubusercontent.com/86834/132925869-b014edb1-f000-4643-85b1-32a901dea966.png)

laurieontech:
https://twitter.com/joshooaj/status/1436509829393125378?s=20

PR #83, opened 2021-09-16
c4h: fix autocomplete styles and webgl support fallback 

vojtaholik:
blob falling down the stairs


apps/compilersforhumans/src/pages/index.tsx
```diff
@@ -1,18 +1,25 @@
...
-import Animation from 'components/animation'
+import Blob from 'components/blob'
 import {useRouter} from 'next/router'
+import {WEBGL} from 'utils/webgl'
 
 export default function Home() {
   const router = useRouter()
+  const [hasWebGLSupport, setWebGLSupport] = React.useState<boolean>(false)
+  React.useEffect(() => {
+    setWebGLSupport(WEBGL.isWebGLAvailable())
```
vojtaholik, line 13: I was able to reproduce the `Failed to create WebGL context` message mentioned in #79 by disabling WebGL support in Firefox. By adding this check we can skip rendering canvas and thus avoid getting whole site to error out.

I think there might be more going on for [this user](https://twitter.com/tburgess57/status/1436507920896757765/photo/1) though as some of their warning messages mention GL Drivers and system restrictions. Would be interesting to see if this fixes it for them.

apps/compilersforhumans/src/styles/globals.css
```diff
@@ -34,6 +34,37 @@
 .text-stroke-white {
 }
 
+/* Fix Autocomplete styles for color themes */
+
+@media (prefers-color-scheme: dark) {
+  input:-webkit-autofill,
```
vojtaholik, line 40: This fixes autofill contrast issues reported in #77. Tested in Chrome and Safari on both desktop and mobile. Couldn't test on Android device, but looks like it was the same issue.

PR #85, opened 2021-09-17
shared `<SubscribeForm />` component

vojtaholik:
Form component that takes first name and email address and subscribes to either a form, tag, or sequence in ConvertKit using `/api/convertkit/subscribe` endpoint.

gif

### DIFF
--- a/packages/convertkit/src/api/subscribe.ts
+++ b/packages/convertkit/src/api/subscribe.ts
@@ -1,21 +1,35 @@
 import {NextApiRequest, NextApiResponse} from 'next'
 import {convertkitAxios} from '@skillrecordings/axios'
 import serverCookie from 'cookie'
+import {CONVERTKIT_SIGNUP_FORM, CONVERTKIT_TOKEN} from '@skillrecordings/config'
 
 const CK_SUBSCRIBER_KEY = process.env.NEXT_PUBLIC_CONVERTKIT_SUBSCRIBER_KEY
 const subscribe = async (req: NextApiRequest, res: NextApiResponse) => {
   if (req.method === 'POST') {
     try {
-      const {email_address, first_name, fields, tag} = req.body
-      const url = tag
-        ? `/tags/${tag}/subscribe`
-        : `/forms/${process.env.NEXT_PUBLIC_CONVERTKIT_SIGNUP_FORM}/subscribe`
+      const {
+        email_address,
+        first_name,
+        fields,
+        form = CONVERTKIT_SIGNUP_FORM,
+        tag,
+        sequence,
+      } = req.body
+      const getEndpoint = () => {
+        if (tag) {
+          return `/tags/${tag}/subscribe`
+        }
+        if (sequence) {
+          return `/sequences/${sequence}/subscribe`
+        }
+        return `/forms/${form}/subscribe`
+      }
       const subscriber = await convertkitAxios
-        .post(url, {
+        .post(getEndpoint(), {
           email: email_address,
           first_name,
           fields,
-          api_key: process.env.NEXT_PUBLIC_CONVERTKIT_TOKEN,
+          api_key: CONVERTKIT_TOKEN,
         })
         .then(({data}) => {
           return data.subscription.subscriber
@@ -34,7 +48,7 @@ const subscribe = async (req: NextApiRequest, res: NextApiResponse) => {
       res.setHeader('Set-Cookie', convertkitCookie)
 
       res.status(200).json(subscriber)
-    } catch (error) {
+    } catch (error: any) {
       console.log(error)
       res.status(200).end(error.message)
     }
--- a/packages/convertkit/src/api/subscribe.ts
+++ b/packages/convertkit/src/api/subscribe.ts
@@ -1,9 +1,12 @@
 import {NextApiRequest, NextApiResponse} from 'next'
 import {convertkitAxios} from '@skillrecordings/axios'
 import serverCookie from 'cookie'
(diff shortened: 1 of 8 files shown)


packages/convertkit/src/forms/subscribe/README.md
```diff
@@ -0,0 +1,76 @@
...
+```scss
+[data-sr-convertkit-subscribe-form] {
+  [data-sr-input] {}
+  [data-sr-input-label] {}
+  [data-sr-button] {}
+  [data-sr-button][disabled] {}
+}
+[data-sr-convertkit-subscribe-form='submitted'] {}
+[data-sr-convertkit-subscribe-form='error'] {}
+```
```
vojtaholik, line 76: Component's README with requirements and examples.

packages/convertkit/src/api/subscribe.ts
```diff
@@ -1,21 +1,35 @@
...
+      } = req.body
+      const getEndpoint = () => {
+        if (tag) {
+          return `/tags/${tag}/subscribe`
+        }
+        if (sequence) {
+          return `/sequences/${sequence}/subscribe`
+        }
+        return `/forms/${form}/subscribe`
+      }
```
vojtaholik, line 26: We can only do one of these while defaulting to standard form subscribe action. [ConvertKit API docs](https://developers.convertkit.com/#tag-a-subscriber) for reference.

packages/convertkit/src/forms/subscribe/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+  | ({
+      tag?: never
+      form?: never
+      /**
+       * Sequence ID to subscribe to.
+       * @type number
+       */
+      sequence: number
+    } & SubscribeFormOptions)
+  | ({tag?: never; form?: never; sequence?: never} & SubscribeFormOptions)
```
vojtaholik, line 53: Since we're picking an endpoint based on whether `tag`, `form` or `sequence` id is provided, we can allow for **only providing one of them**. Or none, in which case we fallback to default setting.

packages/convertkit/src/forms/subscribe/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+  tag,
+  form,
+  sequence,
+  actionLabel = 'Subscribe',
+  onSuccessRedirectUrl = '/confirm',
+}) => {
+  const [isSubmitting, setSubmitting] = React.useState<boolean>(false)
+  const router = useRouter()
+
+  const formik = useFormik({
```
vojtaholik, line 65: formik for the form

packages/convertkit/src/forms/subscribe/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+            placeholder="you@company.com"
+            type="email"
+            required
+          />
+          <Button isLoading={isSubmitting}>{actionLabel}</Button>
+        </>
+      )}
+      {formik.status === 'submitted' && <p>Thanks!</p>}
+      {formik.status === 'error' && <p>Something went wrong.</p>}
+    </form>
```
vojtaholik, line 132: and this is the UI part consisting of shared `Input` and `Button` atom components.

vojtaholik:
Not sure why I'm getting this error on `turbo build`. 

```
ERROR  your task graph is invalid. This is usually due to circular dependencies in either your pipeline configuration and/or among your workspace's packages that are in this execution scope
```

packages/convertkit/src/api/subscribe.ts
```diff
@@ -1,21 +1,35 @@
 import {NextApiRequest, NextApiResponse} from 'next'
 import {convertkitAxios} from '@skillrecordings/axios'
 import serverCookie from 'cookie'
+import {CONVERTKIT_SIGNUP_FORM, CONVERTKIT_TOKEN} from '@skillrecordings/config'
 
 const CK_SUBSCRIBER_KEY = process.env.NEXT_PUBLIC_CONVERTKIT_SUBSCRIBER_KEY
```
joelhooks, line 6: should this be in config too with a sensible default?

packages/convertkit/src/forms/subscribe/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+   * @type string
+   */
+  actionLabel?: string
+  /**
+   * The URL to redirect to after succesfully submitting a form. Default is `/confirm`.
+   * Submitted `email_address` will be automatically passed as URL param.
+   * Providing `false` value will skip the redirect entirely.
+   * @type string | false
+   */
+  onSuccessRedirectUrl?: string | false
```
joelhooks, line 22: i think this should be a callback and not explicitly redirect to a url, so just `onSuccess` and let us (the user) choose what action to take

packages/convertkit/src/forms/subscribe/index.tsx
joelhooks: convertkit let’s you send an arbitrary number of tags to any of these endpoints. we use that on the backend, but haven’t on the front end. it will make the logic more challenging

packages/convertkit/src/forms/subscribe/index.tsx
vojtaholik: great idea 👍

packages/convertkit/src/forms/subscribe/index.tsx
vojtaholik: True that. Added this to README under improvements.

PR #86, opened 2021-09-17
handle changing videos with the player

joelhooks:
changes

### DIFF
--- a/apps/next-product-starter/src/pages/player.tsx
+++ b/apps/next-product-starter/src/pages/player.tsx
@@ -1,27 +1,92 @@
 import * as React from 'react'
 import Layout from '@skillrecordings/react/dist/layouts'
 import {Player, VideoProvider, HLSSource} from '@skillrecordings/player'
+import {first, noop} from 'lodash'
+
+type VideoResource = {
+  title: string
+  url: string
+  subtitlesUrl?: string
+  notesUrl?: string
+}
 
 const PlayerPage = () => {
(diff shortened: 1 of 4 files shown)


apps/next-product-starter/src/pages/player.tsx
```diff
@@ -1,28 +1,79 @@
...
+              id="notes"
+              src={currentVideo.notesUrl}
+              kind="metadata"
+              label="notes"
+            />
+          )}
         </Player>
       </VideoProvider>
+      <VideoResourceList>
+        {videos.map((videoResource) => {
```
joelhooks, line 56: my first instinct was to pass in the `videos` as a prop to the `VideoResourceList` but this should primarily be a layout concern and the *shape of the data* isn't something that it needs to be specifically concerned with.

so instead the list is a wrapper and the end-user is responsible for actually styling the row based on the data that exists for the specific resource.

this could also mean that non-video resources might be present as well.

packages/player/src/hooks/use-notes-cues.ts
```diff
@@ -10,28 +10,54 @@ import {cueCountFromTracks} from '../utils/cue-count-from-tracks'
...
+      videoService.send({type: 'CLEAR_METADATA_TRACKS'})
+
+      // {TextTrackList} objects aren't real arrays so you have to
+      // convert them accordingly
+      const tracks = Array.from(textTracks || []).filter((track) => {
+        return ['metadata'].includes(track.kind)
+      })
 
-    tracks.forEach((track) => {
-      if (track.mode !== 'showing') {
```
joelhooks, line 30: since we are clearing them above, the mode doesn't matter here. It was here to prevent adding multiple instances to the array.

PR #95, opened 2021-09-20
favor xstate for state management of cues

joelhooks:
cue ball


packages/player/src/hooks/use-metadata-cues.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * This is a wrapper around xStates useSelector because we want to have
+ * a "better" comparison using lodash's isEqual which does a deeper compare
+ * between the prev/next values of the cues array from our selectCues
+ * selector
+ */
+export const useMetadataCues = () => {
+  const videoService = useVideo()
+  return useSelector(videoService, selectCues, (prev, next) =>
+    isEqual(prev, next),
```
joelhooks, line 15: was having a hard time getting this selector to update when cues were added so this uses lodash to check for equality between prev/next values in the optional comparison function of `useSelector`

instead of repeating the comparison function, I felt it was simpler to make a custom hook to wrap the comparison function

DrShpongle (approved):

PR #96, opened 2021-09-20
subscribe to ck form via serverless function

joelhooks:
subscribe


packages/convertkit/src/api/subscribe.ts
```diff
@@ -11,7 +11,7 @@ const subscribe = async (req: NextApiRequest, res: NextApiResponse) => {
   if (req.method === 'POST') {
     try {
       const {
-        email_address,
+        email,
```
joelhooks, line 14: convertkit refers to this specifically as `email` for this use case and other subscribe actions and we were transforming it away so I'm changing it back. I don't know for sure why we reverted to the `email_address` format so if this doesn't work let me know

packages/convertkit/src/forms/subscribe/subscribe-to-convertkit-form.tsx
```diff
@@ -0,0 +1,144 @@
...
+ */
+const SubscribeToConvertkitForm: React.FC<SubscribeFormProps> = ({
+  formId,
+  submitButtonElem,
+  errorMessage = <p>Something went wrong.</p>,
+  successMessage = <p>Thanks!</p>,
+  actionLabel = 'Subscribe',
+  onError = () => {},
+  onSuccess = () => {},
+  subscribeApiURL = SUBSCRIBE_API_URL,
```
joelhooks, line 49: technically, if we weren't using `email_address` you could submit to CK directly with this form, but I want to be able to specify it in the final application

packages/convertkit/src/api/subscribe.ts
```diff
@@ -50,7 +50,7 @@ const subscribe = async (req: NextApiRequest, res: NextApiResponse) => {
 
       res.setHeader('Set-Cookie', convertkitCookie)
 
-      res.status(200).json(subscriber)
+      res.status(200).json({...subscriber, email: email_address, first_name})
```
joelhooks, line 53: we have the subscriber (jsut an id), their submitted email address, and first name so we can return this information

packages/convertkit/src/forms/subscribe/subscribe-to-convertkit-form.tsx
```diff
@@ -0,0 +1,144 @@
...
+ * @constructor
+ */
+const SubscribeToConvertkitForm: React.FC<SubscribeFormProps> = ({
+  formId,
+  submitButtonElem,
+  errorMessage = <p>Something went wrong.</p>,
+  successMessage = <p>Thanks!</p>,
+  actionLabel = 'Subscribe',
+  onError = () => {},
+  onSuccess = () => {},
```
joelhooks, line 48: this is a noop unless they pass it in. if no function is passed in then the success message is displayed. Similar for `error`.

subscribe will pass `email`, `id`, and `first_name` so consumers can redirect appropriately

packages/convertkit/src/forms/subscribe/subscribe-to-convertkit-form.tsx
joelhooks: so if we want it to be `/api/list/subscribe` or whatever we could pass that in

packages/convertkit/src/forms/subscribe/subscribe-to-convertkit-form.tsx
```diff
@@ -0,0 +1,140 @@
...
+    validateOnChange: false,
+    enableReinitialize: true,
+    onSubmit: async (values) => {
+      const {email_address, first_name} = values
+      setSubmitting(true)
+      axios
+        .post(subscribeApiURL, {email_address, first_name, form: formId})
+        .then((response: any) => {
+          const subscriber: ConvertkitSubscriber = response.data
+          onSuccess(subscriber)
```
joelhooks, line 75: we don't make **any** assumptions about what happens after the form succeeds. it can simply switch to the success message and be done, but our cookie is set and we have a full CK subscriber object to pass along so some apps might choose a redirect, but we aren't forcing that behavior.

packages/convertkit/src/forms/subscribe/subscribe-to-convertkit-form.tsx
```diff
@@ -0,0 +1,140 @@
...
+          formik.setStatus('success')
```
joelhooks, line 76: previously this was in `finally` below but that would mean that it would be set for errors as well so moved it up here.

packages/convertkit/src/forms/subscribe/subscribe-to-convertkit-form.tsx
```diff
@@ -0,0 +1,145 @@
...
+}
+
+type SubscribeFormProps = {
+  actionLabel?: string
+  successMessage?: string | React.ReactElement
+  errorMessage?: string | React.ReactElement
+  submitButtonElem?: React.ReactElement
+  onError?: (error?: any) => void
+  onSuccess?: (subscriber?: ConvertkitSubscriber) => void
+  formId: number
```
vojtaholik, line 23: Can this be optional since we're defaulting to `CONVERTKIT_SIGNUP_FORM` env var?

packages/convertkit/src/api/subscribe.ts
vojtaholik: the reason is that when we were using default form action, the email field had to be named `email_address` for it to work properly. but we can change it back to `email` with this approach I think.

vojtaholik:
I like this a lot, thanks 🙌  it's a safe one to merge and I will continue working on integrating it in another PR.

vojtaholik (approved):
I like this a lot, thanks 🙌 it's also safe to merge and I will continue working on integrating it in another PR.

PR #97, opened 2021-09-21
we can send in a pricing URL

joelhooks:
how much

### DIFF
--- a/apps/next-product-starter/src/pages/buy-subscription.tsx
+++ b/apps/next-product-starter/src/pages/buy-subscription.tsx
@@ -1,0 +1,72 @@
+import * as React from 'react'
+import {FunctionComponent} from 'react'
+import DevBundles from 'data/bundles.development.json'
+import ProdBundles from 'data/bundles.production.json'
+import {SellableResource} from '@skillrecordings/types'
+import Layout from '@skillrecordings/react/dist/layouts'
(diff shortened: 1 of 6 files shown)


apps/next-product-starter/src/pages/buy-subscription.tsx
```diff
@@ -0,0 +1,72 @@
+import * as React from 'react'
```
joelhooks, line 1: I didn't affect change here, just dropped it in as a cut and paste placeholder

packages/commerce/src/hooks/use-commerce-machine.tsx
```diff
@@ -8,38 +8,44 @@ type CreateCommerceMachineProps = {
...
   upgradeFromSellable?: SellableResource
   stripePriceId?: string
+  pricingApiUrl?: string
 }
 
 const createCommerceMachine = ({
   sellable,
   upgradeFromSellable,
   stripePriceId,
+  pricingApiUrl = `${process.env.NEXT_PUBLIC_AUTH_DOMAIN}/api/v1/sellable_purchases`,
```
joelhooks, line 18: without material changes to this machine, we can send an overridable pricingApiUrl in so that when prices are loaded it's from the correct spot

PR #98, opened 2021-09-21
subscribe form tweaks

vojtaholik:
gif

### DIFF
--- a/apps/engmanagement/src/pages/[slug].tsx
+++ b/apps/engmanagement/src/pages/[slug].tsx
@@ -130,22 +130,15 @@ const Article = ({
                 This article is for subscribers only. Enter your email to
                 continue reading.
               </h4>
-              {ckTagId ? (
-                <SubscribeForm
-                  data-sr-convertkit-subscribe-form="article"
-                  tag={ckTagId}
-                  actionLabel="Continue Reading"
-                  onSuccessRedirectUrl={`${router.asPath}?continue=true`}
-                  successMessage="Thanks! The article is being unlocked..."
-                />
-              ) : (
-                <SubscribeForm
-                  data-sr-convertkit-subscribe-form="article"
-                  actionLabel="Continue Reading"
-                  onSuccessRedirectUrl={`${router.asPath}?continue=true`}
-                  successMessage="Thanks! The article is being unlocked..."
-                />
-              )}
+              <SubscribeForm
+                // subscribing to 2610221 form will trigger automation tagging
+                // with ckTagId that matches the one coming from Sanity
+                form={2610221}
+                data-sr-convertkit-subscribe-form="article"
+                actionLabel="Continue Reading"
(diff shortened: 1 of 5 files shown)


apps/engmanagement/src/pages/[slug].tsx
```diff
@@ -130,22 +130,15 @@ const Article = ({
                 This article is for subscribers only. Enter your email to
                 continue reading.
               </h4>
-              {ckTagId ? (
-                <SubscribeForm
-                  data-sr-convertkit-subscribe-form="article"
-                  tag={ckTagId}
```
Comment (vojtaholik) [inline apps/engmanagement/src/pages/[slug].tsx:136]: using form id instead of tag here which gets us closer to unified state.

packages/config/src/index.ts
```diff
@@ -11,6 +11,8 @@ export const CONVERTKIT_API_SECRET = process.env.CONVERTKIT_API_SECRET
 export const CONVERTKIT_TOKEN = process.env.NEXT_PUBLIC_CONVERTKIT_TOKEN
 export const CONVERTKIT_SIGNUP_FORM =
   process.env.NEXT_PUBLIC_CONVERTKIT_SIGNUP_FORM
+export const CONVERTKIT_SUBSCRIBE_API_URL =
+  process.env.NEXT_PUBLIC_SUBSCRIBE_API_URL || '/api/convertkit/subscribe'
```
vojtaholik, line 15: figured this might as well come from config

packages/convertkit/src/api/subscribe.ts
```diff
@@ -11,7 +11,7 @@ const subscribe = async (req: NextApiRequest, res: NextApiResponse) => {
   if (req.method === 'POST') {
     try {
       const {
-        email_address,
+        email,
```
vojtaholik, line 14: just use `email` instead of `email_address` as that's what ConvertKit expects.

packages/convertkit/src/forms/subscribe/index.tsx
```diff
@@ -157,7 +155,7 @@ const SubscribeForm: React.FC<SubscribeFormProps> = ({
           />
           <Input
             label="Email"
-            name="email_address"
+            name="email"
```
vojtaholik, line 158: didn't forget to change it here as well so that it doesn't break

packages/convertkit/src/forms/subscribe/subscribe-to-convertkit-form.tsx
```diff
@@ -20,14 +21,14 @@ type SubscribeFormProps = {
   submitButtonElem?: React.ReactElement
   onError?: (error?: any) => void
   onSuccess?: (subscriber?: ConvertkitSubscriber) => void
-  formId: number
+  formId?: number
```
vojtaholik, line 24: this is optional since we default to `process.env.NEXT_PUBLIC_CONVERTKIT_SIGNUP_FORM` later on

joelhooks (approved):

PR #100, opened 2021-09-23
Updates copy and order of products on KCD bundle site

dealingwith:
I don't have the monorepo running locally yet, so I'm dependent on a staging build to see the results, but given the level of simplicity in these changes...

fingers crossed


dealingwith:
Looks like the bundle build isn't configured...?

vojtaholik (approved):

PR #129, opened 2021-11-04
finalizing the kcdbundle for launch

joelhooks:
launch

### DIFF
--- a/apps/kcdbundle/src/pages/api/users/index.ts
+++ b/apps/kcdbundle/src/pages/api/users/index.ts
@@ -35,11 +35,14 @@ const current = async (req: NextApiRequest, res: NextApiResponse) => {
                 return slugs.includes(purchase.site)
               })
               .map((purchase: any) => {
-                return purchase.site
+                return {
+                  site: purchase.site,
+                  stripe_customer_id: purchase.stripe_customer_id,
+                }
               })
           : []
 
-        res.status(200).json(uniq(purchases))
+        res.status(200).json(purchases)
       })
       .catch((error) => {
         console.error(error)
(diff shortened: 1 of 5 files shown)


apps/kcdbundle/src/pages/api/users/index.ts
```diff
@@ -35,11 +35,14 @@ const current = async (req: NextApiRequest, res: NextApiResponse) => {
                 return slugs.includes(purchase.site)
               })
               .map((purchase: any) => {
-                return purchase.site
+                return {
+                  site: purchase.site,
+                  stripe_customer_id: purchase.stripe_customer_id,
```
joelhooks, line 40: using the stripe id to match them to existing customer if we can

apps/kcdbundle/src/pages/buy/email.tsx
```diff
@@ -57,19 +57,26 @@ const BuyEmailForm: React.FC<BuyEmailFormProps> = ({
...
-    console.debug(purchases)
-
     if (isEmpty(availableToUpgrade)) {
       throw new Error('no upgrade available')
     }
 
-    return createCheckoutSession(availableToUpgrade).then(async (data) => {
+    return createCheckoutSession(
+      availableToUpgrade,
+      email,
```
joelhooks, line 77: we want to send the email regardless of stripe customer so it can be pre-filled with whatever they enter here (validated)

apps/kcdbundle/src/utils/sessions.ts
```diff
@@ -7,20 +7,28 @@ const CHECKOUT_SESSION_URL =
...
   sellables: SellableResource[],
-  site: string = process.env.NEXT_PUBLIC_SITE_NAME || '',
+  email?: string,
+  stripe_customer_id?: string,
 ) {
   if (!CHECKOUT_SESSION_URL) throw Error('no checkout session URL')
 
+  if (sellables.length === 1) {
+    sellables[0].description =
+      "Since you've already purchased part of the bundle you can complete your bundle with this discounted course!"
```
joelhooks, line 17: we can override the description that shows up on the stripe checkout page

PR #136, opened 2021-11-28
Blocked with SidePanelToggle

DrShpongle:
This is a draft where I try to solve the following: I need to get the `withSidePanel` flag on the Player page to be able conditionall render the SidePanel. I put toggler from the left of fullscreen toggle in the control bar (currently there is 1 and 0 instead of icons). I have created the `withSidePanel` action and related xstate selector that I try to get on the Player page. But it's complicated with the fact that I can get it only withing component under the `VideoProvider`. So I tried to create the component `PlayerWrapper` that gets `fullscreenWrapperRef` from the ref prop, gets `withSidePanel` via selector and returns (with `forwardRef` with appropriate type) both via rendering content of `render` prop. But in this case I get the ref broken in the way I can't understand (probably it doesn't broken, but I get the issue illustrated on attached image).

image

Would be happy to hop on the call to discuss. Probably it's all bs from the start.
Some of my related thoughts here: https://www.notion.so/cue-react-player-805cbf2348064661a61589cf98d405ae


DrShpongle:
fixed, will push the new PR with more tweaks

PR #145, opened 2022-01-04
add an end event handler that we can override

joelhooks:
I didn't test this, but the principle is sound 🤡

### DIFF
--- a/apps/next-product-starter/src/pages/player.tsx
+++ b/apps/next-product-starter/src/pages/player.tsx
@@ -188,7 +188,13 @@ const VideoCuesList: React.FC<any> = () => {
 const Page = () => {
   return (
     <Layout>
-      <VideoProvider>
+      <VideoProvider
+        actions={{
+          onVideoEnded: (_context: any, _event: any) => {
+            console.log('do stuff')
+          },
+        }}
+      >
         <PlayerPage />
       </VideoProvider>
     </Layout>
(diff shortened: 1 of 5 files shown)


apps/next-product-starter/src/pages/player.tsx
```diff
@@ -188,7 +188,13 @@ const VideoCuesList: React.FC<any> = () => {
 const Page = () => {
   return (
     <Layout>
-      <VideoProvider>
+      <VideoProvider
+        actions={{
```
joelhooks, line 192: we can override default actions by sending in an actions object

packages/player/src/components/video.tsx
```diff
@@ -93,6 +93,9 @@ export const Video: React.FC<VideoProps> = ({
       onPlaying={() => {
         videoService.send('DONE_WAITING')
       }}
+      onEnded={() => {
```
joelhooks, line 96: this is the browser api for an ended event. dunno how consistent it is.

https://github.com/joelhooks/cueplayer-react/blob/main/src/components/Video.js#L559

it's what the currently deployed one uses

packages/player/src/machines/video-machine.ts
```diff
@@ -295,7 +295,10 @@ export const videoMachine = createMachine<VideoStateContext, VideoEvent>({
             },
           },
         },
-        ended: {on: {PLAY: 'playing'}},
+        ended: {
+          entry: ['onVideoEnded'],
```
joelhooks, line 299: we weren't doing anything when the playback ended (or even sending the event) but now it should run the action and be prepared to receive events appropriately

DrShpongle:
That works!

DrShpongle (approved):

PR #154, opened 2022-01-21
check query param for ck id as well as cookie with no asymc

joelhooks:
magic

### DIFF
--- a/apps/engmanagement/src/pages/[slug].tsx
+++ b/apps/engmanagement/src/pages/[slug].tsx
@@ -21,6 +21,9 @@ import {
 } from 'components/article-assets'
 import {useRouter} from 'next/router'
 import toast, {Toaster} from 'react-hot-toast'
+import get from 'lodash/get'
+import {ACCESS_TOKEN_KEY, CK_SUBSCRIBER_KEY} from '@skillrecordings/config'
+import * as serverCookie from 'cookie'
 
 type ArticleProps = {
   post: {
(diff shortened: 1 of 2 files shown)


apps/engmanagement/src/pages/[slug].tsx
```diff
@@ -195,9 +207,7 @@ export const getServerSideProps: GetServerSideProps = async (context: any) => {
     slug: currentPost.slug,
   })
 
-  const authorized = initialData.subscribersOnly
-    ? await checkSubscriber(context, initialData.ckTagId)
-    : true
+  const authorized = initialData.subscribersOnly ? !isEmpty(convertkitId) : true
```
Comment (joelhooks) [inline apps/engmanagement/src/pages/[slug].tsx:210]: I think if we have a CK id **at all** we can just show it and no need to check async

apps/engmanagement/src/pages/[slug].tsx
```diff
@@ -181,9 +182,20 @@ const allPostsQuery = groq`
...
+}
+
 export const getServerSideProps: GetServerSideProps = async (context: any) => {
   const allPosts = await sanityClient.fetch(allPostsQuery)
   const currentPost = find(allPosts, {slug: context.params.slug})
+  const cookieHeader = context.req.headers.cookie as string
+
+  const convertkitId =
+    context.query[CK_SUBSCRIBER_KEY] ||
+    getConvertkitFromCookieHeaders(cookieHeader)
```
Comment (joelhooks) [inline apps/engmanagement/src/pages/[slug].tsx:198]: the id can be in a query or the cookie at this point so we can check both real quick and get an id

PR #156, opened 2022-01-25
autoadvance in playlist

vojtaholik:
[Preview ↗︎](https://skill-starter-turbo-fvobcqxu5-skillrecordings.vercel.app/video/supabase-create-a-new-supabase-project)

gif


apps/next-product-starter/src/pages/video/[slug].tsx
```diff
@@ -94,6 +102,7 @@ const PlayerPage = ({resource}: any) => {
         >
           {mounted && (
             <Player
+              autoPlay={getPlayerPrefs().autoplay}
```
Comment (vojtaholik) [inline apps/next-product-starter/src/pages/video/[slug].tsx:105]: regular autoplay, can be skipped but in this case I'm passing it from a cookie

apps/next-product-starter/src/pages/video/[slug].tsx
```diff
@@ -253,11 +264,13 @@ const Page = ({data}: any) => {
         }}
         actions={{
           onVideoEnded: (_context: any, _event: any) => {
-            console.log('do stuff')
+            if (getPlayerPrefs().autoplay && !isEmpty(nextLesson)) {
+              router.push(`/video/${nextLesson.slug}`)
+            }
```
Comment (vojtaholik) [inline apps/next-product-starter/src/pages/video/[slug].tsx:269]: we want to navigate to next lesson when video ends

apps/next-product-starter/src/pages/video/[slug].tsx
```diff
@@ -70,9 +74,13 @@ const PlayerPage = ({resource}: any) => {
...
-    // @ts-ignore
-    videoService.send('LOAD_RESOURCE', {resource: currentResource})
-  }, [])
+    setCurrentResource(resource)
+    videoService.send({type: 'LOAD_RESOURCE', resource: currentResource})
+
+    if (getPlayerPrefs().autoplay) {
+      videoService.send('PLAY')
+    }
+  }, [resource])
```
Comment (vojtaholik) [inline apps/next-product-starter/src/pages/video/[slug].tsx:83]: load resource and start playing actions coming from the app

packages/player/src/components/video.tsx
```diff
@@ -51,33 +53,38 @@ export const Video: React.FC<VideoProps> = ({
...
-
-        // sometimes the event handlers aren't registered before the
-        // `canPlay` event is fired (caching, for instance) so we want
-        // to check and see if it's ready as soon as we get a ref to
-        // the HTMLVideoElement and "manually" fire an event
-        if (c && c.readyState > 3) {
-          videoService.send('LOADED')
-        }
-      }}
+      ref={videoElemRef}
```
vojtaholik, line 77: this fixes following error I was getting when using `videoElemRef`
image

I reckon It was caused by passing an inline arrow as a ref, which would go through unset / set cycle on every render ([read more](https://thoughtspile.github.io/2021/05/17/everything-about-react-refs/)).

apps/next-product-starter/src/pages/video/[slug].tsx
```diff
@@ -70,9 +74,17 @@ const PlayerPage = ({resource}: any) => {
...
+    setCurrentResource(resource)
+    videoService.send({type: 'LOAD_RESOURCE', resource: currentResource})
+
+    const delayAutoAdvance = setTimeout(() => {
+      if (getPlayerPrefs().autoplay) {
+        videoService.send({type: 'PLAY'})
+      }
+    }, 10)
+
+    return () => clearTimeout(delayAutoAdvance)
```
Comment (vojtaholik) [inline apps/next-product-starter/src/pages/video/[slug].tsx:86]: this doesn't seem like the best approach, but it's the only I could think of that worked. the issue here is with `videoElemRef` not being registered right away (and additional checks using machine context aren't helping either), resulting in machine transition into playing state with actual video being paused. it's a strange one.

apps/next-product-starter/src/pages/video/[slug].tsx
```diff
@@ -253,13 +267,15 @@ const Page = ({data}: any) => {
         }}
         actions={{
           onVideoEnded: (_context: any, _event: any) => {
-            console.log('do stuff')
+            if (getPlayerPrefs().autoplay && !isEmpty(nextLesson)) {
+              router.push(`/video/${nextLesson.slug}`)
+            }
```
Comment (joelhooks) [inline apps/next-product-starter/src/pages/video/[slug].tsx:272]:
```suggestion
            const {autoplay} = getPlayerPrefs()
            const hasNextLesson = !isEmpty(nextLesson)
            if (autoplay && hasNextLesson) {
              router.push(`/video/${nextLesson.slug}`)
            }
```

suggestion (style): I'm generally a big fan of taking an explicit approach so that the logic reads

apps/next-product-starter/src/components/player/autoplay-toggle.tsx
```diff
@@ -4,10 +4,11 @@ import Tippy from '@tippyjs/react'
...
 
 const AutoplayToggle = () => {
-  const [autoplay, setAutoplay] = React.useState<boolean>(false)
+  const [autoplayState, setAutoplayState] = React.useState<boolean>(false)
+  const {autoplayPref} = getPlayerPrefs()
 
   React.useEffect(() => {
-    setAutoplay(getPlayerPrefs().autoplay)
+    setAutoplayState(autoplayPref)
   }, [])
```
joelhooks, line 12: `autoplayPref` probably needs to be a dep here. 

nitpick (naming): I don't like postfixing with type, generally, unless there is some sort of naming conflict. In this case we might rename `autoPlayState` something like `isAutoplayActive` or similar to answer the question we are trying to answer versus the underlying generic thing that it is, if that makes sense. This is very fine grained, but just how I'm thinking about names.

apps/next-product-starter/src/components/player/autoplay-toggle.tsx
```diff
@@ -3,12 +3,12 @@ import {getPlayerPrefs, savePlayerPrefs} from '@skillrecordings/player'
...
-const AutoplayToggle = () => {
-  const [autoplayState, setAutoplayState] = React.useState<boolean>(false)
+const AutoPlayToggle = () => {
+  const [isAutoPlayActive, setIsAutoPlayActive] = React.useState<boolean>(false)
   const {autoplay} = getPlayerPrefs()
 
   React.useEffect(() => {
-    setAutoplayState(autoplay)
+    setIsAutoPlayActive(autoplay)
   }, [])
```
joelhooks, line 12:
```suggestion
  }, [autoplay])
```

definitely want that as a dep here because if the value change the state needs to update

joelhooks: I'm not sure if it was left out on purpose, and if it was I'd love to know so I can understand for the future.

PR #159, opened 2022-01-31
player(feat): delete cue notes

vojtaholik:
this implements required logic in the player to delete cue notes. 

on the application side, we can now add a confirmation dialog along with logic for different user roles — only note authors can delete their notes while admins and moderators have more control...

[preview ↗︎](https://skill-starter-turbo-git-vh-playerdelete-e709ab-skillrecordings.vercel.app/video/supabase-create-a-new-supabase-project)

https://user-images.githubusercontent.com/25487857/151830021-313ccf7a-0fa1-4381-a1a9-ffdfcc613dad.mp4

gif


apps/next-product-starter/src/lib/add-cue-note.ts
```diff
@@ -8,7 +8,7 @@ import {LessonResource} from '@skillrecordings/types'
 
 const addCueNote =
   (context: VideoStateContext, _event: VideoEvent) => async (send: any) => {
-    const cue = {
+    const cueNote = {
```
joelhooks, line 11: suggestion: `noteFromCue` might be a better name here

apps/next-product-starter/src/lib/add-cue-note.ts
```diff
@@ -18,21 +18,26 @@ const addCueNote =
...
       })
       .catch(() => {
         send('FAIL')
       })
-      .then(() => {
-        send('DONE_SUBMITTING_NOTE')
+      .then(({data}: any) => {
+        // pass cue text with props to get an id right
+        // away from response and avoid making another request
+        const cueNoteText = `{"id":"${data.id}","text":"${data.text}","type":"${data.type}","image":"${data.image}","startTime":${data.start_time},"endTime":${data.end_time}}`
```
joelhooks, line 34: suggestion: this is hard for me to parse visually and understand what is actually occurring. It looks like the primary transformation is with the start/end time, which might mean we should consider making it consistent everywhere but I'd consider it like this as a more readable approach:

```suggestion
        const {start_time: startTime, end_time:endTime, ...noteData} = data
        const cueNoteText = JSON.stringify({
          startTime,
          endTime,
          ...noteData
        })
```

with this approach I am clearly identify the transformation and using standard API to create the string data we need to build the `VTTCue`

packages/player/src/components/cue-bar.tsx
```diff
@@ -56,7 +56,6 @@ const NoteCue: React.FC<any> = ({cue, duration, className}) => {
       source: 'cue',
     })
     videoService.send('END_SEEKING')
-    videoService.send('PAUSE')
```
joelhooks, line 59: dunno if I agree with this in terms of UX where somebody has explicitly clicked on a cue. in practice it's weird when it keeps playing, but I might not be seeing the whole picture in context

vojtaholik: fixed 👍

DrShpongle (approved):
👍

PR #170, opened 2022-02-26
feat(typescriptcourse): update typography and wire up ck subscribe form

joelhooks:
typography


apps/typescriptcourse/src/components/layout.tsx
```diff
@@ -14,7 +14,11 @@ const Layout: React.FC<LayoutProps> = ({children, meta, ...props}) => {
   }
 
   return (
-    <DefaultLayout {...props} meta={{...defaultMeta, ...meta}}>
+    <DefaultLayout
+      showNavigation={false}
```
joelhooks, line 18: added this showNavigation property as intermediate solution 👎

apps/typescriptcourse/src/styles/globals.css
```diff
@@ -13,3 +13,27 @@ but it will still show up on keyboard focus.
...
+input:-webkit-autofill:hover,
+input:-webkit-autofill:focus,
+textarea:-webkit-autofill,
+textarea:-webkit-autofill:hover,
+textarea:-webkit-autofill:focus,
+select:-webkit-autofill,
+select:-webkit-autofill:hover,
+select:-webkit-autofill:focus {
+  -webkit-box-shadow: 0 0 0px 1000px rgba(0, 0, 0, 0.2) inset !important;
+}
```
joelhooks, line 39: mostly cut and paste from css destructured

joelhooks: can we get basic css in our template for the sign up form(s)?

packages/react/src/layouts/index.tsx
```diff
@@ -7,13 +7,15 @@ export type LayoutProps = {
...
   className?: string
+  showNavigation?: boolean
 }
 
 const Layout: React.FC<LayoutProps> = ({
   children,
   className,
   meta,
   noIndex,
+  showNavigation = true,
```
joelhooks, line 18: would rather use a "slot" or some other pattern but I was getting tripped up on types so this is a "punt" that should be reconsidered.

PR #176, opened 2022-03-04
feat(testingaccessibility): bring player into TA application

joelhooks:
wanted to see about getting mux integrated and ran into a few issues with the player. On was some typing disparities with useSelector that I had to ignore and don't fully understand. I was also required to implement "loadResource" in the provider, which I think shouldn't be the case. should let us just run a video with a essentially just the source URL imo.

video player


vojtaholik:
closing this in favour of #181 and will look into the issues mentioned

PR #178, opened 2022-03-07
feat(testingaccessibility): add auth via next-auth

joelhooks:
just using github provider currently

tokens


apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,22 @@
+import NextAuth from 'next-auth'
+import GithubProvider from 'next-auth/providers/github'
+
+export default NextAuth({
+  // Configure one or more authentication providers
+  providers: [
+    GithubProvider({
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:7]: https://github.com/organizations/testing-accessibility/settings/apps/testingaccessibility-com-auth is where this got set up at and you'd need to fix up the env for it to function locally

apps/testingaccessibility/src/pages/api/jwt.ts
```diff
@@ -0,0 +1,10 @@
+// This is an example of how to read a JSON Web Token from an API route
+import {getToken} from 'next-auth/jwt'
+import type {NextApiRequest, NextApiResponse} from 'next'
+
+const secret = process.env.NEXTAUTH_SECRET
+
+export default async (req: NextApiRequest, res: NextApiResponse) => {
+  const token = await getToken({req, secret})
+  res.send(JSON.stringify(token, null, 2))
```
joelhooks, line 9: this just prints the jwt at `/api/jwt` for testing. Works! 🎉

joelhooks:
this doesn't have a persistence layer so it'd be impossible to do things like associate purchases and similar

apps/testingaccessibility/prisma/schema.prisma
```diff
@@ -0,0 +1,64 @@
+// This is your Prisma schema file,
+// learn more about it in the docs: https://pris.ly/d/prisma-schema
+
+datasource db {
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
+  shadowDatabaseUrl = env("SHADOW_DATABASE_URL")
```
joelhooks, line 7: these are databases in heroku currently that contain the schema below which is the min for logging users in using next-auth

prisma's cloud offering gives you a nice looking data view into the database, but overall wasn't that interesting

apps/testingaccessibility/prisma/schema.prisma
```diff
@@ -0,0 +1,64 @@
...
+  expires_at         Int?
+  token_type         String?
+  scope              String?
+  id_token           String?  @db.Text
+  session_state      String?
+
+  user User @relation(fields: [userId], references: [id], onDelete: Cascade)
+
+  @@unique([provider, providerAccountId])
+}
```
joelhooks, line 38: a user can have many accounts where an account correlates to a provider

apps/testingaccessibility/prisma/schema.prisma
joelhooks: In the FAQ they have this about account linking, which isn't currently supported... basically if you'e got a database hooked up whatever you sign up with first will take that email and you can't readily switch them.

> When I sign in with another account with the same email address, why are accounts not linked automatically?
> Automatic account linking on sign in is not secure between arbitrary providers - with the exception of allowing users to sign in via an email addresses as a fallback (as they must verify their email address as part of the flow).
> 
> When an email address is associated with an OAuth account it does not necessarily mean that it has been verified as belonging to account holder — how email address verification is handled is not part of the OAuth specification and varies between providers (e.g. some do not verify first, some do verify first, others return metadata indicating the verification status).
> 
> With automatic account linking on sign in, this can be exploited by bad actors to hijack accounts by creating an OAuth account associated with the email address of another user.
> 
> For this reason it is not secure to automatically link accounts between arbitrary providers on sign in, which is why this feature is generally not provided by authentication service and is not provided by NextAuth.js.
> 
> Automatic account linking is seen on some sites, sometimes insecurely. It can be technically possible to do automatic account linking securely if you trust all the providers involved to ensure they have securely verified the email address associated with the account, but requires placing trust (and transferring the risk) to those providers to handle the process securely.
> 
> Examples of scenarios where this is secure include with an OAuth provider you control (e.g. that only authorizes users internal to your organization) or with a provider you explicitly trust to have verified the users email address.
> 
> Automatic account linking is not a planned feature of NextAuth.js, however there is scope to improve the user experience of account linking and of handling this flow, in a secure way. Typically this involves providing a fallback option to sign in via email, which is already possible (and recommended), but the current implementation of this flow could be improved on.
> 
> Providing support for secure account linking and unlinking of additional providers - which can only be done if a user is already signed in already - was originally a feature in v1.x but has not been present since v2.0, is planned to return in a future release.

apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,111 @@
...
+        {
+          ...token,
+          iat: Date.now() / 1000,
+          exp: Math.floor(Date.now() / 1000) + 24 * 60 * 60,
+          'https://hasura.io/jwt/claims': {
+            'x-hasura-allowed-roles': ['user'],
+            'x-hasura-default-role': 'user',
+            'x-hasura-role': 'user',
+            'x-hasura-user-id': token?.id,
+          },
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:22]: this is for hasura where some headers are required to be here:

https://hasura.io/docs/latest/graphql/core/auth/authentication/jwt.html#the-spec

this creates a token that you can use to make authenticated requests to hasura based on the identity supplied by a provider

apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,111 @@
...
+      clientSecret: process.env.GOOGLE_SECRET,
+      authorization: {
+        params: {
+          prompt: 'consent',
+          access_type: 'offline',
+          response_type: 'code',
+        },
+      },
+    }),
+  ],
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:51]: google and github providers (keys in vercel env) seems like it would have significant coverage. We can add others. LinkedIn would be another strong contender.

One thing to note is that each provider is a unique account, so they will be identified by their provider account. They don't blend.

apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,111 @@
...
+              ${
+                account?.provider === 'github' && profile?.company
+                  ? `company: "${profile?.company}"`
+                  : ''
+              }
+              ${
+                account?.provider === 'github' && profile?.twitter_username
+                  ? `twitter: "${profile?.twitter_username}"`
+                  : ''
+              }
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:91]: github carries a lot of rich profile information but google does not so I made this abomination. Feel free to make it better but it works 🤡

apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,111 @@
...
+
+        await fetch(process.env.HASURA_GRAPHQL_ENDPOINT || '', {
+          method: 'POST',
+          body: query,
+          headers: {
+            'content-type': 'application/json',
+            'X-Hasura-Admin-Secret': process.env.HASURA_ADMIN_SECRET || '',
+          },
+        })
+      }
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:106]: if there is a `user` property we want to write it to the database. One thing is that this only works once and future logins will quietly fail on the insert.

I thought about maybe trying to fetch the user first and then only writing if it wasn't found, just to be less error-ish.

There might be a better way in general from somebody that knows what they are doing.

apps/testingaccessibility/src/components/app/navigation.tsx
```diff
@@ -44,7 +46,10 @@ const Navigation: FunctionComponent<NavigationProps> = () => {
           </div>
         </a>
       </Link>
-      <div className="text-right sm:pr-3">
+      <div className="text-right sm:pr-3 flex items-center">
```
joelhooks, line 49: if you're logged in it shows your name up top right

apps/testingaccessibility/.env.development
```diff
@@ -1,3 +1,19 @@
...
+HASURA_ADMIN_SECRET=admin_secret
+
+# hasura local env -> `docker-compose up`
+PG_DATABASE_URL=postgres://postgres:postgrespassword@postgres:5432/postgres
+HASURA_GRAPHQL_METADATA_DATABASE_URL=${PG_DATABASE_URL}
+HASURA_GRAPHQL_ENABLE_CONSOLE="false"
+HASURA_GRAPHQL_DEV_MODE="true"
+HASURA_GRAPHQL_ENABLED_LOG_TYPES=startup, http-log, webhook-log, websocket-log, query-log
+HASURA_GRAPHQL_ADMIN_SECRET=${HASURA_ADMIN_SECRET}
+HASURA_GRAPHQL_JWT_SECRET='{"key": "${NEXTAUTH_SECRET}","type": "HS256"}'
```
joelhooks, line 15: moved all of this out of .env.local since it is generally useful for local dev

also learned about the `${VARIABLE}` syntax which is handy

apps/testingaccessibility/.env.template
```diff
@@ -1,28 +1,10 @@
...
-
-NEXT_PUBLIC_STRIPE_CHECKOUT_SESSIONS_CANCEL_URL=http://localhost:3000
-NEXT_PUBLIC_STRIPE_CHECKOUT_SESSIONS_SUCCESS_URL=http://localhost:3000/thanks?session_id={CHECKOUT_SESSION_ID}
-NEXT_PUBLIC_STRIPE_CHECKOUT_SESSIONS_URL=http://app.egghead.af:5000/api/v1/stripe/session?client_id=this-sites-client-id
-NEXT_PUBLIC_STRIPE_TOKEN=pk_test_token
-
-NEXT_PUBLIC_CONVERTKIT_SIGNUP_FORM=next_public_convertkit_signup_form
-NEXT_PUBLIC_CONVERTKIT_TOKEN=next_public_convertkit_token
-NEXT_PUBLIC_CONVERTKIT_SUBSCRIBER_KEY=ck_subscriber_id
-CONVERTKIT_BASE_URL=https://api.convertkit.com/v3/
```
joelhooks, line 25: felt like the "template" should be focused on what you need to be up and running locally since the prod/dev env files have all this stuff anyway

PR #194, opened 2022-03-22
chore(testingaccessibility): revert to hasura

joelhooks:
revert


apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -2,15 +2,51 @@ import NextAuth from 'next-auth'
...
+
+      return encodedToken
+    },
+    decode: async (params) => {
+      if (!params.token) return null
+
+      const verify = jwt.verify(params.token, params.secret)
+      return verify as JWT
+    },
+  },
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:49]: create encode the jwt so that it is valid for hasura as a bearer token

PR #200, opened 2022-03-28
feat(testingaccessibility): handle a default 'sale' coupon

joelhooks:
on sale


apps/testingaccessibility/src/utils/format-prices-for-product.ts
```diff
@@ -52,18 +60,49 @@ export async function formatPricesForProduct({
...
+    defaultPriceProduct = {
+      ...defaultPriceProduct,
+      calculatedPrice: getCalculatedPriced({
+        unitPrice: defaultPriceProduct.unitPrice,
+        percentOfDiscount: appliedCoupon.percentage_discount,
+        quantity,
+      }),
+      appliedCoupon,
+    }
+  }
```
joelhooks, line 104: if the couponId is passed in and it's `site` type, the default price is the price with that coupon applied

apps/testingaccessibility/src/utils/send-server-email.ts
```diff
@@ -0,0 +1,120 @@
...
+}: {
+  email: string
+  callbackUrl?: string
+}) {
+  const signInUrl = `${baseUrl}/api/auth/signin/email`
+  const url = parseUrl(`${baseUrl}/api/auth`)
+  const secret = createSecret({userOptions: nextAuthOptions, url})
+  const {cookie} = createCSRFToken({options: {secret}, isPost: false})
+  const cookieHeader = `next-auth.csrf-token=${cookie}`
+  const clientToken = await getClientCsrfToken(cookieHeader)
```
joelhooks, line 19: in the api environment, specifically in a webhook for stripe, there are no cookies in the header so the `next-auth` csrf protection doesn't work, but we want to send emails in a webhook handler so all of the above effectively recreates the header that we can use to send a magic link email

PR #205, opened 2022-04-02
feat(testingaccessibility): purchase upgrades

joelhooks:
upgrade


apps/testingaccessibility/src/components/Pricing.tsx
```diff
@@ -127,20 +121,43 @@ export const Pricing: React.FC = () => {
...
-                    action={`/api/stripe/checkout?productId=${formattedPrice?.id}&couponId=${appliedCoupon?.id}&quantity=${quantity}`}
-                    method="POST"
-                  >
+                  {purchased ? (
                     <section>
                       <button
-                        className="w-full flex items-center justify-center px-5 py-3 border border-transparent text-base font-medium rounded-md text-white bg-black hover:bg-gray-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500"
+                        className="disabled w-full flex items-center justify-center px-5 py-3 border border-transparent text-base font-medium rounded-md text-white bg-gray-500 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500"
                         type="submit"
                         role="link"
```
marcysutton, line 129: We should take these `role="link"` attributes off of here, since there isn't a good reason to have them. If they navigate people somewhere rather than posting or toggling, they should be links with URLs.

I can do a more thorough accessibility review whenever, since I saw something else in the page above!

joelhooks: got these roles removed!

PR #214, opened 2022-04-12
feat(testingaccessibility): content schemas

vojtaholik:
- document schemas for package, module, section and lesson
- block schemas for embedding external image, video, and code

image
image
image

gif

### DIFF
--- a/apps/testingaccessibility/studio/schemas/documents/package.js
+++ b/apps/testingaccessibility/studio/schemas/documents/package.js
@@ -1,0 +1,51 @@
+import React from 'react'
+
+export default {
+  name: 'package',
+  title: 'Package',
+  type: 'document',
+  fields: [
+    {
+      name: 'title',
+      title: 'Title',
+      type: 'string',
+    },
+    {
+      name: 'id',
+      title: 'ID',
+      type: 'string',
+    },
+    {
+      name: 'description',
+      title: 'Description',
+      type: 'text',
+      rows: 3,
+    },
+    {
+      name: 'features',
+      title: 'Features',
+      type: 'array',
+      of: [{type: 'feature'}],
+    },
+    {
+      name: 'modules',
+      title: 'Modules',
+      type: 'array',
+      of: [{type: 'reference', to: [{type: 'module'}]}],
+    },
+    {name: 'image', title: 'Image', type: 'externalImage'},
+  ],
+  preview: {
+    select: {
+      title: 'title',
+      media: 'image.url',
+    },
+    prepare(selection) {
+      const {media, title} = selection
+      return {
+        title: title,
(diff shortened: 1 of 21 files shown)


apps/testingaccessibility/studio/schemas/documents/package.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+
+export default {
```
joelhooks, line 3: I want to make one small change after sleeping on this... package -> product as this is directly linked to the Stripe concept of a product.

course > product > module > section > lesson > resource

"course" at the top doesn't really matter, it's just how instructors/twitter tends to refer to it. I'd rather it be "product" but I think it makes more sense to align with stripe at that "transition"

This will likely have a `productId` field that connects it to the Prisma schema for a product.

apps/testingaccessibility/studio/schemas/objects/bodyVideo.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react'
+import ReactHlsPlayer from 'react-hls-player'
+// import ReactMarkdown from 'react-markdown'
+
+export default {
```
joelhooks, line 5: all videos will be in MUX, if that has any bearing here. They have a Sanity integration...

apps/testingaccessibility/studio/schemas/documents/package.js
vojtaholik: done ✅  it's using shorter `id` instead of `productId`. I figured it won't change much, but happy to use the longer form or maybe add a description... and yes I'm using it as a connection point to Prisma schema, works like a charm.

joelhooks: I'd use productId similar to how it would be in the Prisma schema for a relationship versus "id" which would semantically represent the "id" of the Sanity instance for me.

vojtaholik: done 👍

apps/testingaccessibility/studio/schemas/objects/bodyVideo.js
vojtaholik: It's clunky and I'm having issues with getting the preview to render plus if used in PortableText, only `_ref` string is getting passed down to a component which isn't supposed to be the case. I'll look into it more but for now we can use it to upload mp4s while still providing the hls url as a string.

joelhooks: ya I don't think it's really intended for this purpose and we can likely just manage MUX manually

NicollGuarnizo (approved):

PR #227, opened 2022-05-01
fix(testingaccessibility): catch errors and use webhook secret

joelhooks:
catch

### DIFF
--- a/apps/testingaccessibility/process.d.ts
+++ b/apps/testingaccessibility/process.d.ts
@@ -11,5 +11,6 @@ declare namespace NodeJS {
     POSTMARK_KEY: string
     STRIPE_SECRET_TOKEN: string
     STRIPE_WEBHOOK_SECRET: string
+    NEXT_PUBLIC_ALGOLIA_API_WRITE_KEY: string
   }
 }
--- a/apps/testingaccessibility/process.d.ts
+++ b/apps/testingaccessibility/process.d.ts
@@ -11,6 +11,8 @@ declare namespace NodeJS {
     POSTMARK_KEY: string
     STRIPE_SECRET_TOKEN: string
     STRIPE_WEBHOOK_SECRET: string
-    NEXT_PUBLIC_ALGOLIA_API_WRITE_KEY: string
+    ALGOLIA_API_WRITE_KEY: string
+    SANITY_WEBHOOK_SECRET: string
+    NEXT_PUBLIC_ALGOLIA_APPLICATION_ID: string
   }
 }
--- a/apps/testingaccessibility/src/pages/api/sanity/webhooks/add-all-records-to-algolia.ts
+++ b/apps/testingaccessibility/src/pages/api/sanity/webhooks/add-all-records-to-algolia.ts
@@ -35,7 +35,7 @@ export const addAllRecordsToAlgolia = async (
     }
   } catch (error) {
     console.error(error)
-    res.status(500).json({success: false, message: error.message})
+    res.status(500).json({success: false, message: error?.message})
   }
 }
 
--- a/apps/testingaccessibility/src/pages/api/sanity/webhooks/add-all-records-to-algolia.ts
+++ b/apps/testingaccessibility/src/pages/api/sanity/webhooks/add-all-records-to-algolia.ts
@@ -4,9 +4,9 @@ import {sanityAlgolia} from 'utils/algolia'
 import {setupHttpTracing} from '@vercel/tracing-js'
 import {withSentry} from '@sentry/nextjs'
 import {tracer} from 'utils/honeycomb-tracer'
-import get from 'lodash/get'
+import {isValidRequest} from '@sanity/webhook'
 
-const secret = process.env.NEXT_PUBLIC_ALGOLIA_API_WRITE_KEY as string
+const secret = process.env.NEXT_PUBLIC_ALGOLIA_API_WRITE_KEY
 
 export const addAllRecordsToAlgolia = async (
   req: NextApiRequest,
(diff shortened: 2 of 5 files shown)


apps/testingaccessibility/process.d.ts
```diff
@@ -11,5 +11,8 @@ declare namespace NodeJS {
     POSTMARK_KEY: string
     STRIPE_SECRET_TOKEN: string
     STRIPE_WEBHOOK_SECRET: string
+    ALGOLIA_API_WRITE_KEY: string
+    SANITY_WEBHOOK_SECRET: string
+    NEXT_PUBLIC_ALGOLIA_APPLICATION_ID: string
```
joelhooks, line 16: adding type here stops the need for casting these `as string` etc

apps/testingaccessibility/src/pages/api/sanity/webhooks/add-all-records-to-algolia.ts
```diff
@@ -4,9 +4,7 @@ import {sanityAlgolia} from 'utils/algolia'
 import {setupHttpTracing} from '@vercel/tracing-js'
 import {withSentry} from '@sentry/nextjs'
 import {tracer} from 'utils/honeycomb-tracer'
-import get from 'lodash/get'
-
-const secret = process.env.NEXT_PUBLIC_ALGOLIA_API_WRITE_KEY as string
```
joelhooks, line 9: the algolia write key was being used in the sanity request validator which as far as I can tell requires the Sanity webhook secret which wasn't in the environment variables so it's added below

🚨 also note: **private keys should never be prefixed with `NEXT_PUBLIC`**.

apps/testingaccessibility/src/pages/api/sanity/webhooks/add-all-records-to-algolia.ts
```diff
@@ -18,30 +16,27 @@ export const addAllRecordsToAlgolia = async (
     req,
     res,
   })
-  // Basic security to prevent others from hitting this API
-  const s = get(req.query, 's')
-  if (s !== secret) {
-    res.status(403).json({success: false, message: 'Invalid signature'})
-    return
-  }
```
joelhooks, line 26: this felt like it needed `isValidRequest` instead, but I wasn't 100% on it

apps/testingaccessibility/src/pages/api/sanity/webhooks/add-all-records-to-algolia.ts
```diff
@@ -18,30 +16,27 @@ export const addAllRecordsToAlgolia = async (
...
 
-  // Fetch the _id of all the documents we want to index
-  const types = ['section', 'lesson']
-  const query = `* [_type in $types][]._id`
+  try {
+    if (!isValidRequest(req, process.env.SANITY_WEBHOOK_SECRET)) {
```
joelhooks, line 21: us the `SANITY_WEBHOOK_SECRET` to validate the webhook

apps/testingaccessibility/src/pages/api/sanity/webhooks/add-all-records-to-algolia.ts
```diff
@@ -18,30 +16,27 @@ export const addAllRecordsToAlgolia = async (
...
       res.status(200).send('Success!')
-    })
-    .catch((error) => {
-      res.status(500).json({success: false, message: error})
-    })
+    }
+  } catch (error) {
+    console.error(error)
+    res.status(500).json({success: false, message: error.message})
+  }
```
joelhooks, line 39: instead of catching in the promise we use the try/catch block around the async/await and capture the error while also writing it to the console on our side.

noticed that in the sanity logs the `message` property was a blank object so here we are responding with the `error.message` so that the Sanity logs will have more info and also writing the full error to our console (it was being swallowed previously)

apps/testingaccessibility/src/pages/api/sanity/webhooks/add-records-to-algolia.ts
```diff
@@ -1,14 +1,11 @@
 import {NextApiRequest, NextApiResponse} from 'next'
-import type {SanityClient} from '@sanity/client'
 import {sanityClient} from 'utils/sanity-client'
 import {sanityAlgolia} from 'utils/algolia'
 import {setupHttpTracing} from '@vercel/tracing-js'
 import {withSentry} from '@sentry/nextjs'
 import {tracer} from 'utils/honeycomb-tracer'
 import {isValidRequest} from '@sanity/webhook'
 
-const secret = process.env.NEXT_PUBLIC_ALGOLIA_API_WRITE_KEY as string
```
joelhooks, line 10: this was definitely an error since the Algolia write key would be very unlikely to also be the Sanity webhook signing secret (I checked, and they weren't the same)

apps/testingaccessibility/src/utils/algolia/sanity-algolia-client.ts
```diff
@@ -1,19 +1,22 @@
-import algoliasearch from 'algoliasearch'
-import indexer, {flattenBlocks} from 'sanity-algolia'
+import algoliaSearchClientCreator from 'algoliasearch'
+import sanityAlgoliaIndexer from 'sanity-algolia'
```
joelhooks, line 2: I did some renaming in here because all the "clients" and "apis" were confusing to me 😅

apps/testingaccessibility/src/utils/sanity-client.ts
```diff
@@ -1,6 +1,7 @@
 import client from '@sanity/client'
+import type {SanityClient} from '@sanity/client'
 
-export const sanityClient = client({
+export const sanityClient: SanityClient = client({
   projectId: 'erxflg2z',
```
joelhooks, line 5: the projectId should be an env variable for sure

PR #235, opened 2022-05-09
feat(kcdbundle): dynamic pricing

vojtaholik:
this uses our `prices` endpoint so that we get updated values when the current offer changes or runs out. 

image

gif


joelhooks (approved):
🚀🏆

PR #240, opened 2022-05-11
test(testingaccessibility): getResourcePaths utils

vojtaholik:
lisa wants to do a test

### DIFF
--- a/apps/testingaccessibility/src/utils/get-resource-paths.test.ts
+++ b/apps/testingaccessibility/src/utils/get-resource-paths.test.ts
@@ -1,0 +1,142 @@
+import {getPathForLesson, getPathForSection} from './get-resource-paths'
+
+test('correctly resolves lesson path', async () => {
+  const lessonSlug = 'testing-zoom-in-the-browser'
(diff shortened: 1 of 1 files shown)


apps/testingaccessibility/src/utils/get-resource-paths.test.ts
```diff
@@ -0,0 +1,142 @@
...
+        lessons: null,
+        slug: 'intro-and-setup',
+        title: 'Intro & Setup',
+      },
+      {
+        lessons: [
+          {
+            slug: 'identify-issues-by-hitting-tab',
+            title: 'Identify Issues by Hitting Tab',
+          },
```
vojtaholik, line 39: I wasn't sure if making an actual request to sanity is a good idea given the content is behind a paywall. so I decided to hardcode response of a minimal `modules` query corresponding to what we're doing in the app.

joelhooks: in this case you wouldn't need to mock the sanity request and instead you'd have data in the correct "shape" so this is great

joelhooks (approved):
⭐️

PR #242, opened 2022-05-12
Add some bin commands to Testing Accessibility

garrettdimon:
This provides a starter kit for how we can streamline the development process. It's more of a first draft than the end game, but hopefully it helps. Since I was trying to get Testing Accessibility set up locally, this helped work through that process.

These are written in bash simply to make them more globally useful, but they could just as easily be written in any scripting language. For instance, in a Rails app, they may make more sense to be written in Ruby.

These scripts could also proliferate very quickly and be significantly improved. We could even start nesting them if it made sense. While the ideal would be that every Product exposes a subset of these commands, the nice thing is that as long as they're documented, it's not that big of a deal if it's a subset or a superset. At the very least `bin/setup` and `bin/doctor` could be consistent.

For example:

```
bin/
  doctor
  setup
  test
  review/
    all
    syntax
    dependencies
    tests
  start/
    server
    database
    webhooks
  db/
    create
    seed
    migrate
  billing/
    stripe
    paypal
```

The primary driver here is to minimize the cognitive load of context-switching between Products and reduce the effort for someone to safely begin contributing to a new project without having to learn precisely how to run the various commands for it. The commands can also call each other to build more robust one-stop-shop options.

Another nice thing is that these commands can be used for CI in order to maintain consistency between how the commands are run locally and how they're run in CI/Staging/Production.

Finally, I didn't have access to the Postmark or Stripe accounts for Testing Accessibility in order to provide the URLs for retrieving those secret values. And in the case of GitHub, it wasn't clear where to get those values—I assume we have OAuth apps set up somewhere, but I couldn't find them.


apps/testingaccessibility/bin/seed
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Description:
+#   Seeds the database.
+# Usage:
+#   bin/seed
+
+product='testing-accessibility'
+branch='next-steps'
```
joelhooks, line 9: this should be set to another or variable database

next-steps is already seeded!

joelhooks: “database create branch $name” would be great

garrettdimon: Ended up going with `pscale branch switch $planetscale_branch --database $database --create`.

That way it just switches if the branch exists but creates it if it doesn't exist.

It also derives the Planetscale DB name from the Git branch.

joelhooks: should this be removed then because it's still hard coded to my branch in this script from what I can tell

stuff above looks good

garrettdimon: Good catch. Forgot to adjust that one.

joelhooks (commented):
⭐️ This looks great to me, I'm gonna leave it to @vojtaholik to hit approve!

vojtaholik (approved):

PR #244, opened 2022-05-13
feat(testingaccessibility): feedback form

vojtaholik:
https://user-images.githubusercontent.com/25487857/168257703-fb3c9fa3-34af-4b82-b622-bfc52acf0987.mp4

feedback field is using simple version of [tiptap editor](https://tiptap.dev/) (really one of the best out there). works great with formik! would be awesome for notes authoring as well.

gif


vojtaholik:
I used pnpm v7 to generate the lock file causing builds to fail... 😭

PR #251, opened 2022-05-17
update(testingaccessibility): uses skill-api for sending feedback

joelhooks:
feedback


apps/testingaccessibility/src/pages/api/progress/index.ts
```diff
@@ -13,7 +13,10 @@ const getProgress = async (req: NextApiRequest, res: NextApiResponse) => {
     req,
     res,
   })
-  const sessionToken = await getDecodedToken(req)
+  const sessionToken = await getToken({
+    req,
+    secret: process.env.NEXTAUTH_SECRET,
+  })
```
joelhooks, line 19: having a challenging time with the typing of the decoded token

looking at the code for next-auth, this would actually be the default behavior anyway, but didn't test it.

apps/testingaccessibility/src/pages/learn/[module]/[section]/[lesson]/index.tsx
```diff
@@ -72,20 +57,8 @@ const sectionQuery = groq`*[_type == "section" && slug.current == $slug][0]{
...
-    (await getPurchasesForUser(sessionToken.sub))
-  const productId = purchases && get(last(purchases), 'productId')
-
-  // fetch product from sanity based on user's productId associated with their purchase
-  const product = await sanityClient.fetch(productQuery, {
-    productId: productId,
-  })
-
   // get array of available lessons
+  const {product} = await getPurchasedProduct(req)
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/learn/[module]/[section]/[lesson]/index.tsx:61]: this pattern was repeated in several places so I consolidated it. This should also make it simpler to add more permission stuff.

apps/testingaccessibility/src/pages/learn/index.tsx
```diff
@@ -63,27 +59,16 @@ const productQuery = groq`*[_type == "product" && productId == $productId][0]{
   }`
 
 export const getServerSideProps: GetServerSideProps = async ({req}) => {
-  const sessionToken = await getDecodedToken(req)
+  const {product, purchases} = await getPurchasedProduct(req, productQuery)
```
joelhooks, line 62: the `productQuery` here was a lot bigger and I didn't know if it should just be the default or if we wanted smaller queries for specific reasons so I made it an optional arg

apps/testingaccessibility/src/server/get-purchased-product.ts
```diff
@@ -0,0 +1,55 @@
...
+    sections[]->{
+      "slug": slug.current,
+      title,
+      lessons[]->{
+        "slug": slug.current,
+        title
+      }
+    }
+  }
+  }`
```
joelhooks, line 22: this looked like it was the biggest and most complete so I used it as a default

apps/testingaccessibility/src/server/get-purchased-product.ts
```diff
@@ -0,0 +1,55 @@
...
+  const sessionToken = await getToken({
+    req,
+    secret: process.env.NEXTAUTH_SECRET,
+  })
+  if (sessionToken && sessionToken.sub) {
+    const {getPurchasesForUser} = getSdk()
+    const purchases = await getPurchasesForUser(sessionToken.sub)
+
+    if (purchases) {
+      const productId = get(last(purchases), 'productId')
```
joelhooks, line 43: this makes me cringe a bit based on things oddly sorting unexpectedly, but left it for now

packages/scripts/eslint-preset.js
```diff
@@ -1,5 +1,8 @@
 module.exports = {
   extends: ['react-app'],
+  rules: {
+    'react/react-in-jsx-scope': 'off',
```
joelhooks, line 4: seems like this rule isn't needed anymore so I turned it off because it's annoying

packages/skill-api/src/client/index.ts
```diff
@@ -1 +1,3 @@
 export * from './send-feedback'
+export * from './get-decoded-token'
+export * from './get-feedback-emoji'
```
joelhooks, line 3: moved these to `client` because where `lib` is more like `server` and should likely be named as such

joelhooks (commented):
I didn't test this very much locally, so please give it a quick check!

apps/testingaccessibility/src/server/get-purchased-product.ts
vojtaholik: I agree but wasn't sure what the best approach would be here. They could be sorted on Sanity level which would then allow us to write a sorting function based off of that.

packages/skill-api/src/client/index.ts
vojtaholik: I appreciate this context comment 👍

apps/testingaccessibility/src/server/get-purchased-product.ts
vojtaholik: looks right to me and everything seems to be working 👍

vojtaholik (approved):
While testing the feedback form I noticed we were missing `NEXT_PUBLIC_SITE_TITLE=Testing Accessibility` so got it added. I also cleaned up some unused imports.

PR #252, opened 2022-05-18
feat(testingaccessibility): purchases added to session token

joelhooks:
session

### DIFF
--- a/apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
+++ b/apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
@@ -53,6 +53,12 @@ export const nextAuthOptions: NextAuthOptions = {
   ],
   callbacks: {
     async session({session, user, token}) {
+      if (token?.id) {
+        const {getPurchasesForUser} = getSdk()
+        const purchases = await getPurchasesForUser(token.id as string)
+        token.purchases = purchases
+      }
+
       const encodedToken = jwt.sign(token, process.env.NEXTAUTH_SECRET || '', {
         algorithm: 'HS256',
       })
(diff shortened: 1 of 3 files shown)


apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -53,6 +53,12 @@ export const nextAuthOptions: NextAuthOptions = {
   ],
   callbacks: {
     async session({session, user, token}) {
+      if (token?.id) {
+        const {getPurchasesForUser} = getSdk()
+        const purchases = await getPurchasesForUser(token.id as string)
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:58]: this could potentially be a performance issue but it seemed fast and we were doing it on all the learn routes anyway so this just gives us a very convenient way to access the purchases for a user in middleware etc since they will be on the token

apps/testingaccessibility/src/pages/learn/_middleware.ts
```diff
@@ -4,10 +4,9 @@ import {NextApiRequest} from 'next'
 import {getToken} from 'next-auth/jwt'
 
 export async function middleware(req: NextApiRequest, ev: NextFetchEvent) {
-  const session = await getToken({
-    req,
-    secret: process.env.NEXTAUTH_SECRET,
-  })
+  const session = await getToken({req})
```
joelhooks, line 7: no need to pass in the secret because it's the default

apps/testingaccessibility/src/server/get-purchased-product.ts
```diff
@@ -31,15 +31,12 @@ export async function getPurchasedProduct(
...
-  const sessionToken = await getToken({
-    req,
-    secret: process.env.NEXTAUTH_SECRET,
-  })
+  const sessionToken = await getToken({req})
   if (sessionToken && sessionToken.sub) {
-    const {getPurchasesForUser} = getSdk()
-    const purchases = await getPurchasesForUser(sessionToken.sub)
+    // no need to reload purchases since they are on the session
+    const purchases = sessionToken.purchases
```
joelhooks, line 37: we can avoid this now since it is on the token

vojtaholik (approved):

PR #260, opened 2022-05-26
feat(testingaccessibility): make the token available in user context

joelhooks:
context

### DIFF
--- a/apps/testingaccessibility/src/components/app/navigation.tsx
+++ b/apps/testingaccessibility/src/components/app/navigation.tsx
@@ -5,17 +5,42 @@ import {isSellingLive} from 'utils/is-selling-live'
 import {useSession, signOut} from 'next-auth/react'
 import {Menu, Transition} from '@headlessui/react'
 import {NextRouter, useRouter} from 'next/router'
-import {Logo} from 'components/images'
 import isEmpty from 'lodash/isEmpty'
 import toast from 'react-hot-toast'
 import Link from 'next/link'
 import cx from 'classnames'
 import Image from 'next/image'
+import {useUser} from '../../context/user-context'
+import {getAbilityFromToken} from '../../server/ability'
+import {useQuery} from 'react-query'
+
+// TODO: This is a temporary solution to get the ability from the token
+//       It's named poorly and is clumping stuff together, but it works!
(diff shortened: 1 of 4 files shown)


apps/testingaccessibility/src/components/app/navigation.tsx
```diff
@@ -5,37 +5,63 @@ import {isSellingLive} from 'utils/is-selling-live'
...
 import Image from 'next/image'
+import {useUser} from '../../context/user-context'
+import {getAbilityFromToken} from '../../server/ability'
+import {useQuery} from 'react-query'
+
+// TODO: This is a temporary solution to get the ability from the token
+//       It's named poorly and is clumping stuff together, but it works!
+function useRefinedSession() {
+  const {data, status} = useSession()
+  const {user} = useUser()
```
joelhooks, line 21: added a new context since the decoded token was available outside of the "session"

there might be a more direct way to do this with next-auth but I didn't see it

apps/testingaccessibility/src/components/app/navigation.tsx
```diff
@@ -5,37 +5,63 @@ import {isSellingLive} from 'utils/is-selling-live'
...
+
+  const isSignedIn = Boolean(data?.user)
+  const isLoadingUser = status === 'loading'
+
+  // this is all client side so it isn't secure secure but for display
+  // purposes it's fine
+  const {data: ability} = useQuery(
+    ['ability', user],
+    async () => await getAbilityFromToken(user),
+  )
```
joelhooks, line 31: this works pretty well! we can use this all over.

vojtaholik: nice!

vojtaholik (approved):
I like where this is going 👍

apps/testingaccessibility/src/components/app/navigation.tsx
vojtaholik: I noticed this comment earlier but wasn't able to get it working: https://github.com/skillrecordings/products/blob/main/apps/testingaccessibility/src/pages/api/auth/%5B...nextauth%5D.ts/#L65

PR #261, opened 2022-05-30
feat(testingaccessibility): cache ability rules in session

joelhooks:
caching


apps/testingaccessibility/src/components/app/navigation.tsx
```diff
@@ -10,33 +10,21 @@ import toast from 'react-hot-toast'
...
-    isSignedIn,
-    isLoadingUser,
-    data,
-    status,
-    user,
-    ability,
+    isSignedIn: status === 'authenticated',
+    isLoadingUser: status === 'loading',
   }
 }
```
joelhooks, line 29: `useNavState` is *specific* to this component with some handy booleans for layout purposes below. Currently they are only based on status, but could also utilize abilities

apps/testingaccessibility/src/hooks/use-current-ability.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Uses next-auth session data to return a viewer's current ability.
+ */
+export function useCurrentAbility() {
+  const {data: sessionData, status} = useSession()
+
+  return {
+    status,
+    ability: getCurrentAbility({rules: sessionData?.rules}),
```
joelhooks, line 12: most of the rules require that the user is authenticated, but there might be universal rules or defaults where an ability would only be true for an unauthenticated user, for instance

currently all of our abilities are related to purchases and auth'd users

apps/testingaccessibility/src/pages/_app.tsx
```diff
@@ -14,7 +14,6 @@ import {QueryClient, QueryClientProvider} from 'react-query'
 import {initNProgress} from 'utils/nprogress'
 import {ProgressProvider} from 'context/progress-context'
 import {FeedbackProvider} from 'context/feedback-context'
-import {UserProvider} from '../context/user-context'
```
joelhooks, line 17: was able to remove this 🎉

apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -57,14 +58,17 @@ export const nextAuthOptions: NextAuthOptions = {
         const {getPurchasesForUser} = getSdk()
         const purchases = await getPurchasesForUser(token.id as string)
         token.purchases = purchases
+        session.rules = await defineRulesForPurchases(purchases)
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts]: at first I put the purchases on the session but [CASL recommends caching the rules on the session](https://casl.js.org/v5/en/cookbook/cache-rules) so I tried that and it worked well!

This should be fresh and gets refreshed consistently

apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -57,14 +58,17 @@ export const nextAuthOptions: NextAuthOptions = {
         const {getPurchasesForUser} = getSdk()
         const purchases = await getPurchasesForUser(token.id as string)
         token.purchases = purchases
+        session.rules = defineRulesForPurchases(purchases)
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:61]: this is the **magic** where we create **rules** based on the **purchases** of a specific user id

rules are refreshed on page load and stored on the session so session should always have the current rules

the rules are "cached" on the session so they can be used with `useSession` to use abilities at runtime with `useCurrentAbility`

apps/testingaccessibility/src/pages/learn/index.tsx
```diff
@@ -62,17 +62,26 @@ const productQuery = groq`*[_type == "product" && productId == $productId][0]{
...
-  const ability = await getAbilityFromToken(token)
+  const session = await getSession({req})
+
+  // TODO: instead of fetching the product we can generate rules on the session
+  //       which may mean we can avoid this async call but it also isn't hurting
+  //       anything right now. The ability isn't being used to make any decisions
+  //       at this point.
+  const {product, purchases} = await getPurchasedProduct(req, productQuery)
+
+  const ability = getCurrentAbility({rules: session?.rules})
```
joelhooks, line 73: technically in this case we could generate the rules based on the purchases above, but instead we'll just send in the cached rules so we aren't doing the "work" again

```
const ability = getCurrentAbility({purchases})
```

but in most circumstances we wouldn't be loading all purchases so just using rules here

**you can generate the ability using either rules or purchases**

apps/testingaccessibility/src/pages/learn/index.tsx
```diff
@@ -62,17 +62,26 @@ const productQuery = groq`*[_type == "product" && productId == $productId][0]{
...
+  )
+  console.log(`LEARN: can view content: ${ability.can('view', 'Content')}`)
 
   console.log(
-    `can view product: ${ability.can('view', subject('Product', product))}`,
+    `LEARN: can view product: ${ability.can(
+      'view',
+      subject('Product', product),
+    )}`,
   )
```
joelhooks, line 85: we can make decisions about what to render or how to redirect based on abilities as needed

apps/testingaccessibility/src/server/ability.ts
```diff
@@ -1,64 +1,73 @@
...
+  return new AppAbility(rules)
+}
 
-  console.log({token})
+/**
+ * Creates a structure of rules to use in the ability from list of purchases
+ *
+ * @param purchases
+ */
+export function defineRulesForPurchases(purchases: any[]) {
```
joelhooks, line 52: this can be called directly with purchases array to get a rules abject that can be cached and used later

apps/testingaccessibility/src/context/progress-context.tsx
```diff
@@ -24,36 +26,37 @@ export function useProgress() {
...
-    if (status === 'authenticated') {
-      fetchProgress().catch(console.error)
-    }
-  }, [fetchProgress, status])
+  return {
+    progress: data,
+    status: loadingProgressStatus,
+    refetch,
+  }
+}
```
joelhooks, line 49: `useQuery` does some nice caching and state management and works well for this type of loading activity

here I'm just wrapping it in a hook, but you could **probably** get away with just using the hook and not the context/provider.

apps/testingaccessibility/src/hooks/use-nav-state.ts
```diff
@@ -0,0 +1,13 @@
+import {useCurrentAbility} from './use-current-ability'
+
+/**
+ * TODO: consider removing this and using abilities?
```
joelhooks, line 4: these booleans are generally counter to the idea behind CASL which is more "can" over "is" but for these two particular pieces of state we are are looking at a state machine type structure where we transition from "loading" to "authenticated"

https://xstate.js.org/docs/packages/xstate-fsm/#super-quick-start

introducing a literal state machine is interesting

joelhooks: this is only interesting if we think this state is going to become more complex, might also consider just checkoint `state === 'somestate'` in the component and just deleting this altogether

joelhooks: tried this out and it was a little silly :)

apps/testingaccessibility/src/components/app/navigation.tsx
```diff
@@ -10,35 +10,7 @@ import toast from 'react-hot-toast'
...
-
-  return {
-    isSignedIn,
-    isLoadingUser,
-    data,
-    status,
-    user,
-    ability,
-  }
-}
```
joelhooks, line 41: this was mixing several concerns that are now broken off into their own bits

apps/testingaccessibility/src/context/progress-context.tsx
```diff
@@ -24,36 +26,37 @@ export function useProgress() {
...
+  const {progress, status, refetch} = useLoadProgress()
   return (
     <ProgressContext.Provider
       value={{
         progress,
-        isLoadingProgress,
+        isLoadingProgress: status === 'loading',
         toggleLessonComplete: async (slug: string) => {
-          await toggleLessonProgressForUser({slug}).then(() => fetchProgress())
+          return await toggleLessonProgressForUser({slug}).finally(refetch)
```
joelhooks, line 59: TODO: test if the refetch works as expected. 

@vojtaholik I noticed you'd added some options to the useQuery call. I think using the `status` prop in the queryKey should keep it refreshed and eliminate the need to configure it more complexly.

apps/testingaccessibility/src/hooks/use-nav-state.ts
joelhooks: my conclusion here is that for what this is used for it is fine, but if it starts to grow or get used for other areas we night consider making it more "interesting"

this can probably move to the skill-api

vojtaholik: I agree that this is fine for our current use case, and I like that it's clearly labeled as such. Similarly, I can see the counterarguments and how thinking about it in a more holistic way might lead to some interesting solutions. Already the return statement below becomes a bit weird when I add `canInviteTeam` to it.

apps/testingaccessibility/src/hooks/use-current-ability.ts
vojtaholik: ConvertKit stuff would fit great in here as well? Pre-launch abilities are all CK driven.

vojtaholik (approved):

PR #264, opened 2022-06-03
update(testingaccessibility): add "Invite Team" links + add `canViewInvoice` check

vojtaholik:
Adding `Invite Team` links to both main and footer navigation for users with bulk purchase.

While testing this, I noticed that team members (without purchase) don't have `merchantChargeId` associated and thus shouldn't have access to invoice. So I added `canViewInvoice` ability and used it to determine presence of `Invoices` links.

screenshot


apps/testingaccessibility/src/server/ability.ts
```diff
@@ -69,5 +70,9 @@ export function defineRulesForPurchases(purchases: any[]) {
     })
   }
 
+  if (hasInvoice(purchases)) {
+    can('view', 'Invoice')
```
joelhooks, line 74: nice one!

apps/testingaccessibility/src/hooks/use-nav-state.ts
```diff
@@ -4,10 +4,12 @@ import {useCurrentAbility} from './use-current-ability'
...
  */
 export function useNavState() {
-  const {status} = useCurrentAbility()
+  const {status, ability} = useCurrentAbility()
 
   return {
     isSignedIn: status === 'authenticated',
     isLoadingUser: status === 'loading',
+    canViewTeam: ability.can('view', 'Team'),
+    canViewInvoice: ability.can('view', 'Invoice'),
```
joelhooks, line 13: love this, seems super flexible

joelhooks (approved):

PR #265, opened 2022-06-06
feat(badass): podcast rough page and sanity

joelhooks:
https://www.loom.com/share/93d00c34d5e641ddb34c597a4cc68bd4

podcaster


Creeland (approved):
Looks good. From here we can model Podcasts in Sanity

PR #266, opened 2022-06-06
feat(badass): badass podcast

joelhooks:
https://www.loom.com/share/d8f0debb26444cdda2c15d400035a44d

podcast


apps/badass/src/lib/podcast.ts
```diff
@@ -0,0 +1,114 @@
+import groq from 'groq'
```
joelhooks, line 1: this provides an "SDK" for loading the podcast data from sanity

PR #268, opened 2022-06-08
feat(testingaccessibility): make a dynamic sitemap

joelhooks:
i am the map


apps/testingaccessibility/src/lib/articles.ts
```diff
@@ -0,0 +1,14 @@
+import {sanityClient} from '../utils/sanity-client'
+import groq from 'groq'
+
+export async function getAllArticles() {
```
joelhooks, line 4: generally any api request etc should be in the 'lib' folder

apps/testingaccessibility/src/pages/server-sitemap.xml/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+    ...reviews.map((review: any) => {
+      return {
+        loc: `${process.env.NEXT_PUBLIC_URL}/accessibility-reviews/${review.slug}`, // Absolute url
+        lastmod: new Date(review.date).toISOString(),
+        changefreq: 'weekly',
+        priority: 0.7,
+      }
+    }),
+  ])
+}
```
joelhooks, line 34: this page is dynamic, cached with SWR and loads articles and reviews to add to the sitemap which is important for search crawlers

this *pattern* will apply to all of our sites

PR #269, opened 2022-06-10
Fixed OG headers on advanced-typescript and removed trailer

mattpocock:


joelhooks:
for some reason it didn't checkout the same branch so here's a new one https://github.com/skillrecordings/products/pull/270

PR #271, opened 2022-06-10
feat(convertkit): encapsulate the logic as a hooks

joelhooks:
hooked


packages/convertkit/src/forms/subscribe-to-convertkit/index.tsx
```diff
@@ -66,76 +35,43 @@ export const redirectUrlBuilder = (
...
   successMessage = <p>Thanks!</p>,
   actionLabel = 'Subscribe',
   onError = () => {},
   onSuccess = () => {},
-  subscribeApiURL = CONVERTKIT_SUBSCRIBE_API_URL,
+  subscribeApiURL,
   id,
   ...rest
 }) => {
-  const [isSubmitting, setSubmitting] = React.useState<boolean>(false)
```
joelhooks, line 80: formik actually tracks this state BUT your on submit handler has to return async value (promise)

https://formik.org/docs/api/formik#onsubmit-values-values-formikbag-formikbag--void--promiseany

below `handleOnSubmit` is async, but doesn't return (so it will be `undefined` I think?)

packages/convertkit/src/hooks/use-convertkit-form.tsx
```diff
@@ -0,0 +1,50 @@
+import {useFormik} from 'formik'
+import * as Yup from 'yup'
+import axios from 'axios'
+import {ConvertkitSubscriber} from '../types'
+import {
+  CONVERTKIT_SIGNUP_FORM,
+  CONVERTKIT_SUBSCRIBE_API_URL,
+} from '@skillrecordings/config'
+
+export function useConvertkitForm({
```
joelhooks, line 10: encapsulating the state and business logic of the component in this hook, which effectively makes it a simplified wrapper around formik's `useFormik` but implies a contract of required values and simplifies the options

packages/convertkit/src/hooks/use-convertkit-form.tsx
```diff
@@ -0,0 +1,50 @@
...
+      first_name: '',
+    },
+    validationSchema: Yup.object().shape({
+      email: Yup.string().email('Invalid email address').required('Required'),
+      first_name: Yup.string(),
+    }),
+    validateOnChange: false,
+    enableReinitialize: true,
+    onSubmit: async ({email, first_name}, {setStatus}) => {
+      return axios
```
joelhooks, line 34: returning this here is important so that the promise will resolve. Eliminated the need to independently track `isSubmitted`

packages/convertkit/src/index.ts
```diff
@@ -6,10 +6,10 @@ import {
 
 import fetchConvertkitSubscriberFromServerCookie from './utils/fetch-convertkit-subscriber'
 
-export {
-  SubscribeToConvertkitForm,
-  redirectUrlBuilder,
-} from './forms/subscribe-to-convertkit'
+export * from './forms/subscribe-to-convertkit'
+export * from './hooks/use-convertkit-form'
```
joelhooks, line 10: the hook gets exported now

packages/convertkit/src/forms/subscribe-to-convertkit/index.tsx
vojtaholik: good catch 👍

packages/convertkit/src/hooks/use-convertkit-form.tsx
vojtaholik: love it. shouldn't this be `.ts` instead of `.tsx` since it doesn't return a component?

PR #273, opened 2022-06-13
Accessibility fixes

marcysutton:
Added some various fixes to the site. There are two outstanding items:

- The Share link SVGs aren't getting `role="presentation"` to make them play nice with VoiceOver. I updated what I thought were the share links in this PR, but I'm not seeing a change to the markup? (see screenshot)

Twitter share icon in the Article page header

- The green focus outline on the green header doesn't meet the WCAG contrast requirement of 3:1. A white focus outline would work better on the green backgrounds.
  - It's also super hard to debug, since it doesn't match on `:focus` or `:focus-visible` in the DevTools and it blurs the element when I click into another contrast testing app like CCA. I had to take a screenshot of a focused element and then sample the colors on that. I guess it's the js-focus-visible technique that makes that difficult...?  How do we apply a different focus outline on a green background?
  
screenshot of focus outline on "All Articles"


vojtaholik:
https://github.com/skillrecordings/products/pull/275 updates visual focus indicator

PR #276, opened 2022-06-20
feat(testingaccessibility): module certificates api

vojtaholik:
This creates endpoint to generate module certificates using [pdfkit](https://github.com/alafr/SVG-to-PDFKit). I decided to make it part of the app since that makes it easy to use shared pieces of logic to determine access, etc.

**Usage example**
```jsx
<form action={`/api/certificate/modules/${module.slug}`}>
  <input name="name" type="text" />
  <button>Get Certificate</button>
</form>
```

The certificate currently looks like this, but I plan on updating its design to make it nicer.

certificate example

---

gif


apps/testingaccessibility/src/pages/api/certificate/modules/[slug].ts
```diff
@@ -0,0 +1,214 @@
...
+
+      const progress = userLessonProgress.map(serialize)
+
+      const {isCompleted} = getModuleProgressForUser(progress, module.sections)
+
+      if (!isCompleted) {
+        return res
+          .status(403)
+          .end(`You must complete the module in order to get a certificate.`)
+      }
```
Comment (vojtaholik) [inline apps/testingaccessibility/src/pages/api/certificate/modules/[slug].ts:87]: they actually need to complete the module

apps/testingaccessibility/src/pages/api/certificate/modules/[slug].ts
```diff
@@ -0,0 +1,214 @@
...
+
+      const moduleImage = await fetchImage(module.image.url)
+      const imageSVG = canvas
+        .image(moduleImage)
+        .width(320)
+        .height(320)
+        .attr('x', '36%')
+        .attr('y', '6%')
+
+      const userName = req.query.name || sessionToken.name || ''
```
Comment (vojtaholik) [inline apps/testingaccessibility/src/pages/api/certificate/modules/[slug].ts:109]: allow for passing any name but fallback to name from session token.

apps/testingaccessibility/src/pages/api/certificate/modules/[slug].ts
```diff
@@ -0,0 +1,214 @@
...
+},
+sections[]->{
+      title,
+      "slug": slug.current,
+      lessons[]->{
+        title,
+        "slug": slug.current
+      }
+}
+}`
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/certificate/modules/[slug].ts:208]: query's like this should probably be moved to the `lib` folder in a `modules.ts` or similar:

https://github.com/skillrecordings/products/blob/27357bbc57be3417bdf376ed71fc673f29e30616/apps/testingaccessibility/src/lib/articles.ts

basically ANY external api calls should be in the lib folder

joelhooks (approved):
this looks good, but I think we need to extract external fetching into the `lib` and potentially a couple of utils

PR #279, opened 2022-06-21
(feat: badass): add an articles section to badass

joelhooks:
I cut and pasted the articles stuff from testing accessibility and there is a big opportunity here to extract the components and sanity stuff into packages.

articles


apps/testingaccessibility/src/pages/[article].tsx
```diff
@@ -8,7 +8,7 @@ import find from 'lodash/find'
 import * as serverCookie from 'cookie'
 import groq from 'groq'
 import {CK_SUBSCRIBER_KEY} from '@skillrecordings/config'
-import checkSubscriber from 'utils/check-subscriber'
+import {checkIfConvertkitSubscriber} from '@skillrecordings/convertkit'
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/[article].tsx:11]: moved this into CK package

PR #289, opened 2022-07-02
chore: sort out the dependencies and align them across apps/packages

joelhooks:
Really struggled trying to upgrade to pnpm 7 and there are a lot of peer dependency issues in the repo. I noted that there was a lot of inconsistency in our versions of core packages, so this is trying to get them better aligned

I'm using pnpm v `6.32.25`

packages


mattpocock:
Had a look at the loom - here are some recommendations:

1. Create a new `@skillrecordings/eslint` package to capture all the eslint dependencies and make them consistent across all apps.

This will remove some overhead when managing this stuff in the future. [Turborepo has an example of this](https://github.com/vercel/turborepo/tree/main/examples/basic/packages/eslint-config-custom)

2. Move `@types/*` into the packages that require them.

Sorry, this just makes them more removable in the future.

3. Set up `manypkg check` to run on postinstall.

This will ensure that packages are always up to date, and that the good work you've done here doesn't get lost in the future. It'll also fail on CI.

Since you want to maintain a differing version of tailwind between packages, you should prefix all the `tailwindcss` imports with `npm:` (following [this advice](https://github.com/Thinkmill/manypkg#external-mismatch))

joelhooks:
@mattpocock we publish to npm and they are used in our egghead repo, which ultimately should probably just be part of the monorepo but currently isn't. does using peers this way have drawbacks outside of "managing them twice"? I'm not sure about pnpm's behavior here either. 

in the case of tailwind (and several of the other dependencies that have multiple versions...Framer maybe a couple others) we want to have a uniform versions and I just couldn't do the upgrade in this step

generally they worked fine prior to this change so I think next step will be to align those library's versions across the repos as well as satisfying pnpm unmet peer errors/warnings

.eslintrc.js
```diff
@@ -1 +1,10 @@
-module.exports = require('./packages/scripts/eslint-preset')
```
joelhooks, line 1: suggested by @mattpocock and I like this approach because it allows for more local configuration

apps/advanced-typescript/package.json
```diff
@@ -3,44 +3,43 @@
...
-  "prettier": {
-    "printWidth": 80,
-    "tabWidth": 2,
-    "useTabs": false,
-    "semi": false,
-    "singleQuote": true,
-    "trailingComma": "all",
-    "bracketSpacing": false,
-    "jsxBracketSameLine": false
+    "tailwindcss": "^3.1.4",
```
joelhooks, line 42: updated tailwind across the entire repo @vojtaholik

joelhooks (commented):
solved a lot of the unmet peer issues, though some remain and manypkg isn't totally appeased yet

it's ready for review though and once this is in we can finish the last details

vojtaholik (approved):

PR #296, opened 2022-07-06
feat(testingaccessibility): differentiate upgrades, new purchases, and logins

joelhooks:
different


apps/testingaccessibility/src/lib/purchases.ts
```diff
@@ -0,0 +1,89 @@
...
+        select: {
+          id: true,
+          name: true,
+        },
+      },
+    },
+  })
+
+  if (!purchase) {
+    return {}
```
joelhooks, line 37: if there's no purchase, bail out.

apps/testingaccessibility/src/lib/purchases.ts
```diff
@@ -0,0 +1,89 @@
+import prisma from '../db'
+
+export async function getPurchaseDetails(purchaseId: string, userId: string) {
```
joelhooks, line 3: this was in teams and welcome so moved it here to reduce duplication

apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -10,6 +10,53 @@ import {getSdk} from '../../../lib/prisma-api'
...
 import mjml2html from 'mjml'
 
+export type MagicLinkEmailType =
+  | 'login'
+  | 'signup'
+  | 'reset'
+  | 'purchase'
+  | 'upgrade'
+
+export const sendVerificationRequest = async (params: {
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:20]: this was previously inline below and couldn't be accessed. This meant our custom template wasn't being used! It also made it impossible to use alternative templates/subjects for different email types (new purchase, upgrade, login)

apps/testingaccessibility/src/pages/api/stripe/checkout.ts
```diff
@@ -181,6 +178,10 @@ export default withSentry(async function stripeCheckoutHandler(
         )
       }
 
+      const successUrl = isUpgrade
+        ? `${process.env.NEXT_PUBLIC_URL}/welcome?session_id={CHECKOUT_SESSION_ID}?upgrade=true`
+        : `${process.env.NEXT_PUBLIC_URL}/thanks/purchase?session_id={CHECKOUT_SESSION_ID}`
```
joelhooks, line 183: with an upgrade we know they are logged in because you can't upgrade so let's send them to welcome and not the thanks "check your inbox" page. The email still gets sent and they can click it to also get to welcome page.

below if they happen to hit welcome and aren't logged in, it forwards them to thanks as a fallback

apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -10,6 +10,53 @@ import {getSdk} from '../../../lib/prisma-api'
...
+
+  let subject
+
+  switch (params.type) {
+    case 'purchase':
+      subject = `Thank you for Purchasing Testing Accessibility (${host})`
+      break
+    default:
+      subject = `Log in to Testing Accessibility (${host})`
+  }
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:45]: I didn't change the template, but now we could use alternative html/text based on type which I think would be very nice.

apps/testingaccessibility/src/pages/welcome/index.tsx
```diff
@@ -9,111 +9,65 @@ import Layout from 'components/app/layout'
...
     })
 
-    if (!purchase) {
+    if (purchase) {
+      purchaseId = purchase.id
+    } else {
       return {
         redirect: {
-          destination: '/learn',
+          destination: `/thanks/purchase?session_id=${session_id}`,
```
joelhooks, line 40: here we have a session id, but couldn't find a purchase so we will bounce them over to /thanks and tell them to check inbox

apps/testingaccessibility/src/pages/welcome/index.tsx
```diff
@@ -9,111 +9,66 @@ import Layout from 'components/app/layout'
...
+  if (token && isString(purchaseId) && isString(token?.sub)) {
+    const {purchase, existingPurchase, availableUpgrades} =
+      await getPurchaseDetails(purchaseId, token.sub)
+    return purchase
+      ? {
+          props: {
+            purchase: serialize(purchase),
+            existingPurchase,
+            availableUpgrades,
+            upgrade: upgrade === 'true',
```
joelhooks, line 56: this is just a query param that we set in the checkout session callback

apps/testingaccessibility/src/pages/welcome/index.tsx
```diff
@@ -177,7 +133,8 @@ const Header: React.FC = () => {
         alt="a shining lighthouse"
       />
       <h1 className="font-bold lg:text-5xl sm:text-4xl text-3xl font-heading">
-        Welcome to Testing Accessibility
+        {upgrade ? `You've Upgraded ` : `Welcome to `}
+        {process.env.NEXT_PUBLIC_SITE_TITLE}
```
joelhooks, line 137: small change here to acknowledge what happened

apps/testingaccessibility/src/utils/send-server-email.ts
```diff
@@ -59,11 +65,12 @@ export async function sendServerEmail({
   const params = new URLSearchParams({callbackUrl, token, email: identifier})
   const _url = `${baseUrl}/api/auth/callback/${emailProvider.id}?${params}`
 
-  await emailProvider.sendVerificationRequest({
```
joelhooks, line 62: this always used the default send function so instead we export it so we are using our custom send

apps/testingaccessibility/src/pages/welcome/index.tsx
joelhooks: Screen Shot 2022-07-06 at 15 27 25

apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
Comment (joelhooks) [reply apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:45]: Screen Shot 2022-07-06 at 15 27 59

vojtaholik (approved):
nice!

PR #298, opened 2022-07-07
feat(testingaccessibility): add tag to user in CK and announce in slack

joelhooks:
announce


apps/testingaccessibility/src/server/post-sale-to-slack.ts
```diff
@@ -0,0 +1,37 @@
...
+            purchaseInfo.stripeProduct.name
+          } for ${`$${purchase.totalAmount}`}${
+            isEmpty(purchase.upgradedFromId) ? '' : ' as an upgrade'
+          }`,
+          color: '#eba234',
+          title: `Sold (${purchaseInfo.quantity}) ${purchaseInfo.stripeProduct.name}`,
+        },
+      ],
+      channel: process.env.SLACK_ANNOUNCE_CHANNEL_ID,
+    })
```
joelhooks, line 32: same as we do in egghead-rails but ported to JS

PR #300, opened 2022-07-08
feat(testingaccessibility): adds a status column to purchases

joelhooks:
this doesn't limit access based on these events but does annotate the purchase so that we will be able to do so easily.

status

### DIFF
--- a/apps/testingaccessibility/prisma/schema.prisma
+++ b/apps/testingaccessibility/prisma/schema.prisma
@@ -181,6 +181,13 @@ model UpgradableProducts {
   @@id([upgradableToId, upgradableFromId])
 }
 
+enum PurchaseStatus {
+    Valid
+    Refunded
+    Disputed
+    Banned
+}
+
 model Purchase {
   id                      String           @id @default(uuid())
   userId                  String?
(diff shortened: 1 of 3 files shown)


apps/testingaccessibility/prisma/schema.prisma
```diff
@@ -181,6 +181,13 @@ model UpgradableProducts {
   @@id([upgradableToId, upgradableFromId])
 }
 
+enum PurchaseStatus {
+    Valid
+    Refunded
+    Disputed
+    Banned
+}
```
joelhooks, line 189: we might have additional statuses later

apps/testingaccessibility/src/pages/api/stripe/webhook.ts
```diff
@@ -50,6 +53,14 @@ const stripeWebhookHandler = async (
...
         await postSaleToSlack(purchaseInfo, purchase)
 
+        res.status(200).send(`This works!`)
+      } else if (event.type === 'charge.refunded') {
+        const chargeId = event.data.object.id
+        await updatePurchaseStatusForCharge(chargeId, PurchaseStatus.Refunded)
+        res.status(200).send(`This works!`)
+      } else if (event.type === 'charge.dispute.created') {
+        const chargeId = event.data.object.id
+        await updatePurchaseStatusForCharge(chargeId, PurchaseStatus.Disputed)
```
joelhooks, line 63: when charges are disputed or refunded we want to mark them so so we can limit access as required

PR #302, opened 2022-07-08
feat(testingaccessibility): give the client the coupon details when active

joelhooks:
This will send the `defaultCoupon` when it is active for a product which includes the expiration datetime as well as some other details.

We might consider a "label" column for the coupon that we could use for the ribbon.

Build   Test Accessible Web Apps Testing Accessibility 2022-07-08 at 11 05 34

active

### DIFF
--- a/apps/testingaccessibility/src/components/pricing.tsx
+++ b/apps/testingaccessibility/src/components/pricing.tsx
@@ -27,6 +27,7 @@ type PricingProps = {
  * @param purchases
  * @param userId - If user is logged in, this is the user's ID.
  * @param index
+ * @param couponId
  * @constructor
  */
 export const Pricing: React.FC<PricingProps> = ({
@@ -37,19 +38,22 @@ export const Pricing: React.FC<PricingProps> = ({
   index,
   couponId,
 }) => {
-  const [coupon, setCoupon] = React.useState<{id: string; type: string}>()
+  const [merchantCoupon, setMerchantMerchantCoupon] = React.useState<{
+    id: string
+    type: string
+  }>()
   const [quantity, setQuantity] = React.useState(1)
   const debouncedQuantity: number = useDebounce<number>(quantity, 250)
   const {productId, name, image, modules, features, action} = product
 
   const {data: formattedPrice, status} = useQuery<FormattedPrice>(
-    ['pricing', coupon, debouncedQuantity, productId, userId, couponId],
+    ['pricing', merchantCoupon, debouncedQuantity, productId, userId, couponId],
     () =>
       fetch('/api/prices', {
         method: 'POST',
         body: JSON.stringify({
           productId,
-          ...(coupon && {coupon: coupon.id}),
+          ...(merchantCoupon && {merchantCouponId: merchantCoupon.id}),
           quantity,
           purchases,
           userId,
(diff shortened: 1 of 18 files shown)


joelhooks:
this opens the possibility for expiring single product coupons that also have a max use

so "60% off until X or until Y people buy"

apps/testingaccessibility/src/components/pricing.tsx
```diff
@@ -62,18 +62,22 @@ export const Pricing: React.FC<PricingProps> = ({
   )
 
   const availableCoupon = formattedPrice?.availableCoupons?.[0]
-  const appliedCoupon = formattedPrice?.appliedCoupon
+  const appliedMerchantCoupon = formattedPrice?.appliedMerchantCoupon
```
joelhooks, line 65: I tried to specify the `merchantCoupon` which is a database reference to the actual stripe coupon 

this is super confusing to me in practice because it requires thinking about which type of coupon we are currently using so being explicit in the name helps. the underlying idea is that we could have multiple merchants (ie paypal), but I'm not sure how that would happen in practice with a global coupon

it might actually be a "many to many" where instead of a "merchantCoupon" a `coupon` would have `merchantCoupons` plural and the final applied coupon would be used for the appropriate merchant

that's complicated future stuff and for now the explicit naming helped me

apps/testingaccessibility/src/pages/api/prices.ts
```diff
@@ -44,13 +44,13 @@ const pricesHandler = async (req: NextApiRequest, res: NextApiResponse) => {
...
-        productId,
-        spanContext,
-      })
+      const {activeMerchantCoupon, defaultCoupon} =
+        await getActiveMerchantCoupon({
+          siteCouponId,
+          code,
+          productId,
+          spanContext,
+        })
```
joelhooks, line 59: instead of just the merchant coupon id we get the whole coupon as well as any backing (default) coupon that we can pass along. It only exists if it is valid and associated with the active merchant coupon.

apps/testingaccessibility/src/pages/api/prices.ts
```diff
@@ -60,13 +66,15 @@ const pricesHandler = async (req: NextApiRequest, res: NextApiResponse) => {
...
+          merchantCouponId: activeMerchantCoupon
+            ? activeMerchantCoupon.id
+            : merchantCouponId,
           ...(upgradeFromPurchaseId && {upgradeFromPurchaseId}),
         },
         spanContext,
       )
 
-      res.status(200).json(product)
+      res.status(200).json({...product, ...(defaultCoupon && {defaultCoupon})})
```
joelhooks, line 77: send the coupon with the rules back to the client

apps/testingaccessibility/src/utils/get-active-merchant-coupon.ts
```diff
@@ -41,29 +43,37 @@ export async function getActiveCouponId({
...
+    activeMerchantCoupon = defaultMerchantCoupon
   }
 
-  return coupon ? coupon : couponId
+  const defaultCoupon = defaultCoupons?.defaultCoupon
+
+  return {
+    activeMerchantCoupon,
+    ...(defaultCoupon &&
+      defaultCoupon.merchantCouponId === activeMerchantCoupon?.id && {
```
joelhooks, line 75: only send back the default coupon (with rules) if it matches the active merchant coupon

apps/testingaccessibility/src/components/primary-newsletter-cta.tsx
```diff
@@ -27,8 +27,8 @@ export const PrimaryNewsletterCta: React.FC = () => {
         />
       </div>
       <h2 className="max-w-lg font-heading mx-auto -mt-4 sm:text-4xl text-3xl leading-none text-center md:text-5xl font-bold sm:mt-0">
-        Join my exclusive, free <span className="whitespace-nowrap">6-part</span>{' '}
-        email course
+        Join my exclusive, free{' '}
+        <span className="whitespace-nowrap">6-part</span> email course
```
joelhooks, line 31: formatter decided the space should be elsewhere

PR #311, opened 2022-07-10
fix(testingaccessibility): ppp checkout coupons now function as expected

joelhooks:
fix it

### DIFF
--- a/apps/testingaccessibility/src/components/pricing.tsx
+++ b/apps/testingaccessibility/src/components/pricing.tsx
@@ -11,6 +11,11 @@ import Image from 'next/image'
 import find from 'lodash/find'
 import cx from 'classnames'
 import {usePriceCheck} from '../context/pricing-check-context'
+import {first} from 'lodash'
+
+function getFirstPPPCoupon(availableCoupons: any[] = []) {
+  return find(availableCoupons, (coupon) => coupon.type === 'ppp') || false
+}
 
 type PricingProps = {
   product: SanityProduct
@@ -39,7 +44,7 @@ export const Pricing: React.FC<PricingProps> = ({
   index,
   couponId,
 }) => {
-  const [merchantCoupon, setMerchantMerchantCoupon] = React.useState<{
+  const [merchantCoupon, setMerchantCoupon] = React.useState<{
     id: string
     type: string
   }>()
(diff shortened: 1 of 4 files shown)


apps/testingaccessibility/src/components/pricing.tsx
```diff
@@ -72,20 +77,28 @@ export const Pricing: React.FC<PricingProps> = ({
...
-  )
+
+  const pppCoupon = getFirstPPPCoupon(formattedPrice?.availableCoupons)
+
+  // if there is no available coupon, hide the box (it's not a toggle)
+  // only show the box if ppp coupon is available
+  // do not show the box if purchased
+  // do not show the box if it's a downgrade
   const showPPPBox =
+    availableCoupon &&
```
joelhooks, line 101: explained in the comment, but this wasn't meant to act as a toggle. Generally we only show the PPP discount if the discount is more than the active discount, so they shouldn't need to decide one way or the other, but page refresh is the "toggle". Might make sense to use a different UI element?

either way, it should go way and not linger as a choice.

apps/testingaccessibility/src/pages/api/prices.ts
```diff
@@ -50,13 +50,21 @@ const pricesHandler = async (req: NextApiRequest, res: NextApiResponse) => {
...
-        })
+      // explicit incoming merchant coupons are honored
+      // without checking for other potential coupons
+      // if there is no explicit incoming merchant coupon
+      // we check for default/global coupon or an incoming code
+      const {activeMerchantCoupon, defaultCoupon} = merchantCouponId
+        ? {
+            activeMerchantCoupon: {id: merchantCouponId},
+            defaultCoupon: false,
+          }
```
joelhooks, line 61: if there is an explicitly applied coupon we can just skip the lookup and save some time on the response

apps/testingaccessibility/src/utils/format-prices-for-product.ts
```diff
@@ -198,7 +198,8 @@ export async function formatPricesForProduct(
         noContextOptions,
       )
 
-    const {id, ...merchantCouponWithoutIdentifier} = appliedMerchantCoupon
+    const {identifier, merchantAccountId, ...merchantCouponWithoutIdentifier} =
```
joelhooks, line 201: this is what was causing the issue. in our app `identifier` is the Stripe coupon code ID and we are sanitizing it before sending it to the client, but I changed this to `id` at some point and that broke the coupon getting sent to stripe properly for PPP

added tests above

joelhooks:
here's the commit that introduces the bug https://github.com/skillrecordings/products/commit/9ba59068e51cfe91623c1fb909ed8c9f5b79c660#diff-067296bf0071cd8352540daaab77ccbe0a328f7613b735857d24b92c0668f455R199

Issue #305, opened 2022-07-09
[Testing Accessibility] Accessibility of Course Certificate PDF


marcysutton:
We need to ensure the course certificates are as accessible as can be. For PDF documents, this means enabling tagging to expose headings, alt text, and whatever else can be exposed accessibly.

Adobe Acrobat has an Accessibility Checker built in that can help highlight issues. My question is: what control do we have over the PDFs that get generated?

Example certificate: [Testing Accessibility Certificate - Marcy Sutton.pdf](https://github.com/skillrecordings/products/files/9077830/Testing.Accessibility.Certificate.-.Marcy.Sutton.pdf)

Here are some tips: https://helpx.adobe.com/acrobat/using/create-verify-pdf-accessibility.html

Here's the report:

[...]

PR #312, opened 2022-07-10
feat(testingaccessibility): #305 certificate behind feature flag default

joelhooks:
turn it off


apps/testingaccessibility/src/pages/learn/index.tsx
```diff
@@ -223,7 +226,8 @@ const Learn: React.FC<{purchases: Purchase[]; product: SanityDocument}> = ({
                           )
                         })}
 
-                        {sections?.length > 0 && (
+                        {process.env.NEXT_PUBLIC_CERTIFICATE_ENABLED &&
```
joelhooks, line 229: by default this env doesn't exist so they are off until it is defined (value doesn't matter)

vojtaholik (approved):

PR #313, opened 2022-07-10
feat(testingaccessibility): ability to generate various percentage off coupons

joelhooks:
discount

### DIFF
--- a/apps/testingaccessibility/src/pages/api/admin/coupon.ts
+++ b/apps/testingaccessibility/src/pages/api/admin/coupon.ts
@@ -5,7 +5,28 @@ import {setupHttpTracing} from '@vercel/tracing-js'
 import {tracer} from 'utils/honeycomb-tracer'
 import {getToken} from 'next-auth/jwt'
 import prisma from '../../../db'
+import {getSdk} from '../../../lib/prisma-api'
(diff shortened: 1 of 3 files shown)


apps/testingaccessibility/src/pages/api/admin/coupon.ts
```diff
@@ -5,8 +5,29 @@ import {setupHttpTracing} from '@vercel/tracing-js'
...
 const ROLES_WITH_ACCESS = ['ADMIN', 'SUPERADMIN']
 
+/**
+ * @link https://www.gavsblog.com/blog/find-closest-number-in-array-javascript
+ * @param percentOff
+ */
+const findClosestDiscount = function (percentOff: number) {
+  // we want a fraction so if it is whole number, we make it fractional
+  percentOff = percentOff <= 1 ? percentOff : percentOff / 100
+  return [1, 0.95, 0.9, 0.75, 0.6, 0.5, 0.4, 0.25, 0.1].reduce((a, b) => {
```
joelhooks, line 18: discounts are clamped to this range, but you can pass in any number and it'll pick the closest. Numbers over 1 will be divided by 100 to get a fraction (or lock it at 100% off if it's bigger than that)

PR #317, opened 2022-07-11
update(testingaccessibility): code blocks

### LINKED ISSUE #308: [Testing Accessibility] Accessibility tweaks to code blocks
The refractor code blocks could use a couple of tweaks to be more accessible. Do we have the ability to add props to that component?

To the rendered `<pre>` element wrapping a code block, I'd like to add:

- `role="region"` to expose a landmark role
- `aria-label={nameOfCodeBlock}` to name that landmark
- `tabIndex="0"` to make it focusable/scrollable

I borrowed these suggestions from the Cupper project: https://thepaciellogroup.github.io/cupper/patterns/coding/code-blocks/

A code block highlighted with Chrome DevTools open

vojtaholik:
fixes #308

gif

### DIFF
--- a/apps/testingaccessibility/src/components/portable-text/index.tsx
+++ b/apps/testingaccessibility/src/components/portable-text/index.tsx
@@ -220,7 +220,7 @@ const PortableTextComponents: PortableTextComponentsProps = {
           <VideoProvider>
             <Video url={url} title={title} />
           </VideoProvider>
-          <figcaption tabIndex={0}>
+          <figcaption>
             <details className="group marker:text-transparent no-marker">
               <summary className="inline-flex space-x-2 items-center cursor-pointer text-gray-600 hover:text-gray-800 transition">
                 <span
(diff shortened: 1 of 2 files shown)


apps/testingaccessibility/src/components/portable-text/index.tsx
```diff
@@ -242,20 +242,32 @@ const PortableTextComponents: PortableTextComponentsProps = {
...
-                : 'javascript'
-              : 'javascript'
-          }
-          value={code}
-          markers={highlightedLines}
-        />
+        <>
+          <pre
+            role="region"
+            aria-label={label ? label : 'code sample'}
```
vojtaholik, line 250: code block label in sanity studio

apps/testingaccessibility/src/components/portable-text/index.tsx
```diff
@@ -242,20 +242,32 @@ const PortableTextComponents: PortableTextComponentsProps = {
...
+            tabIndex={0}
+            className="sr-only"
+          >
+            <code>{code}</code>
+          </pre>
+          <pre aria-hidden="true">
+            <Refractor
```
vojtaholik, line 257: @marcysutton I noticed that screen readers read the whole code as a single sentence/string on the Cupper project, whereas we had lots of pauses due to span tags. This change applies the same approach.

PR #318, opened 2022-07-12
feat(typescriptcourse): seo for articles

joelhooks:
- [x] text get the CTA working using portable [loom - too long]( https://www.loom.com/share/2279ae14c2b74e708a372b4f058d49b9)
- [x] get the sitemaps working [loom, much shorter](https://www.loom.com/share/55b7efad8693453ba0267a78c23f1682)
- [x] add an rss feed [loom](https://www.loom.com/share/5f57cba3f8344f3db9254da0aff28e80)
- [x] style cta form to match site

Screen Shot 2022-07-13 at 20 56 16

seo


apps/testingaccessibility/src/templates/article-template.tsx
```diff
@@ -45,6 +45,7 @@ const ArticleTemplate: React.FC<ArticleTemplateProps> = ({
           publishedTime: date,
         },
         ogImage,
+        url: `${process.env.NEXT_PUBLIC_URL}${router.asPath}`,
```
joelhooks, line 48: my guess is our canonicals were wrong for TA

apps/typescriptcourse/package.json
```diff
@@ -75,7 +76,7 @@
     "mdast-util-to-markdown": "^1.2.6",
     "mdx-prism": "^0.3.4",
     "next": "^12.1.6",
-    "next-seo": "^4.29.0",
+    "next-seo": "^5.4.0",
```
joelhooks, line 79: this is aggravating, but I had to switch back to 5.4 for this project

apps/typescriptcourse/src/components/app/layout.tsx
```diff
@@ -6,7 +6,13 @@ import config from 'config'
...
 
 export type LayoutProps = {
-  meta?: any
+  meta?: {
+    title?: string
+    description?: string
+    titleAppendSiteName?: boolean
+    url?: string
+    ogImage?: {url: string; alt?: string}
+  }
```
joelhooks, line 15: typing this surface a few bugs where a more expanded meta prop was being passed in, it's noted below

apps/typescriptcourse/src/templates/portable-text-article-template.tsx
```diff
@@ -33,18 +33,27 @@ type ArticleTemplateProps = {
   hasSubscribed: boolean
 }
 
-const ArticleTemplate: React.FC<ArticleTemplateProps> = ({
+const PortableTextArticleTemplate: React.FC<ArticleTemplateProps> = ({
```
joelhooks, line 36: this is a more descriptive name for what this component is doing

apps/typescriptcourse/src/templates/portable-text-article-template.tsx
```diff
@@ -58,14 +67,37 @@ const ArticleTemplate: React.FC<ArticleTemplateProps> = ({
...
+                    const redirectUrl = redirectUrlBuilder(
+                      subscriber,
+                      '/confirm',
+                    )
+                    router.push(redirectUrl)
+                  }
+                }}
+                actionLabel={cta.actionLabel}
+              />
+            </CTAContainer>
```
joelhooks, line 92: below is a solid apparatus for switching CTAs, but I want to put a little more thought into it so I simplified down to just the one CTA that isn't smart around the user being a subscriber or not

one thing that's interesting is that this section doesn't need to be server rendered and we can do client side decision making loading it dynamically

if we want to hide half the article, we should consider middleware and rewrites IMO

apps/typescriptcourse/src/templates/root.tsx
```diff
@@ -27,8 +27,10 @@ const HomeTemplate: React.FC<EmailCourseTemplateProps> = ({meta, children}) => {
     }
   }, [router])
 
+  // TODO fix meta for layout
+
   return (
-    <Layout meta={meta} className="flex flex-col min-h-screen">
+    <Layout className="flex flex-col min-h-screen">
```
joelhooks, line 33: this was a chonky meta without the proper SEO bits

apps/typescriptcourse/studio/schemas/documents/article.js
```diff
@@ -63,31 +63,7 @@ export default {
...
-                value: '3122425',
-              },
-              {
-                title: 'workshop-form-manually-a11y-testing (2662749)',
-                value: '2662749',
-              },
-              {title: 'workshop-form-people (3122421)', value: '3122421'},
-              {title: 'workshop-form-semantics (3122422)', value: '3122422'},
-            ],
+            list: [{title: 'Email Course (3071922)', value: '3071922'}],
```
joelhooks, line 66: having these hardcoded in sanity config works, but we might make them dynamic in the Sanity database instead.

joelhooks:
the failing builds above were "fixed" by switching to node 16 on vercel

PR #320, opened 2022-07-12
feat(protailwind): quiz

vojtaholik:
screenshot

gif


vojtaholik:
@joelhooks I keep getting this error on submit which seems to be related to `eggheadAxios`. I tried skipping the `fetchEggheadUser` call in `/api/answer` but that didn't help. No clue where it's coming from, but I think we should add a check for `AUTH_CLIENT_ID` env var or similar.

Pasted Graphic

apps/protailwind/src/pages/answer.tsx
```diff
@@ -35,12 +37,12 @@ export async function getStaticProps() {
     welcome: {
       question: `From the list in the email, what is a new workflow step or approach you could take to evaluate websites or applications for accessibility?`,
       type: `essay`,
-      tagId: 2304869, // ec - ta - 001 Welcome Completed
+      tagId: CK_TAG_TEST, // ec - ta - 001 Welcome Completed
```
joelhooks, line 40: these were failing because they don't exist in PT CK

packages/quiz/src/utils/submit-answer.ts
```diff
@@ -9,10 +9,20 @@ export default async function handleSubmitAnswer(context: QuizContext) {
...
   } = context
+
+  // TODO check Sanity API structure versus static JSON structure
+  //      they should be the same! It appears as though that might not be the case
+  let id =
+    findKey(questionSet, currentQuestion) ||
+    findKey(
+      questionSet,
+      (question) => question.question === currentQuestion.question,
+    )
```
joelhooks, line 20: the id was parsing as `undefined` so I was curious if there was a disparity between Sanity and JSON data formats. If so, we should align them.

turbo.json
```diff
@@ -17,6 +17,17 @@
         "index.js"
       ]
     },
+    "@skillrecordings/config#build": {
+      "dependsOn": [
+        "^build",
+        "$CONVERTKIT_API_SECRET",
+        "$NEXT_PUBLIC_CONVERTKIT_TOKEN",
+        "$NEXT_PUBLIC_CONVERTKIT_SIGNUP_FORM"
+      ],
```
joelhooks, line 26: this is something that I need to spend more time with to understand more completely. Generally I think we need to remove all of these vars from ALL packages and instead pass them in to the package as config from the apps

https://turborepo.org/docs/core-concepts/caching#alter-caching-based-on-environment-variables-and-files

that's the advice from the docs and it makes sense

packages/quiz/src/utils/submit-answer.ts
joelhooks: https://github.com/skillrecordings/products/blob/876fc0ac6a52e0c77571d161dec38bd033a137a4/apps/cssdestructured/src/pages/answer.tsx#L28-L50

joelhooks: here's where it gets transformed:

https://github.com/skillrecordings/products/blob/acf8ceae4f2e969794fc83e21875ef8bce5a5ff3/packages/quiz/src/utils/transform-sanity-quiz.ts#L10-L29

PR #321, opened 2022-07-12
feat(testingaccessibility): sale countdown

vojtaholik:
banner
plan countdown

gif


apps/testingaccessibility/src/components/pricing.tsx
```diff
@@ -231,6 +232,12 @@ export const Pricing: React.FC<PricingProps> = ({
             </fieldset>
           </form>
         )}
+        {!merchantCoupon && (
+          <SaleCountdown
+            coupon={formattedPrice?.defaultCoupon}
+            data-pricing-product-sale-countdown={index}
+          />
+        )}
```
joelhooks, line 240: logically speaking I'd consider checking for the presence of the default coupon and not the absence of the merchant coupon here. same basic effect, but "positive" logic and would be more accurate. the "data" attribute here might use an alias at the prop level

apps/testingaccessibility/src/components/sale-banner.tsx
```diff
@@ -0,0 +1,108 @@
...
+      body: JSON.stringify({
+        productId,
+      }),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    })
+      .then((res) => res.json())
+      .then((formattedPrice: FormattedPrice) => formattedPrice),
+  )
```
joelhooks, line 29: now that I see this we definitely want to rethink the pricing context

apps/testingaccessibility/src/pages/index.tsx
```diff
@@ -24,7 +24,7 @@ const Home: React.FC<CommerceProps> = ({
 }) => {
   const {redeemableCoupon, RedeemDialogForCoupon} = useCoupon(couponFromCode)
   return (
-    <Layout className="bg-white">
+    <Layout className="bg-white" products={products}>
```
joelhooks, line 27: not a big fan of layering state on a layout component like this and think that we need to take products/pricing up to the app as a context, will keep thinking on this one

PR #322, opened 2022-07-12
update(sale banner): no need to pass products + improve a11y

vojtaholik:
gif


apps/testingaccessibility/src/components/sale-banner.tsx
```diff
@@ -1,25 +1,18 @@
...
-  const productId = get(first(products), 'productId')
+const SaleBanner = () => {
+  const router = useRouter()
   const shouldReduceMotion = useReducedMotion()
   const {data: formattedPrice} = useQuery<FormattedPrice>(['banner'], () =>
     fetch('/api/prices', {
       method: 'POST',
-      body: JSON.stringify({
-        productId,
-      }),
```
vojtaholik, line 22: turns out the `productId` is not needed in order to return a coupon 🤔

PR #324, opened 2022-07-12
chore: update next-seo to 5.4.0 in all apps

vojtaholik:
gif


joelhooks (approved):

joelhooks:
This is my hopefully temporary solution for https://github.com/garmeeh/next-seo/issues/966

It's not clear what the root of the issue is. I suspected it had something to do with our package and library structure in the monorepo, but couldn't isolate it so went brute force and cut and pasted it the next-seo code into the project

so far that's worked and has stabilized our meta tags

will keep investigating for an actual solution

PR #331, opened 2022-07-14
chore(testingaccessibility): move generated prisma client location

joelhooks:
This changes the location of the generated prisma client and moves it out of node modules. I found that because of the way that pnpm symlinks this can cause some issues in a monorepo, particularly during CI builds where all the apps are building at once.

@vojtaholik I suspect this is where the issues arose with the enums, but I'm not positive.

migrate


apps/engmanagement/src/pages/api/skill/[...skillRecordings].ts
```diff
@@ -5,7 +5,7 @@ import SkillRecordings, {
 import prisma from '../../../db'
 
 export const skillOptions: SkillRecordingsOptions = {
-  prismaClient: prisma,
+  prismaClient: prisma as any,
```
Comment (joelhooks) [inline apps/engmanagement/src/pages/api/skill/[...skillRecordings].ts:8]: not super stoked on the any use, but it works and on the `skill-api` side we expect the client to match and it does

it'd be nice to have the client generated centrally, but I think it is built around a specific `DATABASE_URL`. Need more research there.

laurosilvacom (approved):

PR #335, opened 2022-07-15
feat: add videoResource to testing accessibility sanity schema

Creeland:
This completes the schema side of things for migrating videos from inline objects to their own documents.

Next step is to write a script migrating existing data over to these documents, and also making sure to replace the inline objects with refs pointed to the correct documents. 

script


vojtaholik (approved):
I haven't tested the migration script, but looks good overall! I assume we'll want to run it against test dataset first.

PR #336, opened 2022-07-15
Add a newsletter page with signup form to TA

tayiorbeii:
* Adds a newsletter page at `/newsletter`
* Adds "Newsletter" to footer (screenshot shows prior "Newsletter Signup" text)
* Uses the PrimaryCTA component with Email course text
* Will display "It looks like you're already signed up!" if they are recognized

preview

Homer is intrigued, and wishes to subscribe to the newsletter

### DIFF
--- a/apps/testingaccessibility/src/components/app/footer.tsx
+++ b/apps/testingaccessibility/src/components/app/footer.tsx
@@ -7,13 +7,14 @@ import Link from 'next/link'
 import toast from 'react-hot-toast'
 import {isSellingLive} from 'utils/is-selling-live'
 import {useNavState} from '../../hooks/use-nav-state'
+import {useConvertkit} from '@skillrecordings/convertkit'
 
 type FooterProps = {}
 
 const Footer: React.FC<FooterProps> = () => {
   const router = useRouter()
   const {isSignedIn, canViewInvoice, canViewTeam} = useNavState()
-
+  const {subscriber} = useConvertkit()
   const {setIsFeedbackDialogOpen} = useFeedback()
   return (
     <footer className="font-nav">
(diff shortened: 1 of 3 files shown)


apps/testingaccessibility/src/components/app/footer.tsx
```diff
@@ -39,6 +39,9 @@ const Footer: React.FC<FooterProps> = () => {
                 <li>
                   <NavLink href="/articles">Articles</NavLink>
                 </li>
+                <li>
+                  <NavLink href="/newsletter">Newsletter</NavLink>
+                </li>
```
vojtaholik, line 44: suggestion: This could be conditional based on whether `ck_subscriber_id` is present.

apps/testingaccessibility/src/pages/newsletter.tsx
```diff
@@ -0,0 +1,30 @@
...
+const NewsletterPage = () => {
+  const title = 'Newsletter'
+  const ogImage = getOgImage(title)
+  const {subscriber, loadingSubscriber} = useConvertkit()
+  return (
+    <Layout meta={{title, ogImage}} className="w-full">
+      <main className="bg-green-700 bg-noise text-white w-full mx-auto py-10 lg:pb-32 sm:pb-20">
+        <h1 className="text-center font-extrabold lg:text-4xl sm:text-3xl text-2xl font-heading">
+          {title}
+        </h1>
```
vojtaholik, line 15: I don't think this is needed, the `PrimaryNewsletterCta` could take a children which would let us use the "Join my exclusive, ..." as h1. The page would look nicer also.

vojtaholik (approved):

PR #341, opened 2022-07-18
fix(testingaccessibility): client side redirect for magic links ❇️ 

joelhooks:
This is to address this issue: https://github.com/nextauthjs/next-auth/issues/4965

Basically, it's a relatively lame attempt to stop aggressive enterprise email scanning systems from pre-clicking login links before a human ever sees them.

Details in source tab...

redirect


apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -30,9 +30,12 @@ export const sendVerificationRequest = async (params: {
     url,
     provider: {server, from},
   } = params
-  const {host} = new URL(url)
+  const rewriteUrl = new URL(url)
   const transport = createTransport(server)
   const {getUserByEmail} = getSdk()
+  const {host} = rewriteUrl
+
+  rewriteUrl.pathname = '/api/callback'
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:38]: rewrites the url going to the inbox with our userland route that does a client-side redirect

apps/testingaccessibility/src/pages/api/callback.ts
```diff
@@ -0,0 +1,38 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Testing Accessibility Login Redirect</title>
+    <script>
+      const url = new URL(window.location.href)
+      url.pathname = "/api/auth/callback/email"
+      window.location.href = url.href
```
joelhooks, line 18: rewrite the URL back to the default next-auth email callback route and send them on their way

this assumes most of these bots aren't executing JS, which is a big assumption, but is likely to catch at least a good number of systems that attempt this sort of thing

usually these systems are protecting against "detonation" attacks where clicking a link causes immediate harm. This "trick" here seems like a pretty low-rent hack against that sort of high-end protection 🤡

might not even be worth it?

PR #342, opened 2022-07-18
feat(testingaccessibility): show nav faster :zap:

joelhooks:
faster

### DIFF
--- a/apps/testingaccessibility/src/components/app/navigation.tsx
+++ b/apps/testingaccessibility/src/components/app/navigation.tsx
@@ -12,6 +12,7 @@ import cx from 'classnames'
 import Image from 'next/image'
 import {useNavState} from '../../hooks/use-nav-state'
 import {useFeedback} from '../../context/feedback-context'
+import {useUser} from '../../hooks/use-user'
 
 const Navigation = () => {
   return (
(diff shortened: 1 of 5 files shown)


apps/testingaccessibility/src/hooks/use-user.ts
```diff
@@ -0,0 +1,15 @@
...
+    async ({queryKey}) => {
+      return await fetch(queryKey[0]).then((res) => res.json())
+    },
+  )
+
+  return {
+    userLoadingStatus,
+    user: data,
+  }
+}
```
joelhooks, line 15: calling this endpoint directly seems faster in my testing and gives us immediate access to the current user data that we want and can use to generate an ability set

apps/testingaccessibility/src/hooks/use-nav-state.ts
```diff
@@ -1,14 +1,14 @@
...
 /**
  * TODO: consider removing this and using abilities?
  */
 export function useNavState() {
-  const {status, ability} = useCurrentAbility()
+  const {ability} = useCurrentAbility()
 
   return {
-    isSignedIn: status === 'authenticated',
-    isLoadingUser: status === 'loading',
```
joelhooks, line 11: the loading aspect here was the only thing that wasn't an ability related shortcut, so I moved it out since passing it down seemed redundant

apps/testingaccessibility/src/components/app/navigation.tsx
```diff
@@ -125,7 +128,7 @@ const MobileNav: React.FC = () => {
                   )}
                 </Menu.Item>
               </div>
-              {!isLoadingUser && isSellingLive && (
+              {userLoadingStatus !== 'loading' && isSellingLive && (
```
joelhooks, line 131: checking the status directly instead of using the helper since "what can they do" is different than "when can they do it"

PR #348, opened 2022-07-19
feat: add a next template to use with create-next-app

joelhooks:
This is a candidate for a product template to use for a new product.

https://nextjs.org/docs/api-reference/create-next-app

@vojtaholik some of the basic CSS and theme configuration is interesting to consider and might be best served as ENV variables for primary color, font, etc since those are available in tailwind, next, etc config files.

We tried to strip away anything opinionated and put as at the "ready to style and create an email course with a quiz" point in the apps evolution with basic structure ready for articles.

This doesn't include the Sanity template, which is a different project -> https://www.sanity.io/guides/creating-a-sanity-and-vercel-1-click-starter

```bash
npx create-next-app@latest --ts --use-pnpm \
 --example https://github.com/skillrecordings/products/tree/023743b896f7c0c203613931454a101676a25fca/apps/skill-next-product-template
```

to try it out with this branch (the above is locked to a specific commit)

It'd probably be more interesting/robust if we had our own `create-skill-app` that extended/forked this: https://github.com/vercel/next.js/blob/canary/packages/create-next-app/create-app.ts 🤔

cookie cutter


vojtaholik:
> some of the basic CSS and theme configuration is interesting to consider and might be best served as ENV variables for primary color, font, etc since those are available in tailwind, next, etc config files.

This is definitely interesting. I'm not sure how far we'd like to take it, so here are few options and considerations:
- If we wanted to take advantage of Tailwind color scales (50—900), we could have something like this in config: `primary: colors[process.env.NEXT_PUBLIC_PRIMARY_COLOR]`. That would only work if primary color would be one of tailwind's default ones. Otherwise it fallbacks to black.
- I don't see much potential in having just a single primary color (#E22CC8) defined and think I would end up extending it even for very basic landing pages.
- Color themes: Simon goes through the process of setting it up in his ProTailwind email course. It's pretty cool and also more robust since it uses tailwind's plugin api... Although I'm not entirely sure how well it would fit in this context at the moment.

Overall I would say that in most cases we would end up needing to define more than one color. `text-base, background-base, text-inverted, background-inverted`. I'm also not sure how I feel about color being "primary".

For fonts 
- We could pass in google fonts url and a font name
- Other option for most basic use case would be passing in 'serif' or 'sans-serif' and fallback to system fonts
- No matter which option we'd take, it would end up making a not very transparent decision about where the font is being used (headings, body copy, etc)

vojtaholik:
does `create-next-app` script work for you? I'm running into following error:
``` bash
ERR_PNPM_NO_MATCHING_VERSION_INSIDE_WORKSPACE  In : No matching version found for @skillrecordings/scripts@* inside the workspace
```

PR #352, opened 2022-07-20
feat(testingaccessibility): local dev mysql :zap:

joelhooks:
set your database url env var in your `.env` file::

```env
DATABASE_URL="mysql://root@localhost:3306/testing-accessibility"
```

within the root of `/apps/testingaccessibility` run the following:

```bash
docker compose up -d 
pnpm db:studio
```

locally


tayiorbeii (approved):
At first I was getting a Prisma error that it couldn't find my `DATABASE_URL` env variable.


I was able to fix this by running `npx prisma generate`, then globally installing the dotenv-cli and manually loading `.env.local` before running the Prisma studio (shoutout Stack Overflow comments)
`npm install -g dotenv-cli`

PR #357, opened 2022-07-21
totaltypescript.com using new template

vojtaholik:
to do:
- [x] style header
- [x] style landing page copy ([notion doc](https://www.notion.so/totaltypescript/Landing-Page-Copy-5271f58cf90a4e7eb42d621f1b26e80e))
- [x] style subscribe form
- [x] style confirm and confirmed pages
- [x] make it look nice on mobile too
- [x] make a share card

share card
landing page
confirm page

gif


apps/total-typescript/package.json
```diff
@@ -3,48 +3,63 @@
...
     "@types/react-dom": "^18.0.6",
     "autoprefixer": "^10.4.2",
     "eslint": "^8.19.0",
     "eslint-config-custom": "workspace:*",
+    "lint-staged": "^11.2.6",
     "postcss": "^8.4.14",
-    "tailwindcss": "^3.1.4",
-    "typescript": "^4.7.4"
+    "prettier": "^2.6.2",
+    "tailwindcss": "^3.1.6",
```
joelhooks, line 50: notice this got bumped, we want to make sure this is consistent across the entire repo, generally so I do a whole project search and replace when I bump a version of a major lib

apps/total-typescript/package.json
```diff
@@ -3,48 +3,63 @@
...
+    "webpack": ">=4.3.0 <5.0.0"
+  },
+  "lint-staged": {
+    "*.{js,json,md,mdx,ts,tsx}": [
+      "prettier --write"
+    ],
+    "src/**/*.{ts,tsx,md,mdx,js}": [
+      "prettier --write"
+    ]
+  },
```
joelhooks, line 61: this is all handled from the top level I believe so doesn't need to be in this package

apps/total-typescript/prettier.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  bracketSpacing: false,
+  jsxBracketSameLine: false,
+  printWidth: 80,
+  semi: false,
+  singleQuote: true,
+  tabWidth: 2,
+  trailingComma: 'all',
+  useTabs: false,
+}
```
joelhooks, line 10: all of our projects in this repo stash this in package.json, if we prefer it here we need to update all of them and the template

apps/total-typescript/process.d.ts
```diff
@@ -0,0 +1,18 @@
...
+    NEXT_PUBLIC_PARTNER_TWITTER: string
+    NEXT_PUBLIC_SEO_KEYWORDS: string
+    NEXT_PUBLIC_PRODUCT_DESCRIPTION: string
+    NEXT_PUBLIC_CONVERTKIT_SIGNUP_FORM: string
+    NEXT_PUBLIC_CONVERTKIT_TOKEN: string
+    NEXT_PUBLIC_CONVERTKIT_SUBSCRIBER_KEY: string
+    CONVERTKIT_BASE_URL: string
+    NEXT_PUBLIC_GOOGLE_ANALYTICS: string
+  }
+}
```
joelhooks, line 18: if these are all required because of IDE/compile errors, can they get added to the template?

joelhooks (commented):
added a couple procedural monorepo comments

apps/total-typescript/package.json
vojtaholik: reverted 👍

apps/total-typescript/package.json
vojtaholik: I noticed this is the case for more apps

apps/total-typescript/prettier.config.js
vojtaholik: leftover debris from the original version. got it moved to package.json. 👍

apps/total-typescript/package.json
joelhooks: ya I removed it from the template recently since it's handled at the root level

we can triple check it, doesn't really matter in practice unless it doesn't work at all with this removed. then we need to update template

joelhooks (approved):

PR #364, opened 2022-07-22
Hide article dates on TA

tayiorbeii:
* Changes article page & template date formatting from `22 June, 2022` to `June 22, 2022`
* Hide publish dates from articles index and pages

Before:
image

After:
image

skeleton arm pulling calendar pages


joelhooks:
if this PR can get some "why" added we can reopen/merge it. Learners love dates. It is also confusing why we'd change the date format and then comment it out. Why not just remove it?

PR #366, opened 2022-07-22
feat(create-skill-app): template config files with handlebars

joelhooks:
config


apps/testingaccessibility/package.json
```diff
@@ -95,7 +95,6 @@
     "classnames": "^2.3.1",
     "client-oauth2": "^4.3.3",
     "cookie": "^0.5.0",
-    "crypto": "^1.0.1",
```
joelhooks, line 98: this package is part of standard lib now

packages/create-skill-app/index.ts
```diff
@@ -47,6 +48,75 @@ async function run(): Promise<void> {
...
+    },
+    {
+      type: 'text',
+      name: 'nextAuthDevSecret',
+      message: `next-auth secret (only for dev)`,
+      initial: crypto.randomBytes(32).toString('hex'),
+    },
+  ]
+
+  const projectData = await prompts(questions)
```
joelhooks, line 118: makes it ez to add additional questions and the `prompts` lib has a lot of flexible options. If you just mash enter all the way through everything works fine.

PR #369, opened 2022-07-26
feat: using custom prisma adapter to modify next-auth token behavior

joelhooks:
expired


apps/testingaccessibility/prisma/schema.prisma
```diff
@@ -234,6 +234,7 @@ model User {
 model VerificationToken {
   token      String    @id
   identifier String
-  expires    DateTime? 
+  expires    DateTime?
+  createdAt  DateTime?          @default(now())
```
joelhooks, line 238: adding a createdAt so we can calculate a distance

apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -30,12 +30,9 @@ export const sendVerificationRequest = async (params: {
     url,
     provider: {server, from},
   } = params
-  const rewriteUrl = new URL(url)
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:33]: not using the rewrite scheme now

apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,14 +1,14 @@
...
 import jwt from 'jsonwebtoken'
-import {PrismaAdapter} from '@next-auth/prisma-adapter'
 import prisma from '../../../db'
 import {createTransport} from 'nodemailer'
 import {withSentry} from '@sentry/nextjs'
 import {getSdk} from '../../../lib/prisma-api'
 import {defineRulesForPurchases} from '../../../server/ability'
 import mjml2html from 'mjml'
 import chalk from 'chalk'
+import {PrismaAdapter} from '../../../server/skill-next-auth-prisma-adapter'
```
Comment (joelhooks) [inline apps/testingaccessibility/src/pages/api/auth/[...nextauth].ts:11]: use a local adapter for prisma

apps/testingaccessibility/src/server/skill-next-auth-prisma-adapter.ts
```diff
@@ -0,0 +1,88 @@
+import type {PrismaClient, Prisma} from '../../generated/prisma/client'
+import type {Adapter} from 'next-auth/adapters'
+import {addSeconds, isAfter} from 'date-fns'
+
+export function PrismaAdapter(p: PrismaClient): Adapter {
```
joelhooks, line 5: in practice this is very straight forward. there were a lot of TS errors, but it's a direct cut and paste and functions as expected

apps/testingaccessibility/src/server/skill-next-auth-prisma-adapter.ts
```diff
@@ -0,0 +1,88 @@
...
+            // @ts-ignore
+            const {id: _, ...verificationToken} = token
+            return verificationToken
+          } else {
+            // @ts-ignore
+            const {id: _, ...verificationToken} =
+              await p.verificationToken.delete({
+                where: {identifier_token},
+              })
+            return verificationToken
```
joelhooks, line 76: this is the default official adapter functionality

apps/testingaccessibility/src/server/skill-next-auth-prisma-adapter.ts
```diff
@@ -0,0 +1,88 @@
...
+          const expireMultipleClicks = addSeconds(
+            token.createdAt,
+            TIMEOUT_IN_SECONDS,
+          )
+          const now = new Date()
+
+          if (isAfter(expireMultipleClicks, now)) {
+            // @ts-ignore
+            const {id: _, ...verificationToken} = token
+            return verificationToken
```
joelhooks, line 69: for our local adapter we load the token and see if it was created within the last 90 seconds, and if so we don't expire it on click. it will still expire after a day regardless, this just gives some padding so robo-clicks don't expire the token

joelhooks: one side effect here is that we will have more dead tokens in the DB so we want to set up some sort of mechanism to remove any expired junk tokens

Creeland: Could we do that here? When a user signs in all tokens older than 24 hours are deleted

Creeland (approved):

PR #371, opened 2022-07-26
feat(protailwind): add dynamic unsubscribe page

garrettdimon:
Creates an unsubscribe page that has dynamically updatable messages based on the query string. It's designed to work in conjunction with a ConvertKit rule.

CleanShot 2022-07-26 at 14 59 14@2x

The page was initially pulled from the [Testing Accessibility unsubscribed.tsx page](https://github.com/skillrecordings/products/blob/main/apps/testingaccessibility/src/pages/unsubscribed.tsx), but we noticed that there was [some logic to subscribe people instead of unusubscribe them](https://github.com/skillrecordings/products/blob/main/apps/testingaccessibility/src/pages/unsubscribed.tsx#L24).

Since it looked likely to be [a copy/paste issue from the confirmed.tsx page](https://github.com/skillrecordings/products/blob/main/apps/testingaccessibility/src/pages/confirmed.tsx#L24), I removed it. I plan on removing it from the other pages separately so we never accidentally re-subscribe someone who is trying to unsubscribe.

And some screenshots of the page as it is currently:

`http://localhost:3017/unsubscribed'

CleanShot 2022-07-26 at 14 51 23@2x

`http://localhost:3017/unsubscribed?sequence=1`

CleanShot 2022-07-26 at 14 51 01@2x

It does not have an image like the confirmed page. I think it works without one, but @vojtaholik, if you have one handy to add, let me know, and I'll update it.

CleanShot 2022-07-26 at 14 53 05@2x


apps/protailwind/src/pages/unsubscribed.tsx
```diff
@@ -0,0 +1,39 @@
...
+const Unsubscribed = () => {
+  const router = useRouter()
+  const sequenceId = Number(router.query?.sequence)
+
+  const message = (sequenceId: number) => {
+    switch (sequenceId) {
+      case 1: // unsubscribed: email course
+        return 'Pro Tailwind email course'
+      case 2: // unsubscribed: live workshop info
+        return 'Pro Tailwind live workshops list'
```
joelhooks, line 15: suggestion: send "slugs" versus ints

vojtaholik: I agree with this suggestion, otherwise looks good to me! @garrettdimon 👍

vojtaholik (approved):
I agree with Joel's suggestion, but otherwise looks good to me! @garrettdimon 👍

PR #373, opened 2022-07-26
fix(unsubscribed): remove copy/paste re-subscribe converkit API calls

garrettdimon:
Both of these pages the same copy/paste issues as [the Pro Tailwind unsubscribed.tsx page](https://github.com/skillrecordings/products/pull/371).

Removed the part that could resubscribe people if the right `tag` is passed in the query string.


apps/compilersforhumans/src/pages/unsubscribed.tsx
```diff
@@ -20,14 +15,6 @@ const Unsubscribed = () => {
     }
   }
 
-  React.useEffect(() => {
-    if (tagId && !isEmpty(subscriber)) {
-      axios
-        .post('/api/convertkit/subscribe', {...subscriber, tag: tagId})
-        .catch((err) => console.log(err))
-    }
```
joelhooks, line 28: this bit would be a "resubscribe" handler if we had that sort of button on the screen, but we probably don't need that.

vojtaholik (approved):

PR #375, opened 2022-07-27
feat: @skillrecordings/database is central prisma client/schema

joelhooks:
centralized


.npmrc
```diff
@@ -1 +1,2 @@
-strict-peer-dependencies=false
\ No newline at end of file
+strict-peer-dependencies=false
+public-hoist-pattern[]=*prisma*
```
joelhooks, line 2: this was key to the whole thing or webpack complained at build time

see: https://github.com/prisma/prisma/issues/12921

apps/engmanagement/package.json
```diff
@@ -128,7 +127,6 @@
     "next-images": "^1.8.4",
     "postcss-preset-env": "^7.4.2",
     "prettier": "^2.6.2",
-    "prisma": "^3.15.2",
```
joelhooks, line 131: no need for any prisma stuff in individual packages/apps, just import `@skillrecordings/database`

apps/engmanagement/src/pages/api/auth/[...nextauth].ts
```diff
@@ -12,7 +12,7 @@ export const nextAuthOptions: NextAuthOptions = {
   session: {
     strategy: 'jwt',
   },
-  adapter: PrismaAdapter(prisma as any),
+  adapter: PrismaAdapter(prisma),
```
Comment (joelhooks) [inline apps/engmanagement/src/pages/api/auth/[...nextauth].ts:15]: `any` is no longer required since everything uses the same prisma client source from teh database packages so types are in alignment

PR #378, opened 2022-07-29
chore(testingaccessibility): display upgrade pricing info

joelhooks:
upgrades


apps/testingaccessibility/src/components/pricing.tsx
```diff
@@ -83,9 +83,13 @@ export const Pricing: React.FC<PricingProps> = ({
...
-  const percentOff =
-    appliedMerchantCoupon &&
-    Math.floor(appliedMerchantCoupon.percentageDiscount * 100)
+  const percentOff = appliedMerchantCoupon
+    ? Math.floor(appliedMerchantCoupon.percentageDiscount * 100)
+    : formattedPrice && isDiscount(formattedPrice)
+    ? Math.floor(
+        (formattedPrice.calculatedPrice / formattedPrice.unitPrice) * 100,
+      )
+    : 0
```
joelhooks, line 92: upgrade discounts are calculated on the fly so if no coupon is present and the calculated process is still a discount it's an upgrade

PR #379, opened 2022-07-29
chore(database): upgrade to prisma 4

joelhooks:
prism


apps/testingaccessibility/src/lib/prisma-api.ts
```diff
@@ -220,15 +223,35 @@ export function getSdk(
...
+          },
+        })
+
+      if (merchantCustomersForUser.length > 1) {
+        Sentry.addBreadcrumb({
+          category: 'commerce',
+          level: Sentry.Severity.Warning,
+          message: `user [${user.email}] has multiple stripe customers`,
+        })
+      }
```
joelhooks, line 253: it's currently possible for a given user (email address) to have multiple customers in stripe. this will happen if they make a purchase while logged out in particular

PR #381, opened 2022-07-29
chore(testingaccessibility): remove use of nextjs middleware

joelhooks:
was getting odd local behavior with next middleware and next-auth and the usage in middleware was very light so easier to just skip it. We weren't doing any edge functions.

noticed a log that was complaining about an incompatible library in the jwt package that won't function in edge rntime. might be it?

middleman


vojtaholik (approved):
👍

PR #385, opened 2022-07-29
feat(skill-api): use the /api/skill route for commerce

joelhooks:
Moving all of the commerce endpoints into a shared repo the uses the `/api/skill/*` pattern

This is my third time to do this work and I feel like it's gonna stick. Previously I was being hampered with prisma situation, but with that in its own package, it made it WAY better.

We've got a bunch of "package debris" 

### can to be removed (egghead commerce stack related)
`commerce`
`auth`
`viewer`
`axios`
`stripe`

### Packages to be added as next step:

- [x] * `stripe` - replaces with stripe package that fits current usage
- [x] *  `prisma-commerce` - various price loading utils
- [x] *  `honeycomb-tracer` - honeycomb utils

commerce


apps/testingaccessibility/src/components/pricing.tsx
```diff
@@ -188,7 +188,7 @@ export const Pricing: React.FC<React.PropsWithChildren<PricingProps>> = ({
           </div>
         ) : (
           <form
-            action={`/api/stripe/checkout?productId=${
+            action={`/api/skill/checkout/stripe?productId=${
```
joelhooks, line 191: the `checkout` is the "action" and `stripe` is the provider. Technically we'd be able to utilize different providers in the future but that's not in yet.

apps/testingaccessibility/src/server/ability.ts
```diff
@@ -5,7 +5,7 @@ import {
   hasValidPurchase,
   hasInvoice,
 } from '../utils/purchase-validators'
-import {PurchaseStatus} from '../utils/purchase-status'
+import {PurchaseStatus} from '@skillrecordings/skill-api/dist/enums/purchase-status'
```
joelhooks, line 8: this was a head scratcher, but basically you can't import from root because it introduces node apis that the client side chokes on

packages/skill-api/src/core/index.ts
```diff
@@ -69,6 +74,14 @@ export async function SkillRecordingsHandler<
           context: req?.body?.context,
           config: userOptions,
         })
+      case 'prices':
+        return await loadPrices({params})
+      case 'checkout':
+        return stripeCheckout({params})
```
joelhooks, line 80: oh ya!

PR #391, opened 2022-08-02
feat(testingaccessibility): certificates of completion

vojtaholik:
when learner completes a module, they can download a certificate as png image which looks like this:

my certificate


apps/testingaccessibility/src/utils/download-certificate-image.ts
```diff
@@ -0,0 +1,22 @@
...
+export const downloadCertificateImage = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+  sessionToken: any,
+  module: SanityDocument,
+) => {
+  const userName = req.query.name || sessionToken.name || ''
+
+  const certificateImageUrl =
+    process.env.CERTIFICATE_URI +
```
vojtaholik, line 14: keeping this secret 🔒

joelhooks (approved):

PR #394, opened 2022-08-02
tailwind design for tsforjs

nicollguarnizo:
This PR implements the site's design in tailwind, plus some minor tweaks. 

Preview: https://ts-for-js-git-ng-tsforjs-design-skillrecordings.vercel.app/

TS for JS Devs (4)

The next step will be deleting all the unused files and styled components, I will create another PR for this.


apps/tsforjs/src/components/Accordion.tsx
```diff
@@ -0,0 +1,107 @@
+import React, {useState} from 'react'
+import Image from 'next/image'
+
+export default function Accordion() {
```
joelhooks, line 4: Something to consider when reaching for these types of patters is checking Radix. They have a lot of core components and start from an accessible baseline that can make better experience css with less effort

I didn't dig in deep to see if it's a fit here, but saw "accordion" and wanted to share https://www.radix-ui.com/docs/primitives/components/accordion

joelhooks (approved):
Was there any unused code to be removed?

apps/tsforjs/src/components/Accordion.tsx
nicollguarnizo: @joelhooks, great suggestion, so much cleaner and easy! I implemented the accordion using Radix. 

https://github.com/skillrecordings/products/pull/399

In that PR you can see the new accordion, and I'm also deleting all the unused code.

PR #399, opened 2022-08-03
tsforjs cleaning codebase + improving components 

nicollguarnizo:
-  This PR deletes all the unused code from tsforjs.
- Adds the confirm and confirmed pages 
Screen Shot 2022-08-03 at 6 27 59 PM
- Finally, I went with your suggestion @joelhooks, and I implemented the accordion using Radix. ->


apps/badass/package.json
```diff
@@ -30,7 +30,7 @@
   "dependencies": {
     "@headlessui/react": "^1.6.1",
     "@heroicons/react": "^1.0.6",
-    "@next/mdx": "^12.1.0",
+    "@next/mdx": "^12.2.4",
```
joelhooks, line 33: @nicollguarnizo total-typescript kept failing but only in prod so I bumped all of these to the same version and it worked. my general guess is that something "slipped" in the package lock file, but this kind of stuff is hard to understand. generally we want all the versions of everything to be the same as far as I can tell

joelhooks (approved):

PR #401, opened 2022-08-04
fix(quiz): answer

vojtaholik:
gif


packages/quiz/src/pages/api/answer.ts
```diff
@@ -21,7 +22,7 @@ const answer: NextApiHandler = async (
       const cookieHeader = req.headers.cookie as string
       const [subscriber, ckCookie] =
         await fetchConvertkitSubscriberFromServerCookie(cookieHeader)
-      const questionId = survey.id
+      const questionId = toLower(survey.id)
```
vojtaholik, line 25: Due to convertkit's transforming of field labels to lowercase when created, it broke when we used camelCase.

garrettdimon: Hope it was as easy to find as it was to fix! 🙌🏻

PR #407, opened 2022-08-05
feat(fronttoback): sanity and mjml driven emails

vojtaholik:
email screenshot
emails page screenshot
sanity screenshot
api response screenshot

gif


garrettdimon (commented):
This is just kind of amazing. Some thoughts/observations/questions in no particular order. And I wouldn't even say these are requests so much as ideas for you to decide if they're easy enough or worth implementing.

1. Do you feel good enough about this that it makes sense to extract to a shared package?
2. Given that these will frequently be sequences, should there be an option in Sanity to provide a sequence number and an email number to help ensure they stay in the right order? That definitely helped when keeping updates in sync manually with ConvertKit. The email list could be grouped by sequence then too. Even if there's only one sequence but we end up having other random emails, it's probably worth it to have some sort of attribute for grouping the emails.
3. Should/could things like fonts, colors (email button, text color, link color, etc.) be extracted into constants or pulled from a configuration file? That's likely not worth until we're confident this is worth extracting into a shared package.
4. Is there any reason not to proactively register all (or most) of the syntax highlighting languages? HTML/YAML are the only ones jumping out at me at the moment, but unless there's a huge performance hit, it seems like we should just anticipate that most of the popular languages will get run through this at some point.
5. Is it worth to create a Figma template that correlates closely with the images we know most emails will end up needing so that it's kind of fill-in-the-blank and automatically exports all the layers with the right names?
6. Are there any default templates that we should just add so they can automatically be imported into Sanity? Confirmation emails, "welcome to the list" emails, etc. Or would they be so different from project to project, it's not worth it?
7. Is it worth implementing common images that are likely standardized on the sites already? Logo, Signature, Avatar, etc. It was a really nice touch having the Pro Tailwind logo at the bottom of all of those emails. That might be a nice pattern to use with all of them by default.
8. What would be involved in having these access (or maybe even output?) quiz data? I'm imaging somewhere where the quiz is actually written in Sanity (either separately or as part of the email) and then it could be dumped to a JSON file that `answer.tsx` could load. Given that quizzes are just more content, it feels like they should eventually be in Sanity as well, I imagine.

That's all I've got for now. It's so exciting to see the proof of concept come together so quickly. Hopefully next week I can dive in and see about connecting the bits so we can use it for the next email sequence.

PR #409, opened 2022-08-09
fix(ftb): it's tagId not ckTagId

joelhooks:
the sanity was setup with `ckTagId` property and client expects `tagId` so it was broken

could also transfor this in groq, but I made it a hard change

broken


apps/fronttoback/src/lib/quizzes.ts
```diff
@@ -17,7 +17,7 @@ export async function getQuiz(slug: string) {
     questions[]->{
         questionId, 
         "question": body, 
-        ckTagId, 
+        tagId,
```
joelhooks, line 20: this was super hard to figure out 😭

apps/fronttoback/studio/schemas/documents/question.js
```diff
@@ -46,7 +51,14 @@ export default {
...
   preview: {
     select: {
-      title: 'questionId',
+      label: 'label',
+      questionId: 'questionId',
+    },
+    prepare(selection) {
+      const {label, questionId} = selection
+      return {
+        title: `[${questionId}] ${label}`,
```
joelhooks, line 60: I had a very hard time correlating convertkit sequences, tags, etc with the sanity models. we need to add human labels to these and make it obvious what goes to what when possible

packages/convertkit/src/hooks/use-convertkit.tsx
```diff
@@ -39,25 +37,30 @@ export const ConvertkitProvider: React.FC<
...
-      .then(({data}) => {
-        setSubscriber(data)
-      })
-      .finally(() => setLoadingSubscriber(false))
+      //get the subscriber (which sets the cookie)
+      axios
+        .get(getSubscriberApiUrl, {
+          params: {
+            [CK_SUBSCRIBER_KEY]: ckSubscriberId,
+          },
```
joelhooks, line 57: this will also set the cookie so I made it a little nicer above

packages/skill-api/src/core/services/convertkit/convertkit-load-subscriber.ts
```diff
@@ -13,6 +13,9 @@ export async function convertkitLoadSubscriber({
 }): Promise<OutgoingResponse> {
   try {
     const convertkitId =
+      params.req.query[
+        process.env.NEXT_PUBLIC_CONVERTKIT_SUBSCRIBER_KEY || 'ck_subscriber_id'
+      ] ||
```
joelhooks, line 18: if they are on the query, use that and override cookie

packages/skill-api/src/lib/postmark.ts
```diff
@@ -1,14 +1,16 @@
...
 
-const transport = nodemailer.createTransport(
-  postmarkTransport({
-    auth: {
-      apiKey: process.env.POSTMARK_KEY || '',
-    },
-  }),
-)
+const transport =
+  process.env.POSTMARK_KEY &&
```
joelhooks, line 5: thyis was erroring and in many cases we are "pre-email"

packages/skill-api/src/next/index.ts
```diff
@@ -117,7 +117,7 @@ export interface SkillRecordingsOptions {
     title: string
   }
   slack?: SlackConfig
-  nextAuthOptions: NextAuthOptions
+  nextAuthOptions?: NextAuthOptions
```
joelhooks, line 120: this can be optional if we aren't emailing people

PR #413, opened 2022-08-09
update(ta): readme and local dev instructions

joelhooks:
local


apps/engmanagement/seed_data/01.sql
```diff
@@ -224,6 +223,7 @@ CREATE TABLE `VerificationToken` (
     `token` VARCHAR(191) NOT NULL,
     `identifier` VARCHAR(191) NOT NULL,
     `expires` DATETIME(3) NULL,
+    `createdAt` DATETIME(3) NULL DEFAULT CURRENT_TIMESTAMP(3),
```
joelhooks, line 226: updated this with current schema to remove step from setup

apps/testingaccessibility/bin/seed
```diff
@@ -9,4 +9,4 @@ database='testing-accessibility'
 git_branch=`git rev-parse --abbrev-ref HEAD`
 planetscale_branch=${$git_branch///-} # Substitute '-' for '/' for Planetscale
 
-pscale database restore-dump $database $planetscale_branch --dir ./seed_data/pscale_data_dump
+pscale database restore-dump $database $planetscale_branch --dir ./seed_data
```
joelhooks, line 12: @garrettdimon this didn't actually work in my tests and I was unfamiliar with the procedure.

apps/testingaccessibility/bin/serve
```diff
@@ -1,8 +0,0 @@
-#!/bin/bash
-
-# Description:
-#   Starts the local development server.
-# Usage:
-#   bin/serve
-
-pnpm dev
```
joelhooks, line 8: I removed these since they are just pnpm commands which would be a lot more familiar in our context so this was noise.

I like the `bin` but I think they need to favor a "npm script" first approach where they are primarily executed as npm scripts

apps/testingaccessibility/bin/validate
```diff
@@ -73,5 +73,6 @@ verify_installation 'brew' '/bin/bash -c "$(curl -fsSL https://raw.githubusercon
 verify_installation 'node'
 verify_installation 'npx'
 verify_installation 'pnpm'
+verify_installation 'docker'
```
joelhooks, line 76: this is great!

apps/testingaccessibility/package.json
```diff
@@ -4,18 +4,22 @@
...
   "scripts": {
     "build": "next build",
+    "build:dev": "cd ../.. && pnpm -- turbo run build test lint --filter=testingaccessibility && cd apps/testingaccessibility",
     "postbuild": "NODE_ENV=production next-sitemap",
-    "build:docker": "cd ../../ && docker build . -f apps/testingaccessibility/Dockerfile -t my-awesome-app:latest --build-arg SCOPE=testingaccessibility",
-    "clean": "pnpm -r exec -- rm -rf .turbo node_modules .next dist generated",
-    "dev": "next dev -p 3013",
+    "clean": "pnpm -r exec -- rm -rf .turbo node_modules .next",
+    "dev": "dev:validate && next dev -p 3013",
+    "dev:validate": "bin/validate",
```
joelhooks, line 11: here's an example of using the "npm script" to execute something in bin

vojtaholik (approved):
this is a huge improvement! 🦖

PR #425, opened 2022-08-22
feat(ew): connect form to convertkit

joelhooks:
rocket launch


apps/epic-web/app/routes/index.tsx
```diff
@@ -1,7 +1,88 @@
...
+export const action: ActionFunction = async ({request}) => {
+  const formPayload = Object.fromEntries(await request.formData())
+
+  const validationSchema = z.object({
+    first_name: z.string(),
+    email: z.string().email(),
+  })
+
+  try {
+    const validatedSchema = validationSchema.parse(formPayload)
```
joelhooks, line 27: zod takes your random object, runs it against the schema, and if it is the thing you get a nice typed object back to utilize

packages/convertkit-sdk/src/convertkit.ts
```diff
@@ -0,0 +1,216 @@
+import isEmpty from 'lodash/isEmpty'
```
joelhooks, line 1: I pulled this out of skill-api (next) and added a form method. also added fallbacks for non-next env vars

joelhooks: we'll want to use this new package in skill-api too for sure, just gonna do it in a future pr

PR #439, opened 2022-08-30
feat(kl): add sanity

vojtaholik:
This PR gets sanity setup for keyboard legend and unblocks progress, however I'm going to use skill template to regenerate the app as a next step since it has a bunch of old stuff in it.

gif


apps/keyboardlegend/.env.production
```diff
@@ -23,3 +23,8 @@ CONVERTKIT_BASE_URL=https://api.convertkit.com/v3/
 CONVERTKIT_API_SECRET=converkit_api_secret
 
 NEXT_PUBLIC_CERTIFICATE_URI=next_public_certificate_uri
+
+NEXT_PUBLIC_URL=https://keyboardlegend.com
```
tayiorbeii, line 27: @vojtaholik this should be keyboardlegend.dev

tayiorbeii (approved):
@vojtaholik  change .com to .dev when regenerating from template

PR #447, opened 2022-09-11
feat(tt): adds database

joelhooks:
database


apps/total-typescript/src/pages/tutorials/[module]/[lesson]/index.tsx
```diff
@@ -2,38 +2,36 @@ import React from 'react'
...
 import {checkIfConvertkitSubscriber} from '@skillrecordings/convertkit'
 
 export const getServerSideProps: GetServerSideProps = async (context) => {
   const {params} = context
   const lessonSlug = params?.lesson as string
   const subscriber = await checkIfConvertkitSubscriber(context)
 
-  const blockedLesson = await getBlockedLesson(lessonSlug)
+  const tutorial = await getTutorial(params?.module as string)
+  const lesson = await getLesson(lessonSlug)
```
Comment (joelhooks) [inline apps/total-typescript/src/pages/tutorials/[module]/[lesson]/index.tsx:14]: instead of loading the lesson twice I'm just stripping out the video property using the same logic below

I think tutorial and subscriber could be loaded client side, but we should also have subscriber as a cookie, so we can probably get it a lot faster than a CK round trip in cases where it's cached 🤔

PR #449, opened 2022-09-11
feat(tt): ability to mark solution complete

joelhooks:
changes the `continue` button to `complete` on the solution page and completes the solution lesson

problem here IMO is that you should be able to mark a `section` as complete instead of lessons because in this case a section is a pair with the solution/exercise so it's better represented by the container. another consideration would be that they might want to mark it complete INSTEAD of watching the solution.

finish

### DIFF
--- a/apps/total-typescript/src/pages/api/progress/[lesson].ts
+++ b/apps/total-typescript/src/pages/api/progress/[lesson].ts
@@ -18,7 +18,7 @@ const lessonProgressHandler = async (
 ) => {
   if (req.method === 'POST') {
     try {
-      const {findOrCreateUser, toggleLessonProgressForUser} = getSdk()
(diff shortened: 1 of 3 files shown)


apps/total-typescript/src/pages/api/progress/[lesson].ts
```diff
@@ -36,7 +36,7 @@ const lessonProgressHandler = async (
 
       const {user} = await findOrCreateUser(subscriber.email_address)
 
-      const progress = await toggleLessonProgressForUser({
+      const progress = await completeLessonProgressForUser({
```
Comment (joelhooks) [inline apps/total-typescript/src/pages/api/progress/[lesson].ts:39]: you can't uncomplete a lesson

vojtaholik (approved):
works like a charm! I agree about the considerations stated here, but approving since it's a big improvement and the UI isn't closely tied to user progress at the moment.

PR #451, opened 2022-09-12
update(tt): just use default form

vojtaholik:
this removes option to subscribe to specific form.

gif


apps/total-typescript/package.json
```diff
@@ -8,6 +8,7 @@
     "postbuild": "NODE_ENV=production next-sitemap",
     "start": "next start -p 3016",
     "lint": "eslint src/**/*.{ts,tsx} --fix && tsc --noEmit",
+    "db:start": "docker compose up -d",
```
vojtaholik, line 11: this was helpful when starting up so I left it

PR #452, opened 2022-09-12
feat(tt): use an exercise model instead of lessons

joelhooks:
shell game


joelhooks (commented):
this is kind of sloppy, but we should be able to delete `section` and `lesson` docs in Sanity and just have modules/exercises

vojtaholik (approved):
this is so much better and makes a lot more sense. 💯

PR #465, opened 2022-09-16
feat(tt): API route that updates Video Resources with transcripts

Creeland:
Adds an API route that receives completed transcripts from Castingwords and adds them to Video Resource documents in Sanity. Currently, we have to manually add transcripts to documents

This is the next step towards the automated creation of video resources with transcripts. 

Next step is to create a workflow that orders transcripts from Castingwords

cast spell


apps/total-typescript/src/pages/api/sanity/transcript/update.ts
```diff
@@ -0,0 +1,117 @@
+import {withSentry} from '@sentry/nextjs'
+import {NextApiRequest, NextApiResponse} from 'next'
+import sanityClient from '@sanity/client'
+import {json} from 'stream/consumers'
+import {uniqueId} from 'lodash'
+import {randomUUID} from 'crypto'
+
+const addTranscriptToVideoDocument = async (
```
joelhooks, line 8: is this a casting words webhook receiver?

if so it should be `/api/skill/transcript/castingwords`

what is doing with the transcript is a second order concern likely by a "service"

apps/total-typescript/src/pages/api/sanity/transcript/update.ts
```diff
@@ -0,0 +1,117 @@
...
+  req: NextApiRequest,
+  res: NextApiResponse,
+) => {
+  if (req.method === 'POST') {
+    const client = sanityClient({
```
joelhooks, line 13: we have a configured client here https://github.com/skillrecordings/products/blob/main/apps/total-typescript/src/utils/sanity-client.ts

apps/total-typescript/src/pages/api/sanity/transcript/update.ts
```diff
@@ -0,0 +1,117 @@
...
+              return
+            })
+        })
+        .catch((err) => {
+          console.log(err)
+          res.status(500).json({
+            message: `Invalid query for Sanity document`,
+          })
+        })
+    } else {
```
joelhooks, line 104: is this invalid or just something we don't handle? we should probably handle it accordingly if it's just a status not being handled.

Creeland: This error is thrown when it can't find a document with the audiofile id. I can update the message to be more specific to that

apps/total-typescript/src/pages/api/castingwords/create.ts
```diff
@@ -0,0 +1,63 @@
+import {withSentry} from '@sentry/nextjs'
+import {NextApiRequest, NextApiResponse} from 'next'
+import {isValidSignature, SIGNATURE_HEADER_NAME} from '@sanity/webhook'
+
+const secret = process.env.SANITY_WEBHOOK_ORDER_TRANSCRIPT_SECRET
+
+const createCastingwordsOrder = async (
```
joelhooks, line 7: if this is a sanity webhook receiver it should be named accordingly. somalian to the comment below 

a sanity webhook recover isn't a representative of a single procedure like "order a transcript" it is an indicator that something g has changed on the sanity side and internally will orchestrate all actions that occur in response to the event received.

joelhooks: `/api/skill/webhook/sanity`

unless this is different from my suggested workflow and functioning as a Remote Procedure Call

apps/total-typescript/src/lib/castingwords.test.ts
```diff
@@ -0,0 +1,19 @@
...
+  expect(url).toEqual(
+    `https://castingwords.com/store/API4/order_url?sku=test1&test=1&url=https%3A%2F%2Fexample.com%2Fvideo.mp4&api_key=castingwords-api-key`,
+  )
+})
+
+test('orderTranscript should call castingwords url and return order', async () => {
+  const order = await orderTranscript(
+    'https://www.dropbox.com/s/bssbwpz4khvvosr/beginners-01.problem.mp4?dl=1',
+  )
+  expect(order.order).toEqual(`teSt`)
```
joelhooks, line 18: in this case the specific value doesn't matter but "toBeDefined" would be enough

joelhooks (approved):

PR #470, opened 2022-09-20
fixing the author for badass articles

nicollguarnizo:
I realized that when someone shares an article from badass.dev the author was marcy, this small PR changes the author for @joelhooks 

Screen Shot 2022-09-20 at 11 47 52 AM


apps/badass/src/templates/article-template.tsx
```diff
@@ -244,7 +244,7 @@ const Share: React.FC<React.PropsWithChildren<{title: string}>> = ({title}) => {
   const url = process.env.NEXT_PUBLIC_URL + router.asPath
   const className =
     'p-3 hover:bg-white hover:bg-opacity-10 transition rounded-full focus-visible:ring-white'
-  const message = `${title} by @marcysutton`
+  const message = `${title} by @jhooks`
```
joelhooks, line 247: the problem is that this value is hard coded here instead of utilizing our "config" or some other source

joelhooks (approved):

PR #475, opened 2022-09-22
feat(tt): test using ssg instead of ssr

joelhooks:
this is a fairly big shift. originally I wanted to use `trpc` but have struggled to get it's SSR functionality working consistently.

I'm not stoked about page performance using getServerSideProps, even when using SWR cache headers. When you use getServerSideProps, **every** page transition loads from the server, so there is a bit of lag.

For our use case, we only really want to block content at the lesson level, and in that circumstance we are verifying a convertkit subscriber or user account.

static


apps/total-typescript/src/pages/api/subscriber.ts
```diff
@@ -1,5 +1,52 @@
...
+const convertkitSubscriberHandler = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+) => {
+  if (req.method === 'GET') {
+    let subscriber
+
+    if (req.cookies['ck_subscriber']) {
+      subscriber = JSON.parse(req.cookies['ck_subscriber'])
+    } else if (req.query.learner) {
```
joelhooks, line 18: if they visit with `learner` query param we look the id up to see if we've got a match

apps/total-typescript/src/pages/api/subscriber.ts
```diff
@@ -1,5 +1,52 @@
...
+      const user = await prisma.user.findUnique({
+        where: {id: req.query.learner as string},
+      })
+
+      subscriber = user?.email && (await getSubscriberByEmail(user?.email))
+    } else if (req.query.ckSubscriberId) {
+      subscriber = await fetchSubscriber(req.query.ckSubscriberId as string)
```
joelhooks, line 25: if we've got a subscriber id, we use that

apps/total-typescript/src/pages/api/subscriber.ts
```diff
@@ -1,5 +1,52 @@
...
+        {
+          secure: process.env.NODE_ENV === 'production',
+          path: '/',
+          httpOnly: true,
+          sameSite: 'lax',
+          maxAge: 31556952,
+        },
+      )
+
+      res.setHeader('Set-Cookie', convertkitCookie)
```
joelhooks, line 41: we want to cache it in the cookie

apps/total-typescript/src/pages/api/subscriber.ts
```diff
@@ -1,5 +1,52 @@
...
+      res.status(200).json(subscriber)
+    } else {
+      res.status(200).end()
```
joelhooks, line 44: this actually throws an error on the client, but we noop it

apps/total-typescript/src/pages/server-sitemap.xml/index.tsx
```diff
@@ -1,41 +1,11 @@
...
-          //solutions
-          loc: `${tutorialRootUrl}/${exercise.slug.current}/solution`, // Absolute url
-          lastmod: new Date(exercise._updatedAt).toISOString(),
-          changefreq: 'weekly',
-          priority: 0.7,
-        }
-      }),
-    ]
-  })
+  const tutorialSitemap: any[] = []
```
joelhooks, line 8: if I just remove this the tips sitemap throws a type error and I don't understand why

removing these because they are added automatically if you use SSG

apps/total-typescript/src/pages/tutorials/[module]/[exercise]/index.tsx
```diff
@@ -1,36 +1,42 @@
...
-  if (!exercise) {
-    return {
-      notFound: true,
-    }
-  }
-
   return {
-    props: {exercise, module, subscriber},
+    props: {exercise, module},
+    revalidate: 10,
```
Comment (joelhooks) [inline apps/total-typescript/src/pages/tutorials/[module]/[exercise]/index.tsx:16]: this turns on isr. the `revalidate` is 10 seconds, after that it regenerates the static page

https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration

in theory this means you should be able to update the cdn (Sanity) and it will load with new data and recache. They recommend turning `useCdn: false` in cdn, but I don't know how that works in practice

apps/total-typescript/src/pages/tutorials/[module]/[exercise]/index.tsx
```diff
@@ -1,36 +1,42 @@
...
   }
 }
 
-const ExercisePage: React.FC<any> = ({exercise, module, subscriber}) => {
-  return exercise ? (
```
Comment (joelhooks) [inline apps/total-typescript/src/pages/tutorials/[module]/[exercise]/index.tsx:27]: it's always there now, static...

apps/total-typescript/src/pages/tutorials/[module]/[exercise]/index.tsx
```diff
@@ -1,36 +1,42 @@
...
+          params: {
+            module: tutorial.slug.current,
+            exercise: exercise.slug.current,
+          },
+        }
+      })
+    }, [])
+    .flatMap((path: any) => path)
+  return {paths, fallback: 'blocking'}
+}
```
Comment (joelhooks) [inline apps/total-typescript/src/pages/tutorials/[module]/[exercise]/index.tsx:36]: this tells it what exercise paths to build statically and requires the module and exercise to construct the path params

you can limit this to a handlful of pages, and any excluded pages will be built and cached at runtime.

PR #480, opened 2022-09-23
feat(tt): tip progress in indexeddb (local only)

joelhooks:
This uses the ✅ all over and shouldn't be approved without that being changed, so it's draft.

Using IndexedDB we can have a database of events. Currently only looking at the video complete event for tips. This could later be sync'd to database and/or prefer the database as needed.

Using react-query means it will cache for general display.

screenshot

gif


apps/total-typescript/src/hooks/use-mux-player.tsx
```diff
@@ -33,11 +33,19 @@ type VideoProviderProps = {
   lesson: VideoResource
   path?: string
   muxPlayerRef: any
+  onEnded?: () => Promise<any>
```
joelhooks, line 36: wanted to be able to track in the local context, specifically to write to database

apps/total-typescript/src/hooks/use-tip-complete.ts
```diff
@@ -0,0 +1,30 @@
...
+})
+
+export type TipEvent = z.infer<typeof TipEventSchema>
+
+export const useTipComplete = (tipSlug: string) => {
+  const {getManyByIndex} = useIndexedDBStore('progress')
+  const {data: completionEvents, status} = useQuery(
+    ['completionEvents', tipSlug],
+    async () => {
+      const tipEvents = await getManyByIndex('lesson', tipSlug)
```
joelhooks, line 19: access indexeddb for all objects with 'lesson' property that matches the tip

apps/total-typescript/src/hooks/use-tip-complete.ts
```diff
@@ -0,0 +1,33 @@
...
+      return z
+        .array(TipEventSchema)
+        .parse(tipEvents)
+        .filter((tipEvent) => {
+          return (
+            tipEvent.eventName === 'completed video' &&
+            tipEvent.module === 'tips'
```
joelhooks, line 26: as the hook name implies, we only care about tip completions

apps/total-typescript/src/pages/tips/index.tsx
```diff
@@ -129,7 +131,10 @@ export const TipTeaser: React.FC<{tip: Tip}> = ({tip}) => {
             },
           }}
         >
-          <a className="hover:underline">{title}</a>
+          <a className="hover:underline">
+            {tipCompleted ? '✅ ' : ''}
```
joelhooks, line 135: can sprinkle this around however you like, not context dependent

apps/total-typescript/src/templates/tip-template.tsx
```diff
@@ -44,8 +50,22 @@ const TipTemplate: React.FC<TipPageProps> = ({tip, tips}) => {
...
   })
 
+  const handleVideoEnded = async () => {
+    await add({
+      eventName: 'completed video',
+      module: 'tips',
+      lesson: tip.slug.current,
+      createdOn: new Date(),
+    }).then(console.debug)
+  }
```
joelhooks, line 60: we are adding the dip here

apps/total-typescript/src/templates/tip-template.tsx
```diff
@@ -212,63 +231,81 @@ const TipOverlay: React.FC<{tips: Tip[]}> = ({tips}) => {
...
+          {take(shuffle(tips), 9).map((tip) => (
+            <VideoOverlayTipCard suggestedTip={tip} />
+          ))}
         </div>
       </div>
     </div>
   )
 }
 
+const VideoOverlayTipCard: React.FC<{suggestedTip: Tip}> = ({suggestedTip}) => {
```
joelhooks, line 243: extracted this out so it's a FC instead of a "callback" so it can use hooks directly

vojtaholik (approved):

PR #488, opened 2022-09-28
cleanup: ignore sitemap files from git

jbranchaud:
The sitemap files should only be needed as build assets in Vercel. We
shouldn't need to keep them under version control. They create a lot of
churn and noise in the git working directory, commits, and PRs. By
adding them to the `.gitignore` for the monorepo, we can get rid of that
clutter.

Here is a screenshot of my `git status` before this change, to illustrate the amount of noise that sitemaps create in the git working directory.

I think we can do the same with the `robot.txt` files, but wanted to take it one step at a time.

CleanShot 2022-09-27 at 17 11 50@2x

site map


.gitignore
```diff
@@ -6,3 +6,7 @@ dist
 .pnpm-debug.log
 .DS_Store
 tsconfig.tsbuildinfo
+
+# build artifacts
+apps/*/public/sitemap.xml
+apps/*/public/sitemap-[0-99].xml
```
joelhooks, line 12: TIL

joelhooks (approved):

.gitignore
jbranchaud: This .gitignore pattern matching syntax was new to me, found it here https://www.golinuxcloud.com/gitignore-examples/#5_Examples_of_pattern_matching_in_gitignore

CleanShot 2022-09-30 at 12 57 59@2x

PR #490, opened 2022-09-28
feat(ta): differentiate team/individual purchases

jbranchaud:
The Thank You page should differentiate between team and individual
purchases. There is some additional useful info we can provide to a team
purchaser.

This PR is a conversation starter. There is some additional info we can access to provide more context to a team purchaser (e.g. number of seats, invoice link, and team invite link).
- What do we want to say?
- How should it be displayed?
- What other context could be useful? (e.g. a link to their invoice)

_This commit needs some help with design and copywriting._

CleanShot 2022-09-28 at 17 56 35@2x

team


joelhooks:
@vojtaholik I think this could go further and have an invoice link and share code

I was also thinking that a good next step here would be to make the stripe success callback a login link so they are logged in on this screen.

vojtaholik:
I think that the `thanks/purchase` page should ideally has the same content as `/welcome`
which essentially means logging them in right away as Joel has suggested

welcome page

joelhooks:
Sounds like we've got a strong basis for some next steps! ️

vojtaholik (approved):
if we could log them in and essentially skip this step and present them with content from welcome page instead, that would be a huge improvement

PR #493, opened 2022-10-04
feat(csa): updated devPort in package.json

vojtaholik:
gif

### DIFF
--- a/packages/create-skill-app/create-app.ts
+++ b/packages/create-skill-app/create-app.ts
@@ -140,9 +140,9 @@ export async function createApp({
         case 'README-template.md': {
           return 'README.md'
         }
-        // case 'package-template.json': {
-        //   return 'package.json'
-        // }
+        case 'package-template.json': {
+          return 'package.json'
+        }
         default: {
           return configTemplate
         }
--- a/packages/create-skill-app/create-app.ts
+++ b/packages/create-skill-app/create-app.ts
@@ -90,9 +90,22 @@ export async function createApp({
     },
   })
 
+  const defaultDevPort = Number(3000).toString()
+
   const packageJson = {
     name: appName,
     ...defaultPackage,
+    scripts: {
+      ...defaultPackage.scripts,
+      dev: defaultPackage.scripts.dev.replace(
+        defaultDevPort,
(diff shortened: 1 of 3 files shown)


packages/create-skill-app/create-app.ts
```diff
@@ -127,6 +140,9 @@ export async function createApp({
         case 'README-template.md': {
           return 'README.md'
         }
+        case 'package-template.json': {
+          return 'package.json'
+        }
```
vojtaholik, line 145: we could source package.json from a template, but I haven't done that since having two of them seems cumbersome and easy to miss...

packages/create-skill-app/create-app.ts
```diff
@@ -90,9 +90,22 @@ export async function createApp({
...
+      ...defaultPackage.scripts,
+      dev: defaultPackage.scripts.dev.replace(
+        defaultDevPort,
+        projectData.devPort,
+      ),
+      start: defaultPackage.scripts.start.replace(
+        defaultDevPort,
+        projectData.devPort,
+      ),
+    },
```
vojtaholik, line 108: so I ended up processing it like this instead. `dev` and `start` commands will both get the correct dev port.

packages/create-skill-app/index.ts
```diff
@@ -80,7 +80,7 @@ async function run(): Promise<void> {
     {
       type: 'text',
       name: 'lastName',
-      message: `What is our partner's first name?`,
+      message: `What is our partner's last name?`,
```
vojtaholik, line 83: noticed this little typo

packages/create-skill-app/templates/next/package.json
```diff
@@ -8,7 +8,8 @@
     "lint": "eslint src/**/*.{ts,tsx} --fix && tsc --noEmit",
     "db:studio": "prisma studio",
     "db:push": "prisma db push",
-    "db:schema": "prisma migrate dev --create-only"
+    "db:schema": "prisma migrate dev --create-only",
+    "db:start": "docker compose up -d"
```
vojtaholik, line 12: added `db:start` command to spin up docker

PR #496, opened 2022-10-04
build(ts): allow build when missing algolia key

jbranchaud:
The build fails locally when the ALGOLIA_API_WRITE_KEY is missing. It
fails in an obscure way that makes it difficult to get the app up and
running.

The obscure build error is: ` TypeError [ERR_HTTP_INVALID_HEADER_VALUE]: Invalid value "undefined" for header "x-algolia-api-key"`.

This commit moves the key to a more appropriate file
(`env.local.template`) and then updates the Algolia client code to
fallback to an empty string if it is missing.

Update: Use Zod to parse and validate the `ALGOLIA_API_WRITE_KEY` env var which
smooths out the potentially undefined value and deals with type
inference. In the case that this value is undefined in the context of
the production environment, it raises a Zod validation error. This will
happen during the build if that value is missing which is definitely
when we want to catch it.

algo


apps/testingaccessibility/src/utils/algolia/sanity-algolia-client.ts
```diff
@@ -1,9 +1,31 @@
...
+        code: z.ZodIssueCode.custom,
+        message:
+          'ALGOLIA_API_WRITE_KEY must be defined in the production environment.',
+      })
+
+      return z.NEVER
+    }
+
+    return val || ''
+  })
```
jbranchaud, line 21: This is one of the places where Zod's type inference really shines.

The environment variables can be either a string or undefined, so I use a schema to parse this algolia API key.

In the transform, I'm able make the return value into a string which narrows the inferred type. I'm also able to do a runtime check of the value when in production and turn that info a zod error. The inferred type still ends up as `string` (which the `algoliaSearchClientCreator` expects).

joelhooks (approved):

PR #497, opened 2022-10-05
fix(ta): prevent default export of server-only code

jbranchaud:
The server-side-only algolia client configured with the secret API write
key should only be imported into API code. Because it was being `*`
exported from the `utils/algolia/index.ts` file, it was being bundled
into the frontend code.

This was originally causing silent errors because the algolia client was
being configured with a blank value (that write key is only visible
servers-side). The schema parsing -- meant only to validate the key
server-side -- that I recently added was being triggered on the frontend
which caused that aforementioned error to fully surface.

This commit removes the re-export of the server-side algolia client.
Instead, the API code can import it directly from
`utils/algolia/sanity-algolia-client.ts`. Having to explicitly import it
form this location should help avoid this code getting accidentally
included in the frontend bundle going forward.

search

### DIFF
--- a/apps/testingaccessibility/src/utils/algolia/sanity-algolia-client.ts
+++ b/apps/testingaccessibility/src/utils/algolia/sanity-algolia-client.ts
@@ -21,9 +21,7 @@ const algoliaWriteKeySchema = z
   })
 
 const getAlgoliaApiWriteKeyFromEnvironment = () => {
-  // TODO: make sure this doesn't break production /learn page
-  // return algoliaWriteKeySchema.parse(process.env.ALGOLIA_API_WRITE_KEY)
-  return process.env.ALGOLIA_API_WRITE_KEY
+  return algoliaWriteKeySchema.parse(process.env.ALGOLIA_API_WRITE_KEY)
 }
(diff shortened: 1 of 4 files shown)


apps/testingaccessibility/src/utils/algolia/sanity-algolia-client.ts
```diff
@@ -21,9 +21,7 @@ const algoliaWriteKeySchema = z
   })
 
 const getAlgoliaApiWriteKeyFromEnvironment = () => {
-  // TODO: make sure this doesn't break production /learn page
-  // return algoliaWriteKeySchema.parse(process.env.ALGOLIA_API_WRITE_KEY)
-  return process.env.ALGOLIA_API_WRITE_KEY
+  return algoliaWriteKeySchema.parse(process.env.ALGOLIA_API_WRITE_KEY)
```
jbranchaud, line 24: @vojtaholik had to comment this out because it was breaking the learn page. It can be restored now 🎉

vojtaholik (approved):
nice one. thanks! 🎉

PR #498, opened 2022-10-05
build(ta): remove `build:dev` from TA pnpm scripts

jbranchaud:
According to a recent discussion, the `build:dev` command (introduced in [this PR](https://github.com/skillrecordings/products/commit/bf0fbe34244cd6414352b6f1308e05010313ecf1#diff-284649712bae897eb9f7e577e57534aa1ec714618f8b0b4641d70de63818c8ff))
is not used. Instead, if you want to build the app, you should run `pnpm
build`.

This commits removes `build:dev` from both `package.json` and the
`README.md` to avoid confusion about how to set up and build this app.

confusion

### DIFF
--- a/apps/testingaccessibility/README.md
+++ b/apps/testingaccessibility/README.md
@@ -43,7 +43,7 @@ Copy the template `.env.local.template` file to `.env.local` and `.env.template`
 
 🔒 `env.local` contains local __private environment variables__
 
-* `ALGOLIA_API_WRITE_KEY`: Required to build the project with both `build` and `build:dev`.
+* `ALGOLIA_API_WRITE_KEY`: Required when running `pnpm build` which invokes `next build` which use `NODE_ENV=production`.
 * `CONVERTKIT_API_SECRET`: not required for local development unless actively working on ConvertKit integration. Can be found in 1password.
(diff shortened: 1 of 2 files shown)


apps/testingaccessibility/README.md
```diff
@@ -43,7 +43,7 @@ Copy the template `.env.local.template` file to `.env.local` and `.env.template`
 
 🔒 `env.local` contains local __private environment variables__
 
-* `ALGOLIA_API_WRITE_KEY`: Required to build the project with both `build` and `build:dev`.
+* `ALGOLIA_API_WRITE_KEY`: Required when running `pnpm build` which invokes `next build` which use `NODE_ENV=production`.
```
jbranchaud, line 46:
```suggestion
* `ALGOLIA_API_WRITE_KEY`: Required when running `pnpm build` which invokes `next build` which uses `NODE_ENV=production`.
```

jbranchaud:
I'd suspect we could make a similar change in a couple other apps.

I noticed that `apps/esculafrontend` and `apps/typescriptcourse` both also have this script defined.

I also wonder if the `build:dev` script in the root `package.json` is being used and/or can be removed.

joelhooks (approved):

PR #502, opened 2022-10-07
Fix screen flickers (probably)

DrShpongle:
Related to this issue: https://github.com/total-typescript/zod-tutorial/issues/15
It’s hard to say what exactly could be the reason of this behavior, can’t reproduce it on my side. I applied the change that could help “in theory” (just changed width: 100%  for flex-grow: 1 )


vojtaholik:
this doesn't seem to fix the issue ;/

joelhooks (approved):
fixes https://github.com/total-typescript/zod-tutorial/issues/15

PR #503, opened 2022-10-07
feat(database): add redeemedCouponId to purchases

jbranchaud:
Add a specific column for tracking the bulk coupon that was redeemed in
order to create this $0 purchase. This will allow us to repurpose the
`couponId` column toward representing a one-to-many relationship between
a bulk coupon and the purchases that add seats to it.

This is the **first** step of [the expand-and-contract pattern](https://www.prisma.io/dataguide/types/relational/expand-and-contract-pattern). I'll do each of the following steps in sequence:
- Open a pscale Deploy Request and get this schema change into production
- Open PR to have coupon redemptions write to this new column alongside the original column
- Add and run script to do a _data migration_ any existing coupon redemptions purchase (write `purchase.couponId` value to `purchase.redeemedCouponId` column).
- Open PR: Anywhere the apps are reading from `purchase.couponId` needs to be cut over to `purchase.redeemedCouponId`
- Open PR to remove writes to the `purchase.couponId` column
- No need to do the last expand-and-contract step of removing the old part of the schema because we will then repurpose the `purchase.couponId` toward representing one-to-many bulk coupon purchases (i.e. adding more seats to an existing bulk coupon).

---

More details copied over from [Roam](https://roamresearch.com/#/app/egghead/page/Uub3Ki4rM):

- I originally thought the `couponId` column wasn't being used. In my initial poking around of the app, I only saw the bulk coupon being set. After redeeming a purchase, I realized that `purchases.couponId` is being used after all.
    - Whenever one of the seats for a bulk coupon is redeemed, a `purchase` record is created which points to the originating coupon at `purchase.couponId`.
- To transition to a schema that supports multiple bulks purchases and to be able to trace Coupon Redemption purchases back to the originating coupon, I'm going to need to make a series of changes.
    - A `purchase` record needs to be able to tell us one of two things:
        - The Bulk Coupon it added seats to.
        - The Bulk Coupon it was redeemed from.
        - (or if it was a single-seat purchase, neither of these values should be set)
    - We are going to need two separate Coupon ID foreign keys to do this.
        - If we used the existing `couponId` column to represent both of these things, the data could get mixed up and hard to query from.

The series of changes needed:

1. Add `redeemedCouponId` column to `purchases` (with the Expand-and-Contract pattern) to represent the bulk coupon that was redeemed for the $0 purchase.
    - This would be created anytime a team invite is redeemed for a bulk coupon.
2. Use the now empty `couponId` column (or rename it to `bulkCouponId` or `teamCouponId`) to represent the one-to-many relationship between (Bulk) Coupon and Purchases.
    - i.e. the first and any subsequent purchases of seats for a team license should point to the coupon to which they added capacity.
        - For instance, if you purchase 5 seats, then you get a coupon with capacity of 5. A new purchase for $X points to that coupon. If you buy 4 more seats, then 4 is added to that coupon's original capacity __and__ a new purchase record is created for $Y pointing to that same coupon.
    - This will follow the steps outlined in: We need to use the [[Expand-and-Contract database migration strategy](https://www.prisma.io/dataguide/types/relational/expand-and-contract-pattern)](https://www.prisma.io/dataguide/types/relational/expand-and-contract-pattern) for this.

redeem

### DIFF
--- a/apps/testingaccessibility/README.md
+++ b/apps/testingaccessibility/README.md
@@ -212,13 +212,27 @@ npx prisma studio
 
 When you make a new branch in Planetscale it doesn't bring data over.
 
-In the `seed_data` folder is a database dump.
+#### Minimal Seed Data
+
+You can seed your branch with the basics that are associated with the **test mode** Stripe account. Those seeds are located in `seed_data`.
 
 ```bash
-pscale database restore-dump testing-accessibility next-steps --dir ./seed_data/pscale_data_dump
+pscale database restore-dump testing-accessibility next-steps --dir ./seed_data
 ```
 
-This should set up the basics that are associated with the **test mode** Stripe account
+#### Production-like Seed Data
+
+You can locally dump a copy of the `main` branch using `pscale database dump`:
+
+```bash
+pscale database dump testing-accessibility main --output ./seed_data/pscale_data_dump
+```
+
+Now the `seed_data/pscale_data_dump` folder holds a database dump you can restore to your branch:
+
+```bash
+pscale database restore-dump testing-accessibility next-steps --dir ./seed_data/pscale_data_dump --overwrite-tables
+```
 
 ## Edit content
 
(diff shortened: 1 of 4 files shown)


jbranchaud:
Planetscale Deploy Request for these schema changes: https://app.planetscale.com/skill-recordings/testing-accessibility/deploy-requests/13

jbranchaud:
Replacing this PR with https://github.com/skillrecordings/products/pull/510

PR #509, opened 2022-10-10
feat(db): spike out data migration script setup

jbranchaud:
This is a WIP/spike on how to write a data migration script for a
scenario where the schema has been changed and now data needs to be
migrated to that updated schema.

It should lean on Prisma types and should be runnable across the various
apps using the schema for their databases.

It uses `ts-node` to execute the typescript data migration file. Open to
suggestions if there is something else I should be using for this
particular piece of this. It's what I found.

**Possibly my brain is fried from a long day, but I'm struggling to
understand where the node environment is being derived from when I
execute this script. No matter what directory I execute this from,
`ts-node` has the impression that `DATABASE_URL` is set to the URL
defined in TA's `.env` file. What am I missing? How do I make the
environment variables specific to the project this script is being run
in?**

I'm executing it with `npx ts-node
../../packages/database/migrate-data/<name-of-file>.ts`.

spi

### DIFF
--- a/packages/database/migrate_data/20221010203922_copy_existing_coupon_id_to_redeemed_for_purchase.ts
+++ b/packages/database/migrate_data/20221010203922_copy_existing_coupon_id_to_redeemed_for_purchase.ts
@@ -1,0 +1,37 @@
+// const {prisma} = require('../src/client')
+import {prisma} from '../src/client'
+
+const copyOverCouponIdToRedeemedCouponId = async () => {
+  const purchaseCount = await prisma.purchase.count()
+  const purchasesThatNeedUpdating = await prisma.purchase.findMany({
+    where: {redeemedCouponId: null, couponId: {not: null}},
+  })
+
+  console.log(
+    `Need to update ${purchasesThatNeedUpdating.length} of ${purchaseCount} purchases.`,
+  )
+
+  purchasesThatNeedUpdating.forEach(async (purchase) => {
+    await prisma.purchase.update({
+      where: {
+        id: purchase.id,
+      },
+      data: {
+        redeemedCouponId: purchase.couponId,
+      },
+    })
+  })
+
+  const afterPurchasesThatNeedUpdating = await prisma.purchase.findMany({
+    where: {redeemedCouponId: null, couponId: {not: null}},
+  })
+
+  const numberUpdated =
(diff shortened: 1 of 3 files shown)


packages/database/migrate_data/20221010203922_copy_existing_coupon_id_to_redeemed_for_purchase.ts
```diff
@@ -0,0 +1,37 @@
...
+    where: {redeemedCouponId: null, couponId: {not: null}},
+  })
+
+  const numberUpdated =
+    purchasesThatNeedUpdating.length - afterPurchasesThatNeedUpdating.length
+  console.log(`Successfully updated ${numberUpdated} purchases`)
+}
+
+// copyOverCouponIdToRedeemedCouponId()
+console.log({database_url: process.env.DATABASE_URL})
```
jbranchaud, line 35: Possibly my brain is fried from a long day, but I'm struggling to
understand where the node environment is being derived from when I
execute this script. No matter what directory I execute this from,
`ts-node` has the impression that `DATABASE_URL` is set to the URL
defined in TA's `.env` file. What am I missing? How do I make the
environment variables specific to the project this script is being run
in?

joelhooks: TA is the only product with purchases so I don't think this needs to be codified in the package scope

I'd also hesitate to make it appear in a format that makes it look like it should/is a "rails style migration" and not just a one off since part of the purpose here is to make these changes before more than just TA need them. flexible on this, just want to throw it out there

jbranchaud: > TA is the only product with purchases so I don't think this needs to be codified in the package scope

I was hoping to answer the question of "how do I migrate data across app databases?" sooner rather than later. It won't be long before we have data in both that needs to be massaged.

> I'd also hesitate to make it appear in a format that makes it look like it should/is a "rails style migration"

I went overboard with the filename. It is intended to be a one-off migration.

jbranchaud:
This is subsumed by https://github.com/skillrecordings/products/pull/510

PR #510, opened 2022-10-11
feat(database): add redeemedBulkCouponId to purchases

jbranchaud:
_Opening this in place of https://github.com/skillrecordings/products/pull/503, and big 🙌 to @garrettdimon for pairing on this with me_

Add a specific column for tracking the bulk coupon that was redeemed in
order to create this $0 purchase. This will allow us to repurpose the
`couponId` column toward representing a one-to-many relationship between
a bulk coupon and the purchases that add seats to it.

This is the **second** and **third** steps of [the expand-and-contract pattern](https://www.prisma.io/dataguide/types/relational/expand-and-contract-pattern). The **first** step is done manually in the Planetscale dashboard. Each of the following steps should be done in sequence:
- Open a pscale Deploy Request and get this schema change into production
- Open PR to have coupon redemptions write to this new column alongside the original column
- Add and run script to do a _data migration_ any existing coupon redemptions purchase (write `purchase.couponId` value to `purchase.redeemedCouponId` column).
- Open PR: Anywhere the apps are reading from `purchase.couponId` needs to be cut over to `purchase.redeemedCouponId`
- Open PR to remove writes to the `purchase.couponId` column
- No need to do the last expand-and-contract step of removing the old part of the schema because we will then repurpose the `purchase.couponId` toward representing one-to-many bulk coupon purchases (i.e. adding more seats to an existing bulk coupon).

---

- feat(database): add redeemedBulkCouponId to purchases
- docs(ta): explain two data seeding scenarios
- feat(ta): write coupon ID to both fields
- feat(ta): script to migrate redeemedBulkCouponId
- docs(ta): explain how to run data migrate script

bulk

PR #511, opened 2022-10-12
docs(ta): how to test and deploy schema changes

jbranchaud:
This adds more detail and some planetscale specifics for how to create, evaluate, and deploy Prisma schema changes.

This is all based on what I picked up going through the whole process for https://github.com/skillrecordings/products/pull/510

docs

### DIFF
--- a/apps/testingaccessibility/README.md
+++ b/apps/testingaccessibility/README.md
@@ -192,13 +192,65 @@ The `dev:stripe` node script is a shorthand for `stripe listen --forward-to loca
 pscale connect testing-accessibility next-steps --port 3309
 ```
 
-### Migrate any schema changes
+### Changing the Database Schema
 
-If you make changes to the schema via Prisma you'll need to push it to the DB. We don't use Prisma migrations and let Planetscale handle it:
+Before making any changes to the schema, consider the impact those changes will have on a production system with preexisting live data. You may need to take a multi-step approach to get to your desired schema. [Prisma's Expand-and-Contract Pattern guide](https://www.prisma.io/dataguide/types/relational/expand-and-contract-pattern) is a good reference for what this multi-step approach might look like.
+
+To alter the schema, you'll need to first update `/packages/database/prisma/schema.prisma`.
+
+#### Migrate Schema Changes Locally
+
+When you make changes to the schema via Prisma you'll need to push it to the DB. We don't use Prisma migrations.
+
+You can test out your schema changes locally with the MySQL in Docker instance.
+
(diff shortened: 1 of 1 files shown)


apps/testingaccessibility/README.md
```diff
@@ -189,17 +191,71 @@ The `dev:stripe` node script is a shorthand for `stripe listen --forward-to loca
...
+
+```bash
+npx prisma db push
+```
+
+From here, you can connect to the MySQL database with a SQL client to view the changes or view it from _Prisma Studio_.
+
+#### Develop against the New Schema
+
+To develop against the new schema with the full power of Prisma's TypeScript typings, you'll need to re-generate the Prisma client for the updated schema. When doing this, if you switch partner products, the Prisma client will be pointing at the last partner product database until the client is regenerated.
```
jbranchaud, line 223: > When doing this, if you switch partner products, the Prisma client will be pointing at the last partner product database until the client is regenerated.

@garrettdimon As far as I can tell, this is only an issue when developing and running a data migration script. This level of detail might make more sense in a section that talks about data migrations.

joelhooks: I feel (not super strongly) that this is a red herring, and that the step here is to `validate the database_url` 

The Prisma client should generally be generated from the root of the monorepo using `pnpm -- turbo run generate` (which also runs with `turbo run build` as a dep) and when it's run in that way it uses a throw-away `database_url`

apps/testingaccessibility/README.md
```diff
@@ -189,17 +191,71 @@ The `dev:stripe` node script is a shorthand for `stripe listen --forward-to loca
...
+
+```bash
+npx prisma generate
```
joelhooks, line 226: I think this is where the issue is arising from

`*prisma*` is **hoisted**

https://github.com/skillrecordings/products/blob/092366b3e24f8a002b0a86d447bf864f64941f32/.npmrc#L2

https://pnpm.io/npmrc#public-hoist-pattern

🚨 I do not fully understand public hoist pattern mechanism

So this needs to be `pnpm -- turbo run generate` from the **root of the monorepo** to generate the client that can be used by all applications.

It is **not required** to run `npx prisma generate` from within specific apps, *though it will probably work* since it has an indicator to a schema in `package.json`

`npx primsa generate` is **problematic** since it uses the latest prisma, and might conflict.

the above describes my workflow that I've used switching between apps without issue

garrettdimon (approved):
📖 Yay for docs! Thanks for tweaking the bits I attempted to clarify.

joelhooks (approved):
🌀

PR #515, opened 2022-10-12
build: isolate turbo generate to database pkg

jbranchaud:
This tells turbo then whenever we run `turbo run generate`, it should
scope running it to just the `database` package. Because generate is
only running in `database`, it doesn't need the `dependsOn` anymore.

@joelhooks I think this does what we were trying to do on the huddle earlier.

generate


jbranchaud:
It looks like this only narrowly worked. It is breaking a bunch of other pipeline commands. 🤔

jbranchaud:
failed experiment, closing

PR #516, opened 2022-10-13
chore: bump next

vojtaholik:
gif


apps/engmanagement/src/pages/_app.tsx
```diff
@@ -21,7 +22,7 @@ declare global {
   }
 }
 
-function MyApp({Component, pageProps}: AppProps) {
+function MyApp({Component, pageProps}: AppProps<{session: Session}>) {
```
vojtaholik, line 25: this was throwing a type error without specifying `Session` type

joelhooks (approved):

PR #517, opened 2022-10-13
build: move turbo env var deps to the env key

jbranchaud:
The `$ENV_VAR` syntax in the `dependsOn` array has been deprecated and
you can instead declare those values in the `env` array.

The `env` key is available in turbo v1.4.7
(https://github.com/vercel/turborepo/releases/tag/v1.4.7).

turbo docs: https://turborepo.org/docs/reference/configuration#env

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -7,12 +7,11 @@
       "outputs": ["dist/**", ".next/**"]
     },
     "@skillrecordings/config#build": {
-      "dependsOn": [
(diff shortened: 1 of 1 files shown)


jbranchaud:
We are currently on turbo v1.3.1, so this isn't yet available to this project. Closing for now until we have a chance to get turbo bumped to the latest.

PR #524, opened 2022-10-17
feat(tt): disable castingwords order to test Whisper transcripts

Creeland:
For the next workshop, we are testing out a workflow where we order our transcripts using Whisper instead of Castingwords.

I've "unplugged" castingwords from the create sanity document handler so that we aren't paying for those transcripts to be done while we use Whisper.  

shh


apps/total-typescript/src/pages/api/webhooks/sanity/videoResource/created.ts
```diff
@@ -38,9 +39,12 @@ const sanityVideoResourceWebhook = async (
           playback_policy: ['public'],
         })
 
+        console.info('new mux asset created', muxAsset.id)
```
tayiorbeii, line 42: do you want this console.info?

tayiorbeii (approved):

PR #526, opened 2022-10-17
Discontinue reads and writes to `Coupon.bulkPurchaseId` column

jbranchaud:
This handles Step 5 (discontinue reads) and Step 6 (discontinue writes) of [EaC](https://www.prisma.io/dataguide/types/relational/expand-and-contract-pattern) to the `Coupon.bulkPurchaseId` column

- feat(ta): discontinue reads of bulkPurchaseId
- feat(ta): stop writing to bulkPurchaseId column

### What's Next

The follow-on to this PR will be Step 7, removing the old, no-longer-used structure from the schema.

1. This will involve removing `coupon.BulkPurchaseId`, `coupon.BulkPurchase`, and `purchase.IndividualBulkCoupon` from the `database` package's `schema.prisma` file.
2. These changes will need to be rolled out to each of the Planetscale databases using the schema.

disco


packages/commerce-server/src/record-new-purchase.ts
```diff
@@ -137,7 +137,6 @@ export async function recordNewPurchase(checkoutSessionId: string): Promise<{
         // 1. Create bulk coupon record
         const coupon = await transaction.coupon.create({
           data: {
-            bulkPurchaseId: purchase.id,
```
jbranchaud, line 140: This discontinues the writes, the changes in the other file remove the one place where this value was being read.

garrettdimon (approved):

PR #532, opened 2022-10-19
fix(pnpm): hoist pattern to be very specific

joelhooks:
This worked locally and Epic Web Dev ran and didn't have errors. Don't think the prisma update is a factor, but I thought I'd try explicitly hoisting the client in `.npmrc`

Seeing if it builds here.

hoist


joelhooks:
😭

PR #535, opened 2022-10-19
Use experimental corepack pkg manager pinning

jbranchaud:
This updates the root `package.json` to have a `packageManager` set to `pnpm@7.13.2` (as opposed to the absolute latest, `7.13.5`). That combined with setting `ENABLE_EXPERIMENTAL_COREPACK` to `1` (in Vercel's Env Var Settings) means that Vercel will use the `packageManager` version during builds. This allows our builds to go green.

We currently only have this `ENABLE_EXPERIMENTAL_COREPACK` setting for Preview builds on the `testing-accessibility` project. That was in order to try this out.

CleanShot 2022-10-19 at 11 59 39@2x

Before rolling out this PR, we'll want to apply this setting to other Vercel projects. And we'll need to add them as _Production_ env vars. And mention all of this in the docs.

CleanShot 2022-10-19 at 12 39 46@2x

green


jbranchaud:
We've seen this `ENABLE_EXPERIMENTAL_COREPACK` env var setting work for `testing-accessibility`. We have now added this env var setting for each of the other failing build apps (engmanagement, epic-web, and total-typescript) as _Preview_ env vars only. We're going to see if we can get those builds to go green in Preview as well.

jbranchaud:
Adding `NO MERGE` for now because we want to set these env vars in _Production_ before deploying these changes.

vojtaholik:
this seems to be working well! 🌮

epic react is now failing because of zod x sanity schema for which I've got a fix lined up in https://github.com/skillrecordings/products/pull/536 ([diff](https://github.com/skillrecordings/products/pull/536/files#diff-0c35df0dffc64cb2969f4363f360c091b1f2618d57f92c05b988b93d80c9b347R14))

vojtaholik (approved):
Tested this and everything seems to be back to normal 🎉 🌮  I enabled `ENABLE_EXPERIMENTAL_COREPACK` in production for all 4 projects that use db.

PR #539, opened 2022-10-21
build: check build status against pnpm 7.13.5

jbranchaud:
h


jbranchaud:
Fails as expected, closing

PR #540, opened 2022-10-21
build: check if pnpm 7.13.6 breaks the build

jbranchaud:
I just confirmed that pnpm 7.13.5 still breaks several of our builds (https://github.com/skillrecordings/products/pull/539). This is a secondary experiment to see if the now latest release (7.13.6) will also break it or not. 

maze

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "turbo": "~1.4.7",
     "typescript": "^4.8.3"
   },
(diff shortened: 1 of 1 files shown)


jbranchaud:
This also breaks the builds. We'll need to figure out how to move forward with `pnpm` at some point so that we aren't stuck pinned to `7.13.2` going forward.

PR #541, opened 2022-10-24
build: check if pnpm 7.13.4 break the builds

jbranchaud:
We have PRs with builds confirming that 7.13.5 and 7.13.6 of pnpm break our builds (https://github.com/skillrecordings/products/pull/540). This is to further narrow it down by checking 7.13.4.

check


joelhooks (approved):
can this be closed?

PR #518, opened 2022-10-13
feat(ta): add one-to-many relation for bulk coupons

jbranchaud:
This PR represents the second and third step of the Expand and Contract
pattern for replacing the existing one-to-one relation between Bulk
Coupon and Bulk Purchase with a one-to-many relation of Bulk Coupon to
Purchases.

Before this can be merged, the schema needs to be fully deployed to the
existing production database environments. Those Deploy Requests have
already been opened at:
1. https://app.planetscale.com/skill-recordings/testing-accessibility/deploy-requests/14
2. https://app.planetscale.com/skill-recordings/total-typescript/deploy-requests/3

Visual of the one-to-many relationship we're moving toward:

CleanShot 2022-10-17 at 10 06 45

bulk purchase

PR #486, opened 2022-09-27
feat(ta): `BuyMoreSeats` widget template

vojtaholik:
screenshot

example:
```jsx
import BuyMoreSeats from “components/team/buy-more-seats”

<Card
   title={{content: “Get more seats”, as: “h2”}}
   icon={
      <TicketIcon className="w-5 text-green-500" aria-hidden="true" />
   }
>
   <BuyMoreSeats />
</Card>
```

gif

PR #542, opened 2022-10-25
feat(ta): Allow Team Owner to buy more seats

jbranchaud:
We've updated our data model to allow for multiple bulk purchases, this
now uses that data model along with the `<BuyMoreSeats>` component to
provide a team owner with a way of tacking more seats on to an existing
Multi-Seat Team Account (bulk coupon).

Uses the `<BuyMoreSeats>` component from this PR: https://github.com/skillrecordings/products/pull/486

Builds on the data model work in:
- https://github.com/skillrecordings/products/pull/518
- https://github.com/skillrecordings/products/pull/526

Here is what this flow looks like:

https://user-images.githubusercontent.com/694063/197841858-c33ac93a-8314-4da6-8edb-6fb24aa32b2f.mp4

### Next Steps

- [ ] Don't send the magic link email to an existing user who is adding more seats, they are already signed in.
- [ ] Where to redirect the user after adding to their bulk purchase? The wording of the 'Thanks For Your Purchase' page doesn't make sense for this flow. https://github.com/skillrecordings/products/blob/main/packages/skill-api/src/core/services/stripe-checkout.ts#L169-L171

more seats


apps/testingaccessibility/src/components/team/buy-more-seats.tsx
```diff
@@ -1,21 +1,84 @@
...
+
+  const formActionPath = buildFormActionPath({
+    userId,
+    quantity: debouncedQuantity,
+    productId,
+  })
 
   return (
-    <form className="pt-3" onSubmit={handleOnSubmit}>
+    <form className="pt-3" action={formActionPath} method="POST">
```
jbranchaud, line 67: The 'Buy' button POSTs the form to our stripe checkout API which facilitates the transition to the Stripe Checkout Session.

apps/testingaccessibility/src/components/team/buy-more-seats.tsx
```diff
@@ -1,21 +1,84 @@
...
+        }),
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      })
+        .then((res) => res.json())
+        .then((formattedPrice: FormattedPrice) => {
+          return formattedPrice
+        }),
+  )
```
jbranchaud, line 50: This fetches pricing data similar to the Buy page, but the `productId` is restricted to the product ID of their existing bulk purchase.

I also stripped out the `couponId` and `upgradeFromPurchaseId` query params from this path because I don't think they are relevant to an additional bulk purchase.

https://github.com/skillrecordings/products/blob/main/apps/testingaccessibility/src/components/pricing.tsx#L192-L204

CleanShot 2022-10-25 at 13 45 52@2x

apps/testingaccessibility/src/components/team/buy-more-seats.tsx
```diff
@@ -1,21 +1,84 @@
...
+  const {productId, quantity, userId} = params
+
+  const queryParamString = new URLSearchParams({
+    productId,
+    quantity: String(quantity),
+    userId,
+  }).toString()
+
+  return `/api/skill/checkout/stripe?${queryParamString}`
+}
```
jbranchaud, line 21: build the path's query params with `URLSearchParams` because it is easier to read than string interpolation, less error prone, and offers a bit more type-safety.

apps/testingaccessibility/src/pages/team/index.tsx
```diff
@@ -95,6 +99,14 @@ const TeamPage: React.FC<React.PropsWithChildren<TeamPageProps>> = ({
...
           />
         </Card>
+        <Card
+          title={{content: 'Get more seats', as: 'h2'}}
+          icon={
+            <TicketIcon className="w-5 text-green-500" aria-hidden="true" />
+          }
+        >
+          <BuyMoreSeats productId={purchase.product.id} userId={userId} />
+        </Card>
```
jbranchaud, line 109: From the team page, render the `BuyMoreSeats` component with the `productId` of the existing purchase and this user's `id`.

I think we are fine to only render this here and not also on the welcome page which looks very similar. That feels unnecessarily pushy to have it there too when they have _just_ made their initial purchase.

packages/commerce-server/src/record-new-purchase.ts
```diff
@@ -126,23 +126,50 @@ export async function recordNewPurchase(checkoutSessionId: string): Promise<{
...
+      },
+    },
+  })
 
+  // Note: if the user already has a bulk purchase/coupon, then if they are
+  // only adding 1 seat to the team, then it is still a "bulk purchase" and
+  // we need to add it to their existing Bulk Coupon.
+  const isBulkPurchase = quantity > 1 || !!existingBulkCoupon
+
+  if (purchase && isBulkPurchase) {
```
jbranchaud, line 147: Here in the `record-new-purchase` function (which gets called by the Stripe webhook handler for the `checkout.session.completed` event), we can look up whether this user already has made a Bulk Coupon purchase. If they have, then whether they are adding `1` seat or `10` seats, we want to treat this as an _additional_ bulk purchase to tack on to the existing coupon.

packages/commerce-server/src/record-new-purchase.ts
```diff
@@ -126,23 +126,50 @@ export async function recordNewPurchase(checkoutSessionId: string): Promise<{
...
+        let coupon = null
+        if (existingBulkCoupon) {
+          coupon = await transaction.coupon.update({
+            where: {
+              id: existingBulkCoupon.id,
+            },
+            data: {
+              maxUses: existingBulkCoupon.maxUses + quantity,
+            },
+          })
```
jbranchaud, line 162: If the user already has an existing bulk coupon, then we add the quantity to that coupon.

apps/testingaccessibility/src/components/team/buy-more-seats.tsx
joelhooks, line 50: https://github.com/skillrecordings/products/blob/main/apps/testingaccessibility/src/components/pricing.tsx#L53-L80

I'd consider making a shared hook + lib (service) here with an "option bag" config since it's very very similar as mentioned below.

apps/testingaccessibility/src/pages/team/index.tsx
joelhooks: ya, we don't want to sell them, but we should make sure to mention how they can add more seats via email! 

a future iteration could have this available for individuals as well so they can "upgrade" to a bulk account. we've definitely had folks do that!

packages/commerce-server/src/record-new-purchase.ts
joelhooks: I think we need to get jest involved around this stuff

it probably makes sense to scaffold integration tests

joelhooks (approved):
comments are for future work, looks good

packages/commerce-server/src/record-new-purchase.ts
jbranchaud: @garrettdimon and I just discussed the potential of some package-level Jest tests for this kind of thing as well as app-level e2e tests in TA.

PR #548, opened 2022-10-26
feat(ta): make post-purchase communication aware of the purchase type

jbranchaud:
There are three types of purchases: a individual purchase, a new bulk/team purchase, and a purchase that adds seats to an existing bulk/team purchase.

This is a follow-on PR to https://github.com/skillrecordings/products/pull/542

1. Only send the Welcome email with magic link when it is one of the first two types (new individual purchase or new bulk purchase)
2. Adjust the "Thank You for Your Purchase" page copy based on each of these types.

CleanShot 2022-10-28 at 11 42 10@2x

post purchase


packages/types/src/index.ts
```diff
@@ -125,3 +125,7 @@ export type Choice = {
   label?: string
   image?: string
 }
+
+export const EXISTING_BULK_COUPON = 'EXISTING_BULK_COUPON' as const
+export const NEW_BULK_COUPON = 'NEW_BULK_COUPON' as const
+export const NEW_INDIVIDUAL_PURCHASE = 'NEW_INDIVIDUAL_PURCHASE' as const
```
jbranchaud, line 131: These don't really feel to me like they belong here in the `@skillrecordings/types`, but I couldn't find a better place for them. Both `commerce-server` and app client code (e.g. TA Thank You page) use these values.

What I'm imagining is a place where shared constants and Zod schemas can live, where both server and client code and import them. A schema or validation package perhaps, but there isn't enough of an emerging pattern yet to justify a new package.

Any better ideas?

apps/testingaccessibility/src/pages/thanks/purchase.tsx
```diff
@@ -25,14 +32,54 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
...
+  let purchaseType: z.infer<typeof purchaseTypeSchema>
+  if (purchase.bulkCoupon) {
+    if (purchase.bulkCoupon.bulkCouponPurchases.length > 1) {
+      purchaseType = EXISTING_BULK_COUPON
+    } else {
+      purchaseType = NEW_BULK_COUPON
+    }
+  } else {
+    purchaseType = NEW_INDIVIDUAL_PURCHASE
+  }
```
jbranchaud, line 76: Determine what type of purchase this checkout session was for. This `purchaseType` value can be used by the component to adjust the copy accordingly.

apps/testingaccessibility/src/pages/thanks/purchase.tsx
```diff
@@ -65,30 +116,48 @@ const ThanksVerify: React.FC<React.PropsWithChildren<ThanksProps>> = ({
...
             </h1>
             {isTeamPurchase && (
               <p className="text-sand-100 max-w-md font-medium leading-relaxed mx-auto">
-                Your purchase is for <strong>{seatsPurchased}</strong> seats.
-                You can always add more seats later when your team grows.
+                Your purchase is for <strong>{seatsPurchased}</strong>{' '}
+                {purchaseType === EXISTING_BULK_COUPON && 'additional'} seats.
+                {purchaseType === NEW_BULK_COUPON &&
+                  'You can always add more seats later when your team grows.'}
               </p>
```
jbranchaud, line 129: I took some liberty fine-tuning the copy based on purchase type. Someone with stronger opinions should feel free to take another pass on it.

I'm also interested to hear opinions on whether inlining a bunch of conditions is sufficient here or if we should break out separate components and/or blocks of text.

packages/commerce-server/src/record-new-purchase.ts
```diff
@@ -68,10 +74,18 @@ export type PurchaseInfo = {
   stripeProduct: Stripe.Product
 }
 
+export const purchaseTypeSchema = z.union([
+  z.literal(EXISTING_BULK_COUPON),
+  z.literal(NEW_BULK_COUPON),
+  z.literal(NEW_INDIVIDUAL_PURCHASE),
+])
+type PurchaseType = z.infer<typeof purchaseTypeSchema>
```
jbranchaud, line 82: This gets used both here and in `getServerSideProps` of `apps/testingaccessibility`, so I'd be interested in finding a shared place for this to live outside the `commerce-server`.

Related to what I'm saying in this comment: https://github.com/skillrecordings/products/pull/548/files#r1007082578

apps/testingaccessibility/src/pages/thanks/purchase.tsx
```diff
@@ -65,30 +116,53 @@ const ThanksVerify: React.FC<React.PropsWithChildren<ThanksProps>> = ({
...
+                </h2>
+                <p className="text-sand-100 max-w-md font-medium leading-relaxed mx-auto">
+                  Your team invite share link is the same as before. Visit the{' '}
+                  <Link href="/team/invite">
+                    <a className="py-1 inline-flex text-base font-medium hover:underline transition">
+                      Team Invite
+                    </a>
+                  </Link>{' '}
+                  page to get the share link for distributing to your team.
+                </p>
```
joelhooks, line 163: can we just display the link here in both circumstances?

packages/commerce-server/src/record-new-purchase.ts
joelhooks: if it's "just a type" the `types` packages definitely makes sense to me

packages/types/src/index.ts
joelhooks: I'd probably use a PurchaseType enum for these versus this esoteric const const scheme and don't think the `types` package is an inappropriate spot. If you wanted more specific granularity, a `commerce-types` package would be a great choice too. We don't have to be sparing with the packages. We can eventually create a `commerce` top level workspace/namespace (for instance) if we want to organize/bundle related stuff

joelhooks:
```
enum PurchaseType {
    EXISTING_BULK_COUPON,
    NEW_BULK_COUPON,
    NEW_INDIVIDUAL_PURCHASE
}
```

results in:

```
var PurchaseType;
(function (PurchaseType) {
    PurchaseType[PurchaseType["EXISTING_BULK_COUPON"] = 0] = "EXISTING_BULK_COUPON";
    PurchaseType[PurchaseType["NEW_BULK_COUPON"] = 1] = "NEW_BULK_COUPON";
    PurchaseType[PurchaseType["NEW_INDIVIDUAL_PURCHASE"] = 2] = "NEW_INDIVIDUAL_PURCHASE";
})(PurchaseType || (PurchaseType = {}));
```

and is WAY easier for me to read than the `const const zod` and the enum is simple to use as a type

joelhooks (commented):
typing suggestions and package suggestions totally optional, but can we add the url to distribute to their team on the thanks page?

jbranchaud:
@joelhooks I've updated the `/thanks/purchase` page to include the Invite Link for an existing bulk purchase. See screenshot in the main PR body. I'll look into the typing and package suggestions outside of this PR so as to not hold it up.

@vojtaholik not block on this PR, but curious if you have thoughts on whether the Email Inbox image makes sense for this _existing bulk purchase_ flow since we aren't sending an email in that scenario.

vojtaholik:
@jbranchaud lets replace the email image with the notebook one:
welcome page screenshot

would it possible to re-use this component with the invite link? :)
team ui screenshot

joelhooks (approved):

PR #554, opened 2022-10-28
fix(kl) remove CK constants to fall back to legacy subscriptions

tayiorbeii:
Same as #547 but for Keyboard Legend:

> This removes the environment variables that act as a switch to use our newer ConvertKit package. Doing that makes them fall back to the legacy subscription logic.

Homer's Sandwich


joelhooks:
the solution for this is actually setting `ENABLE_EXPERIMENTAL_COREPACK` as `1` in vercel env vars

PR #558, opened 2022-11-01
fix(ta): cannot serialize undefined, pass null

jbranchaud:
Post Purchase page was broken for individual purchases because Next.js cannot serialize `undefined`. Pass the value down as `null` instead.

null


garrettdimon (approved):
⚡

PR #559, opened 2022-11-01
build(tests): add e2e tests to github actions

garrettdimon:
This sets up the Playwright e2e tests to run via GitHub Actions.

verify


garrettdimon:
Good proof-of-concept, but we'll need something more robust to run the e2e tests against multiple apps each with a different `baseUrl`

PR #561, opened 2022-11-01
feat(ta): use entire team invite component on post-purchase page

jbranchaud:
@vojtaholik requested that the entire `TeamInvite` component be used on the post-purchase page for consistency with the Team Invite page. This PR achieves that by doing a little extra data fetching.

This pairs will with allowing team owners to purchase more seats because the first page they see after their purchase has a copiable link to share with their team. 

Here is what that page will look like:

CleanShot 2022-11-01 at 15 27 29@2x

the bear

### DIFF
--- a/apps/testingaccessibility/src/components/team/index.tsx
+++ b/apps/testingaccessibility/src/components/team/index.tsx
@@ -9,14 +9,14 @@ type InviteTeamProps = {
     product: {id: string; name: string}
   }
   existingPurchaseForSelf: boolean
-  session: any
+  userEmail?: string | null
   setPersonalPurchase: (props: any) => void
 }
 
 const InviteTeam: React.FC<React.PropsWithChildren<InviteTeamProps>> = ({
   purchase,
   existingPurchaseForSelf,
-  session,
+  userEmail,
   setPersonalPurchase,
 }) => {
   const redemptionsLeft =
@@ -29,7 +29,6 @@ const InviteTeam: React.FC<React.PropsWithChildren<InviteTeamProps>> = ({
   const [canRedeem, setCanRedeem] = React.useState(
     Boolean(redemptionsLeft && !existingPurchaseForSelf),
   )
-  const userEmail = session?.user?.email
   const bulkCouponId = purchase?.bulkCoupon?.id
 
   return (
--- a/apps/testingaccessibility/src/components/team/index.tsx
+++ b/apps/testingaccessibility/src/components/team/index.tsx
@@ -1,7 +1,6 @@
 import React from 'react'
 import SelfRedeemButton from './self-redeem-button'
 import CopyInviteLink from './copy-invite-link'
-import Link from 'next/link'
 
 type InviteTeamProps = {
   purchase: {
@@ -65,11 +64,10 @@ const InviteTeam: React.FC<React.PropsWithChildren<InviteTeamProps>> = ({
       )}
       {!redemptionsLeft && (
         <div className="flex items-center justify-between mt-5 pt-5 border-t border-gray-100">
-          <Link href="/#buy">
-            <a className="font-semibold bg-green-500 transition text-white px-4 py-2 hover:bg-green-600 rounded-md flex-shrink-0">
-              Buy more seats
-            </a>
-          </Link>
+          <p className="py-3">
+            You are out of seats. To invite more people to your team, buy more
+            seats.
+          </p>
         </div>
       )}
     </>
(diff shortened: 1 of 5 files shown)


apps/testingaccessibility/src/components/team/index.tsx
```diff
@@ -68,11 +67,10 @@ const InviteTeam: React.FC<React.PropsWithChildren<InviteTeamProps>> = ({
...
         <div className="flex items-center justify-between mt-5 pt-5 border-t border-gray-100">
-          <Link href="/#buy">
-            <a className="font-semibold bg-green-500 transition text-white px-4 py-2 hover:bg-green-600 rounded-md flex-shrink-0">
-              Buy more seats
-            </a>
-          </Link>
+          <p className="py-3">
+            You are out of seats. To invite more people to your team, buy more
+            seats.
+          </p>
```
jbranchaud, line 73: This 'Buy more seats' link doesn't make sense anymore now that the `<BuyMoreSeats />` component is going to display right below this component on the Team Invite page.

apps/testingaccessibility/src/components/team/index.tsx
```diff
@@ -1,28 +1,28 @@
...
-  session,
-  setPersonalPurchase,
+  existingPurchaseForSelf,
+  userEmail,
 }) => {
+  const [newSelfRedemption, setNewSelfRedemption] = React.useState<
+    object | undefined
+  >()
+
+  const hasPurchasedForSelf = existingPurchaseForSelf || newSelfRedemption
```
jbranchaud, line 24: Handle the "did you just redeem the purchase for yourself?" state local to this component, instead of expecting the parent component to pass down a handler.

apps/testingaccessibility/src/components/team/index.tsx
```diff
@@ -1,28 +1,28 @@
...
     product: {id: string; name: string}
   }
-  existingPurchase: {
-    id: string
-    product: {id: string; name: string}
-  }
-  session: any
-  setPersonalPurchase: (props: any) => void
+  existingPurchaseForSelf: boolean
+  userEmail?: string | null
```
jbranchaud, line 12: Tighten the type signature a bit, accept a `userEmail` instead of a whole `session` blob, for instance.

packages/commerce-server/src/prisma-next-serializer.ts
```diff
@@ -8,6 +8,8 @@ export function convertToSerializeForNextResponse(result: any) {
       result[resultKey] = result[resultKey].toISOString()
     } else if (result[resultKey]?.constructor?.name === 'Decimal') {
       result[resultKey] = result[resultKey].toNumber()
+    } else if (result[resultKey] instanceof Object) {
+      result[resultKey] = convertToSerializeForNextResponse(result[resultKey])
```
jbranchaud, line 12: Added this recursive case for when the Prisma object to be serialized had included a relation, the nested relation object needs to be appropriately serialized as well.

apps/testingaccessibility/src/pages/thanks/purchase.tsx
```diff
@@ -161,24 +184,23 @@ const ThanksVerify: React.FC<React.PropsWithChildren<ThanksProps>> = ({
...
+                    />
+                  }
+                >
+                  <InviteTeam
+                    userEmail={email}
+                    purchase={bulkCouponPurchase}
+                    existingPurchaseForSelf={!!selfRedeemedPurchase}
+                  />
+                </Card>
+              </div>
```
jbranchaud, line 203: Here is the `<InviteTeam>` component being displayed on the post-purchase page.

For its default styles to make sense, I had to write it in a card. I also had to wrap a `<div>` around the whole thing with `text-gray-900` or else the text color was rendering as white.

apps/testingaccessibility/src/pages/thanks/purchase.tsx
```diff
@@ -76,11 +85,23 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
...
 
+  let bulkCouponRedemptionPurchase: object | null = null
+  if (purchase.bulkCoupon?.id) {
+    const {getUserByEmail, getBulkCouponRedemptionPurchase} = getSdk()
+    const user = await getUserByEmail(email)
+    bulkCouponRedemptionPurchase = await getBulkCouponRedemptionPurchase(
+      purchase.bulkCoupon.id,
+      user?.id || null,
+    )
+  }
```
jbranchaud, line 96: Use a couple database functions from `getSdk` to look up the user ID (they may not be logged in at this point) and then determine if this person has self-redeemed one of the seats.

jbranchaud:
This PR ended up touching a lot more files than I expected and was still failing in places. I'm going to close this for now. I think it'll be better to revisit this once we have this commerce flow running in a couple apps. I couldn't quite get my head around the right abstraction as is.

PR #566, opened 2022-11-02
feat(ta): better handle failed self-redemption

jbranchaud:
If the self-redemption of a team seat fails, the loading spinner just
keeps spinning and there is no info about what is going. This clears the
loading spinner and provides the user with some info about what they can
do next.

self redeem


apps/testingaccessibility/src/utils/handle-self-redeem.ts
```diff
@@ -2,22 +2,28 @@
  * handle self redeeming of a seat in bulk purchase
  * @param {string} email current user's email from session
  * @param {string} bulkCouponId current user's bulk coupon id from purchase
- * @param {(redeemPurchase) => void} onSuccess callback to be called when purchase is redeemed
+ * @param {(params) => void} callback function to be called after attempting coupon redemption
```
jbranchaud, line 5: Replaces `onSuccess` with generic `callback` handler which responds with either `status: 'success'` or `status: 'failed'`. The caller than either has the purchase object on success or an error message on failure and can handle those scenarios appropriately.

apps/testingaccessibility/src/utils/handle-self-redeem.ts
```diff
@@ -2,22 +2,28 @@
...
   const redeemedPurchase = await redeemFullPriceCoupon({
     email,
     couponId: bulkCouponId,
     sendEmail: false,
   })
   if (redeemedPurchase && !redeemedPurchase.error) {
-    onSuccess(redeemedPurchase)
+    callback({status: 'success', redeemedPurchase})
+  } else {
+    callback({status: 'failed', error: redeemedPurchase.message})
```
jbranchaud, line 27: Pass back what the caller needs to handle success and failure scenarios

apps/testingaccessibility/src/components/team/self-redeem-button.tsx
```diff
@@ -25,8 +26,23 @@ const SelfRedeemButton: React.FC<
...
+                  'You have already redeemed a seat for yourself. Please contact support if you are having trouble accessing it.',
+                )
+              } else {
+                toast.error(
+                  'We were unable to redeem a seat for this account. If the issue persists, please reach out to support.',
+                )
+              }
+            }
+
             setIsLoading(false)
```
jbranchaud, line 46: If successful, call the `onSuccess` handler.

If failed, then surface some info to the user so that they know what is going on.

Lastly, always setting loading back to `false` so that the spinner doesn't sit there indefinitely.

joelhooks (approved):

PR #569, opened 2022-11-02
types(ta): consolidate dup types across components

jbranchaud:
d


apps/testingaccessibility/src/@types/index.ts
```diff
@@ -0,0 +1,10 @@
+export type PurchaseWithBulkCoupon = {
+  merchantChargeId: string | null
+  bulkCoupon: {id: string; maxUses: number; usedCount: number} | null
+  product: {id: string; name: string}
+}
+
+export type MinimalPurchase = {
```
joelhooks, line 7: up until this PR the `@types` folder's job was to house missing types from imported node dependencies

in total typescript my strategy has been to move types into `/lib` close to the incoming instance of whatever the type is supporting in a "let's build an sdk" fashion. this is as opposed to "lets dump stuff in folders by type of file" spreading it around more.

in this case I think these actually belong in the database or commerce packages, or otherwise closer to the "source" instead of this overflow location for external types that we are forced to include

PR #570, opened 2022-11-03
feat(ta): allow Individual to upgrade to Team

jbranchaud:
This feature was less involved than I originally anticipated. I thought it would require schema changes. It doesn't. And because of how we are already processing new and existing bulk purchases, this page can handle seat buying for both individuals and existing teams.

Roam Notes: https://roamresearch.com/#/app/egghead/page/bIpYsr441

---

This adds a page that a logged in user with a purchase can visit. It has
a 'Buy More Seats' component that takes the user to a Stripe Checkout
Session. Once they complete the purchase, a new Bulk Coupon will be
created for the user with that many seats. It doesn't impact their
existing individual purchase, which we maintain for record keeping and
to determine this user's access.

There is no top-level UI for accessing this page. Support can link a person directly to the `/team/buy-more-seats` page if they are wanting to upgrade from Individual to Team.

Here is the page:

CleanShot 2022-11-03 at 12 56 09@2x

Notice, after they complete the Stripe checkout session, they are dropped onto this post-purchase page.

CleanShot 2022-11-03 at 12 40 53@2x

individual

### DIFF
--- a/apps/testingaccessibility/src/pages/team/buy-more-seats.tsx
+++ b/apps/testingaccessibility/src/pages/team/buy-more-seats.tsx
@@ -1,0 +1,80 @@
+import React from 'react'
+import Layout from 'components/app/layout'
+import {convertToSerializeForNextResponse} from '@skillrecordings/commerce-server'
+import {GetServerSideProps} from 'next'
+import BuyMoreSeats from 'components/team/buy-more-seats'
+import {TicketIcon} from '@heroicons/react/outline'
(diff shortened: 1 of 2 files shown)


garrettdimon:
The text says "Please check your inbox for a login link that just got sent." but in the case of the bulk purchase, that feels inaccurate. For a bulk purchase, the CTA should be more along the lines of "go here to send the invites to your team members", right?

joelhooks (approved):

PR #586, opened 2022-11-08
feat: creating a stripe sdk for simplified mocking

joelhooks:
[⭐️ long 45m end-to-end loom video of creating this PR](https://www.loom.com/share/28d14ea87b7b410092abd70d74d8bcb6)

Screen Shot 2022-11-08 at 10 14 09

we are exploring how to mock out services to test some of our commerce workflows. In this PR, I looked at how we were mocking `prisma` in the `@skillrecordings/database` package utilizing a database SDK that wraps `prisma` in a lightweight typed mockable context using `jest-mock-extended`

This PR introduces `@skillrecordings/stripe-sdk` which provides a similar mockable context for testing stripe without using a heavier solution for now using simple data objects as mocks.

We should consider moving **all `stripe` calls** into this SDK instead of accessing the `stripe` library across the entire application and package surface area.

IMO, this is true for these service boundaries in general. They get wrapped.

wrap it up

### DIFF
--- a/packages/commerce-server/src/record-new-purchase.ts
+++ b/packages/commerce-server/src/record-new-purchase.ts
@@ -1,13 +1,13 @@
 import {Stripe} from 'stripe'
 import {first} from 'lodash'
 import {type Purchase, prisma, getSdk} from '@skillrecordings/database'
-import {stripe} from './stripe'
 import {z} from 'zod'
 import {
   EXISTING_BULK_COUPON,
   NEW_BULK_COUPON,
   NEW_INDIVIDUAL_PURCHASE,
 } from '@skillrecordings/types'
+import {getStripeSdk} from '@skillrecordings/stripe-sdk/src'
 
 export class PurchaseError extends Error {
   checkoutSessionId: string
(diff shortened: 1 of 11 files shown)


packages/commerce-server/package.json
```diff
@@ -17,6 +17,7 @@
   "dependencies": {
     "@casl/ability": "^6.0.0",
     "@skillrecordings/database": "workspace:*",
+    "@skillrecordings/stripe-sdk": "workspace:*",
```
joelhooks, line 20: _a new package has entered chat_

packages/commerce-server/src/record-new-purchase.ts
```diff
@@ -29,16 +29,9 @@ export class PurchaseError extends Error {
...
-      expand: [
-        'customer',
-        'line_items.data.price.product',
-        'payment_intent.charges',
-      ],
-    },
-  )
+  const {getCheckoutSession} = getStripeSdk()
+
+  const checkoutSession = await getCheckoutSession(checkoutSessionId)
```
joelhooks, line 34: we should refactor to this style for accessing service APIs through 2nd/3rd party libs imo

packages/stripe-sdk/src/stripe-sdk.ts
```diff
@@ -0,0 +1,19 @@
...
+      return await ctx.stripe.checkout.sessions.retrieve(checkoutSessionId, {
+        expand: [
+          'customer',
+          'line_items.data.price.product',
+          'payment_intent.charges',
+        ],
+      })
+    },
+  }
+}
```
joelhooks, line 19: by default the context carries the normal `stripe` client instance, but we can also mock it

not I didn't write a test and exercise the mock context to see if it works as expected, but hand rolling a mock is also simple and can **iterate into end game automated solution** as needed (versus aiming for that now)

jbranchaud (approved):
🚀

PR #587, opened 2022-11-08
feat(ta): Display bulk coupon messaging when upgrading from Individual to Bulk

jbranchaud:
Primarily, this PR adds an additional purchase type (`INDIVIDUAL_TO_BULK_UPGRADE`) so that we can differentiate between a standard new bulk purchase and a bulk purchase that is an upgrade from an individual purchase.

This allows the server-code to know not to send a Welcome email. And it allows the client-code to display appropriate copy.

CleanShot 2022-11-09 at 09 44 00@2x

This PR also introduces a `commerce-server` unit test that does mocking of both Stripe and the database/prisma calls. There were a lot of small details to iron out to get this working, but this should now provide a working template for further testing in this package.

determine


packages/commerce-server/src/determine-purchase-type.ts
```diff
@@ -0,0 +1,121 @@
...
+  return Boolean(value1 && value2 && value1 === value2)
+}
+
+type DeterminePurchaseTypeOptions = {
+  checkoutSessionId: string
+  prismaCtx?: Context
+  stripeCtx?: StripeContext
+}
+
+export async function determinePurchaseType(
```
jbranchaud, line 42: This is the headline function/module -- there were a couple places that were running bespoke queries and lines of logic to determine the type of purchase that had been made given a checkout session ID.

This function is now the source of truth for making that determination.

It also uses the `packages/database` `getSdk` for all prisma calls and `packages/stripe-sdk` for the Stripe call. This makes it testable with mocks (see the accompanying test file).

packages/commerce-server/src/determine-purchase-type.test.ts
```diff
@@ -0,0 +1,84 @@
...
+function mockIndividualPurchase() {
+  // mock Stripe data for:
+  // const purchaseInfo = await stripeData(checkoutSessionId as string)
+  // const {email, stripeChargeId} = purchaseInfo
+  // no existing purchases for user
+  const newPurchase = {
+    id: 'purchase_123',
+    bulkCoupon: null,
+    redeemedBulkCouponId: null,
+  } as Purchase & {bulkCoupon: Coupon | null}
```
jbranchaud, line 77: Because the actual prisma query selects for a `Purchase` but then includes the `bulkCoupon` relation, the type of this mocked purchase needs to be `Purchase` _plus_ the potential presence of a `bulkCoupon: Coupon`.

packages/commerce-server/src/record-new-purchase.ts
```diff
@@ -28,8 +28,14 @@ export class PurchaseError extends Error {
...
-export async function stripeData(checkoutSessionId: string) {
-  const {getCheckoutSession} = getStripeSdk()
+type StripeDataOptions = {
+  checkoutSessionId: string
+  stripeCtx?: StripeContext
+}
+
+export async function stripeData(options: StripeDataOptions) {
+  const {stripeCtx, checkoutSessionId} = options
+  const {getCheckoutSession} = getStripeSdk({ctx: stripeCtx})
```
jbranchaud, line 38: Updates the `stripeData` function to receive `options` instead of just the `checkoutSessionId` so that an optional `stripeContext` can be passed in. We can pass that down to `getStripeSdk` regardless of whether it is defined, because if it is `undefined`, it will be overridden with the default context.

packages/types/src/index.ts
```diff
@@ -126,6 +126,7 @@ export type Choice = {
   image?: string
 }
 
-export const EXISTING_BULK_COUPON = 'EXISTING_BULK_COUPON' as const
-export const NEW_BULK_COUPON = 'NEW_BULK_COUPON' as const
-export const NEW_INDIVIDUAL_PURCHASE = 'NEW_INDIVIDUAL_PURCHASE' as const
+export const EXISTING_BULK_COUPON = 'EXISTING_BULK_COUPON'
+export const NEW_BULK_COUPON = 'NEW_BULK_COUPON'
+export const NEW_INDIVIDUAL_PURCHASE = 'NEW_INDIVIDUAL_PURCHASE'
+export const INDIVIDUAL_TO_BULK_UPGRADE = 'INDIVIDUAL_TO_BULK_UPGRADE'
```
jbranchaud, line 132: Add new `INDIVIDUAL_TO_BULK_UPGRADE` type alongside the others. And removing the unnecessary `as const` casting.

There was a comment left on these lines in another PR about making them an enum -- I curious to hear more about that, I haven't used enums at all up to this point. I'd prefer to explore that outside this PR tho so as to avoid getting sidetracked from deploying the fix to the post-purchase page.

apps/engmanagement/src/pages/welcome/index.tsx
```diff
@@ -21,7 +21,9 @@ export const getServerSideProps: GetServerSideProps = async ({req, query}) => {
   let purchaseId = purchaseQueryParam
 
   if (session_id) {
-    const {stripeChargeId} = await stripeData(session_id as string)
+    const {stripeChargeId} = await stripeData({
+      checkoutSessionId: session_id as string,
+    })
```
jbranchaud, line 26: One of several examples of `stripeData` and `recordNewPurchase` switching to an `options` object for parameters to support passing in Stripe and Prisma `ctx`.

apps/testingaccessibility/src/pages/thanks/purchase.tsx
```diff
@@ -102,7 +102,9 @@ const ThanksVerify: React.FC<React.PropsWithChildren<ThanksProps>> = ({
   bulkCouponId,
 }) => {
   const isTeamPurchase =
-    purchaseType === NEW_BULK_COUPON || purchaseType === EXISTING_BULK_COUPON
+    purchaseType === NEW_BULK_COUPON ||
+    purchaseType === EXISTING_BULK_COUPON ||
+    purchaseType === INDIVIDUAL_TO_BULK_UPGRADE
   const isNewPurchase =
     purchaseType === NEW_BULK_COUPON || purchaseType === NEW_INDIVIDUAL_PURCHASE
```
jbranchaud, line 109: Here is the change that gets the post-purchase page displaying the right thing when a user is upgrading from individual to bulk/team.

packages/commerce-server/src/determine-purchase-type.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    mockIndividualPurchase()
+
+    const purchaseType = await determinePurchaseType({
+      checkoutSessionId: 'cs_123',
+      prismaCtx: mockCtx,
+      stripeCtx: stripeMockCtx,
+    })
+
+    expect(purchaseType).toBe(NEW_INDIVIDUAL_PURCHASE)
+  })
```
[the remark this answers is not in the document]
joelhooks: 🎉🎉🎉

love to see it.

packages/commerce-server/src/determine-purchase-type.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+function mockStripePurchaseInfo() {
+  // Needed values: email, stripeChargeId
+  const purchaseInfo = {
+    customer: {email: 'customer@example.com'} as Stripe.Customer,
+    line_items: {data: [{price: {product: {}}}]} as Stripe.LineItemListParams,
+    payment_intent: {
+      charges: [{id: 'ch_123'}] as Stripe.ChargeListParams,
+    } as Stripe.PaymentIntent,
+  }
```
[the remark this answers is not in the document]
joelhooks: good info, makes sense

joelhooks (approved):

PR #590, opened 2022-11-09
feat(db): move bulk coupon create into SDK

jbranchaud:
The bulk coupon always gets created along with the purchase if it is a
purchase of a bulk coupon (or bulk coupon addition), so it makes sense
to push that up into the `createMerchantChargeAndPurchase` function.

This has the added benefit of removing the remaining direct Prisma DB
calls from the recordNewPurchase module.

danny dvito


joelhooks (approved):
love to see it

PR #592, opened 2022-11-09
feat(tt): enable pricing lookup on the /buy page

joelhooks:
full path to purchase from /buy to /thanks/purchase

[loom walkthrough click testing](https://www.loom.com/share/a7411a11245d4e9995b6cbfd656b46df)

buy now


apps/total-typescript/src/components/team-invites.tsx
```diff
@@ -0,0 +1,100 @@
...
+              )
+            })}
+          </div>
+        </div>
+      )}
+    </>
+  )
+}
+
+export default TeamInvites
```
jbranchaud, line 100: This component isn't getting used anywhere. It looks like it was probably brought over from Eng. Management where it also wasn't being used. It should be safe to remove this for now.

apps/total-typescript/src/components/team-invites.tsx
```diff
@@ -0,0 +1,100 @@
+import * as React from 'react'
+import {format, parseISO} from 'date-fns'
+import get from 'lodash/get'
+import isEmpty from 'lodash/isEmpty'
+import useClipboard from 'react-use-clipboard'
+import Tippy from '@tippyjs/react'
```
jbranchaud, line 6: if we remove this component, then we can remove the Tippy dependency as well

apps/total-typescript/src/components/team/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+    purchase.bulkCoupon.maxUses > purchase.bulkCoupon.usedCount
+  const numberOfRedemptionsLeft =
+    purchase?.bulkCoupon &&
+    purchase?.bulkCoupon.maxUses - purchase.bulkCoupon.usedCount
+
+  const [canRedeem, setCanRedeem] = React.useState(
+    Boolean(redemptionsLeft && !existingPurchase),
+  )
+  const userEmail = session?.user?.email
+  const bulkCouponId = purchase?.bulkCoupon?.id
```
jbranchaud, line 37: *not this PR*: I'd be interested in finding a `packages` place for this kind of logic, since it is now spread across a couple apps, I wonder if it some of this could be _computed values_ provided through the DB SDK when getting bulk coupon info.

apps/total-typescript/src/components/team/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+    bulkCoupon: {id: string; maxUses: number; usedCount: number} | null
+    product: {id: string; name: string}
+  }
+  existingPurchase: {
+    id: string
+    product: {id: string; name: string}
+  }
+  session: any
+  setPersonalPurchase: (props: any) => void
+}
```
jbranchaud, line 18: *not this PR*: we talked about this in another PR, but it would be nice to consolidate this specific set of types which appear in a bunch of places across apps into either the SDK or commerce-server

apps/total-typescript/src/components/team/self-redeem-button.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import {handleSelfRedeem} from 'path-to-purchase-react/handle-self-redeem'
+import {Button} from '@skillrecordings/react'
+import {Purchase} from '@skillrecordings/database'
+
+const SelfRedeemButton: React.FC<
```
jbranchaud, line 6: *not this PR*: is there a shared place for a simple, reused component like this to live and be shared across apps? Perhaps `packages/react/src/components`? Not sure what precedent we have for this, but this component seems like a good one to experiment with.

apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx
```diff
@@ -0,0 +1,227 @@
...
+          id: merchantProductId as string,
+        },
+        select: {
+          productId: true,
+        },
+      })
+      const productId = merchantProduct?.productId
+      const product = await getProduct({
+        where: {id: productId},
+      })
```
Comment (jbranchaud) [inline apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx:51]: *not this PR*: opportunity to pull some interactions out into the database SDK and the stripe SDK

apps/total-typescript/src/pages/invoices/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+import Link from 'next/link'
+import {getToken} from 'next-auth/jwt'
+import {Purchase} from '@skillrecordings/database'
+import {getCurrentAbility} from '@skillrecordings/ability'
+
+export const getServerSideProps: GetServerSideProps = async ({req}) => {
+  const sessionToken = await getToken({req})
+  const {getPurchasesForUser} = getSdk()
+
+  const ability = getCurrentAbility(sessionToken as any)
```
jbranchaud, line 18: *not this PR*: I'm imagining a thin wrapper around `getToken({ req })` that uses Zod to provide a type union where you either get `null` or you get a `token` which has a type of `{ id: string; sub: string; ... }` so that we can have better type support in the editor and two clear paths:
- if the token is `null`, redirect
- else, you have the values you need to reference the auth'd user

joelhooks: 👍👍

apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx
Comment (joelhooks) [reply apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx:51]: 👍👍

apps/total-typescript/src/components/team/self-redeem-button.tsx
joelhooks: my intent here was to suggest a specific package for this stuff and not cram it into an existing catch all package

having many packages is good imo

apps/total-typescript/src/components/team/index.tsx
joelhooks: absolutely

apps/total-typescript/src/components/team/index.tsx
joelhooks: i'm disliking the "database" package naming, but that is probably the spot for this kind of thing.

apps/total-typescript/src/components/team-invites.tsx
joelhooks: I'm leaving that for the team work that needs to happen

apps/total-typescript/src/components/team-invites.tsx
joelhooks: i'm gonna leave it for now and let it go in next phase

vojtaholik (approved):

PR #598, opened 2022-11-15
fix(pt): add w-square-diagonal to sandpack config file

simonswiss:


simonswiss:
Closing, this was not the right approach - app is currently using the CDN version of Tailwind

PR #603, opened 2022-11-16
feat(tt): add quick-and-dirty logout page

jbranchaud:
I needed a way to log out of an account while building and smoke testing
features. So, I added a `/logout` page with a button that uses
NextAuth's `signOut()` function.

We'll eventually need a more robust logout mechanism. But this will work
for now.

A noteable issue I ran into while implementing this (which I was unable
to debug) is that the button redirects to `localhost:3000` instead of
`localhost:3016`. I've explicitly passed in a full redirect URL, I've
ensure there aren't any env vars set that could be interfering. Beyond
that, I don't have enough expose to NextAuth to know why it would be
redirect to a completely different port.

Even though it redirects to the wrong host, it does successfully clean
up the session, so I can navigate back to `localhost:3016` and continue
with what I'm doing.

CleanShot 2022-11-16 at 10 15 14@2x

log out

### DIFF
--- a/apps/total-typescript/src/pages/logout.tsx
+++ b/apps/total-typescript/src/pages/logout.tsx
@@ -1,0 +1,57 @@
+import React from 'react'
+import {signOut} from 'next-auth/react'
+import {GetServerSideProps} from 'next'
+import {LogoutIcon} from '@heroicons/react/outline'
+import {isEmpty} from 'lodash'
+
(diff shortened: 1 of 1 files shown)


jbranchaud:
@joelhooks pointed out that one can navigate directly to `/api/auth/signout` for the built-in NextAuth sign out page. That presents a Sign Out button you can hit to clear the user session.

PR #604, opened 2022-11-16
dev(tt): remove VERCEL/VERCEL_ from .env.local

jbranchaud:
When we pull the Vercel env vars, it includes a bunch of Vercel-specific
env vars. This updates the `dev:setup` script to strip those out.

The `VERCEL="1"` in particular causes issues with NextAuth because it
makes that tool think it is running in the Vercel CI environment.

ver


apps/total-typescript/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "next dev -p 3016",
-    "dev:setup": "vercel env pull .env.local",
+    "dev:setup": "vercel env pull .env.local && sed -i.bak '/^VERCEL/d' .env.local",
```
joelhooks, line 7: I think this should migrate to a bash script, or node if we want it to work outside of *nix

joelhooks (approved):

PR #605, opened 2022-11-16
fix(db): use same ID for bulk coupon and purchase

jbranchaud:
This was an oversight during a refactor (https://github.com/skillrecordings/products/pull/590). Because the `id` on a `Coupon`
is auto-generated if not given, this was silenting failing by creating a
Coupon with one `id` and a Purchase with a different `bulkCouponId`.
This is a situation where it would be really nice to have foreign key
constraints.

danny dvito


packages/database/src/prisma-api.ts
```diff
@@ -401,6 +402,7 @@ export function getSdk(
         } else {
           coupon = ctx.prisma.coupon.create({
             data: {
+              id: bulkCouponId,
```
jbranchaud, line 405: Here is the fix, this `id` value needs to be set to what we determined above, otherwise it just gets an auto-generated value that doesn't match what we put in `Purchase.bulkCouponId`.

packages/database/src/prisma-api.ts
```diff
@@ -380,13 +380,14 @@ export function getSdk(
...
-      const isBulkPurchase = quantity > 1 || !!existingBulkCoupon
+      const isBulkPurchase = quantity > 1 || Boolean(existingBulkCoupon)
 
       let bulkCouponId = null
       let coupon = null
 
       if (isBulkPurchase) {
-        bulkCouponId = existingBulkCoupon ? existingBulkCoupon.id : v4()
+        bulkCouponId =
+          existingBulkCoupon !== null ? existingBulkCoupon.id : v4()
```
jbranchaud, line 390: No behavioral change here, I just wanted to clean this up by making these values coerce to booleans.

joelhooks (approved):

PR #609, opened 2022-11-21
create-skill-app env template tweaks

garrettdimon:
This is minor, but did some tidying of the .env templates in create-skill-app.

The file headings help make it easier to disambiguate files at a glance, and the rest primarily just syncs up the files so they follow similar formats/groupings.


joelhooks (approved):
🍄

PR #610, opened 2022-11-22
feat(tt): account dropdown in top navigation

vojtaholik:
screenshots

gif


apps/total-typescript/src/components/app/navigation.tsx
```diff
@@ -70,38 +68,175 @@ const DesktopNav = () => {
...
+                    />
+                    {canViewContent && isSellingLive && (
+                      <>
+                        <div className="border-t border-gray-900/50 px-3 pb-3 pt-5 font-mono text-xs font-semibold uppercase tracking-wide">
+                          Account
+                        </div>
+                        <ul>
+                          {!canViewTeam && (
+                            <MobileNavLink path="/team" label="Invite team" />
+                          )}
```
jbranchaud, line 136: This should be `canViewTeam` instead of `!canViewTeam`.

```suggestion
                          {canViewTeam && (
                            <MobileNavLink path="/team" label="Invite team" />
                          )}
```

apps/total-typescript/src/components/app/navigation.tsx
```diff
@@ -70,38 +68,175 @@ const DesktopNav = () => {
...
+                      path="/tips"
+                      label="Tips"
+                      icon={
+                        <FireIcon
+                          className="h-5 w-5 text-orange-400"
+                          aria-hidden="true"
+                        />
+                      }
+                    />
+                    {canViewContent && isSellingLive && (
```
jbranchaud, line 128: I don't know if this check will work for this app. I believe a user who has team ownership (ability to manage a team) may not necessarily be able to view content (they haven't redeemed a purchase for themselves). Plus, the ability checks are typically done on a per module/section/lesson basis. The navigation often won't have the context -- for example, if the user is on the home page.

Thinking about what an alternative could be 🤔

jbranchaud: What if we add an ability like `can('manage', 'Account')` for when we have a signed in user. If they can manage their account, then they can see the drop down which will allow them to click 'Sign Out'.

CleanShot 2022-11-22 at 08 55 56@2x

jbranchaud: I could also see `can('view', 'Account')` -- not sure which one makes more sense.

joelhooks: ability checks can be done anywhere at any time on either client or server, so there's no "typically" just where it was done initially to protect the content. Using `manage` should probably be avoided since it's a "yes" to all actions, in favor of explicit actions.

I'd probably show this is `user` exists, and then be fine grained about what actions are present beyond that and then have `login` if user is not present

jbranchaud: > ability checks can be done anywhere at any time on either client or server, so there's no "typically"

What I meant by that was, the `can('view', 'Content')` only makes sense in the context of a specific module/section/lesson. Since the navigation is requesting ability checks without any content specified, the `view content` check will never resolve true.

apps/total-typescript/package.json
```diff
@@ -20,7 +20,7 @@
     "@amplitude/analytics-browser": "^1.3.0",
     "@amplitude/identify": "^1.10.2",
     "@amplitude/node": "^1.10.2",
-    "@casl/ability": "^6.0.0",
+    "@casl/ability": "^6.3.2",
```
joelhooks, line 23: bumping this to latest

v6 has an alternative syntax for defining abilities so I updated that as well https://casl.js.org/v6/en/advanced/typescript

apps/total-typescript/src/ability/abilitiy.test.ts
```diff
@@ -1,12 +1,37 @@
...
+  expect(
+    ability.can(
+      'edit',
+      subject('User', {
+        id: 'abcdedf',
+        purchases: [],
+      }),
+    ),
+  ).toBe(true)
+})
```
joelhooks, line 27: this ability isn't used yet, but would be a way to enforce rules against specific subjects. It was breaking when I tried it so I thought this would make good documentation of how to perform this

we probably want to utilize this style for verifying access to specific content based on purchases in the future

apps/total-typescript/src/ability/ability.ts
```diff
@@ -14,19 +19,15 @@ export const UserSchema = z.object({
...
-  | 'Account'
-  | 'all'
-export type AppAbility = Ability<[Actions, Subjects]>
-export const AppAbility = Ability as AbilityClass<AppAbility>
+type Abilities =
+  | ['view', 'Content']
+  | ['invite', 'Team']
+  | ['edit', 'User' | User]
+  | ['view', 'Team']
+  | ['view', 'Invoice']
```
joelhooks, line 27: they consolidated these definitions

apps/total-typescript/src/ability/ability.ts
```diff
@@ -14,19 +19,15 @@ export const UserSchema = z.object({
...
+export type AppAbility = MongoAbility<Abilities>
```
joelhooks, line 28: `MongoAbility` is new and alters the syntax in general

apps/total-typescript/src/ability/ability.ts
```diff
@@ -89,12 +90,24 @@ const isFreelyVisible = ({
...
+  const {can, rules} = new AbilityBuilder<AppAbility>(createMongoAbility)
   const {user} = viewerAbilityInput
 
+  can('edit', 'User', {
+    id: user?.id,
+  })
+
+  if (hasChargesForPurchases(user?.purchases)) {
+    can('view', 'Invoice')
+  }
```
joelhooks, line 109: if they have any charges, they can view invoices!

apps/total-typescript/src/components/app/navigation.tsx
```diff
@@ -70,38 +66,173 @@ const DesktopNav = () => {
           <FireIcon className="h-5 w-5 text-orange-400" aria-hidden="true" />
         }
       />
+      {isSellingLive && (
+        <hr
+          className="mx-3 h-1/4 w-px border-transparent bg-gray-700"
+          aria-hidden="true"
+        />
+      )}
+      {status === 'authenticated' ? (
```
joelhooks, line 75: if they are `authenticated` the menu is present, otherwise if `unathenticated` show login

note that `loading` status is not accounted for and causes layout shift

jbranchaud (approved):

PR #612, opened 2022-11-22
feat(tt): display the # of seats used on Team page

jbranchaud:
This PR displays the number of seats used on the team page, so that the owner can have an idea of how many people on their team are redeeming seats. If you made the purchase a while ago or have added additional seats, you might not know the total number of seats and number of seats used. This exposes both values.

I also cleaned up logic and type guards using Zod.

CleanShot 2022-11-22 at 13 12 22@2x

take a seat


apps/total-typescript/src/team/index.tsx
```diff
@@ -23,28 +24,41 @@ const InviteTeam: React.FC<React.PropsWithChildren<InviteTeamProps>> = ({
...
           bulkCouponId &&
           'Send the invite link below to your colleagues to get started:'}
       </p>
-      {redemptionsLeft && bulkCouponId && (
+      {bulkCouponData.usedCount > 0 && (
+        <p className="pb-3 text-xs">
+          Your team has already redeemed {bulkCouponData.usedCount} of{' '}
+          {bulkCouponData.maxUses} seats.
+        </p>
+      )}
```
jbranchaud, line 60: If some seats have been used, display that information to the Team Owner.

apps/total-typescript/src/team/index.tsx
```diff
@@ -23,28 +24,41 @@ const InviteTeam: React.FC<React.PropsWithChildren<InviteTeamProps>> = ({
...
+    .transform(({maxUses, usedCount}) => {
+      return {
+        maxUses,
+        usedCount,
+        numberOfRedemptionsLeft: maxUses - usedCount,
+        redemptionsLeft: maxUses > usedCount,
+      }
+    })
+
+  const bulkCouponData = bulkCouponSchema.parse(purchase.bulkCoupon)
```
jbranchaud, line 38: I feel good about this refactoring -- it cleans up the type guard / presence checking by using a declarative Zod transformation that exposes the original two values plus the two computed values.

joelhooks: looks great, `redemptionsLeft` is phrased like a number not a boolean and could use a `has` prefix imo

joelhooks (approved):

apps/total-typescript/src/team/index.tsx
jbranchaud: good catch -- the phrasing felt off, I like the sound of `hasRedemptionsLeft`

PR #615, opened 2022-11-22
Refactor the data parsing in the Team Invite component

jbranchaud:
- refactor(tt): rename boolean to `hasRedemptionsLeft` (per this comment: https://github.com/skillrecordings/products/pull/612#discussion_r1029854482)
- refactor(tt): move bulk coupon ID into Zod schema
- feat(tt): better error if bulk coupon is missing

This also upgrades to the latest Zod which gives access to `z.NEVER`
which is the standard way of bailing early from a Zod `transform()`
function.

rew


apps/total-typescript/src/team/index.tsx
```diff
@@ -25,40 +25,55 @@ const InviteTeam: React.FC<React.PropsWithChildren<InviteTeamProps>> = ({
...
+    .nullable()
+    .transform((data, ctx) => {
+      if (data === null) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message:
+            'The purchase is either not a bulk purchase or was queried without its bulk coupon.',
+        })
+        return z.NEVER
+      }
```
jbranchaud, line 38: The `purchase.bulkCoupon` can potentially come in as `null`. This now accounts for that and surfaces a much better runtime error message if that is the case.

apps/total-typescript/src/team/index.tsx
```diff
@@ -25,40 +25,55 @@ const InviteTeam: React.FC<React.PropsWithChildren<InviteTeamProps>> = ({
...
+
+      const {id, maxUses, usedCount} = data
       return {
+        bulkCouponId: id,
         maxUses,
         usedCount,
         numberOfRedemptionsLeft: maxUses - usedCount,
-        redemptionsLeft: maxUses > usedCount,
+        hasRedemptionsLeft: maxUses > usedCount,
```
jbranchaud, line 46: update naming of boolean variable

apps/total-typescript/src/team/index.tsx
```diff
@@ -25,40 +25,55 @@ const InviteTeam: React.FC<React.PropsWithChildren<InviteTeamProps>> = ({
...
       }
     })
 
-  const bulkCouponData = bulkCouponSchema.parse(purchase.bulkCoupon)
+  const {
+    bulkCouponId,
```
jbranchaud, line 51: Moved `bulkCouponId` into the Zod schema since that value is coming from the `purchase.bulkCoupon` object like the rest of it.

apps/total-typescript/src/team/index.tsx
```diff
@@ -25,40 +25,55 @@ const InviteTeam: React.FC<React.PropsWithChildren<InviteTeamProps>> = ({
...
-  const bulkCouponId = purchase?.bulkCoupon?.id
 
   return (
     <>
       <p className="py-3">
-        You have{' '}
-        <strong>{bulkCouponData.numberOfRedemptionsLeft} seats left</strong>.{' '}
-        {bulkCouponData.redemptionsLeft &&
+        You have <strong>{numberOfRedemptionsLeft} seats left</strong>.{' '}
+        {hasRedemptionsLeft &&
```
jbranchaud, line 67: Extracted the individual values from the data object above the return so that they can be used directly and less verbosely here.

joelhooks (approved):

PR #616, opened 2022-11-22
feat(tt): fix styles and use Total TypeScript site title on invoice items

jbranchaud:
- styles(tt): replace greens with blues (cyans)
- copy(tt): use site title instead of TA name

## Before

CleanShot 2022-11-22 at 16 36 59@2x

## After

CleanShot 2022-11-22 at 16 39 01@2x

cyan


apps/total-typescript/src/pages/invoices/index.tsx
```diff
@@ -53,12 +53,13 @@ const Learn: React.FC<React.PropsWithChildren<{purchases: Purchase[]}>> = ({
...
                           aria-hidden
-                          className="w-6 text-green-500"
+                          className="w-6 text-cyan-500"
                         />
                       </div>
                       <div>
                         <h2 className="text-lg font-semibold leading-tight">
-                          Testing Accessibility ({purchase.product.name}){' '}
+                          {process.env.NEXT_PUBLIC_SITE_TITLE} (
+                          {purchase.product.name}){' '}
```
jbranchaud, line 62: I mostly came here to update the background color and related styles, but noticed that we should be using the public site title here, so included that as well.

vojtaholik (approved):
❤️

PR #621, opened 2022-11-23
refactor(ta): use existing SDK method for DB call

jbranchaud:
refract

### DIFF
--- a/apps/testingaccessibility/src/pages/thanks/purchase.tsx
+++ b/apps/testingaccessibility/src/pages/thanks/purchase.tsx
@@ -15,7 +15,7 @@ import {
   NEW_INDIVIDUAL_PURCHASE,
   INDIVIDUAL_TO_BULK_UPGRADE,
 } from '@skillrecordings/types'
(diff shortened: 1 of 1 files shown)


apps/testingaccessibility/src/pages/thanks/purchase.tsx
```diff
@@ -44,27 +44,7 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
...
-    },
-    include: {
-      bulkCoupon: {
-        include: {
-          bulkCouponPurchases: true,
-        },
-      },
-    },
-  })
+  const purchase = await getSdk().getPurchaseForStripeCharge(stripeChargeId)
```
joelhooks, line 47: I'd prefer to use consistent destructuring syntax everywhere instead of this chaining approach if possible

`const {getPurchaseForStripeCharge} = getSdk()`

even when there is "only one method being used" it introduces an alternative structure which is confusing to parse for me

joelhooks (approved):

PR #627, opened 2022-11-26
feat: add newsletter page to TypeScript Course

jsjoeio:
I thought it would be a good idea to add a /newsletter page so we can direct people here as needed.

cc @laurosilvacom 

image


laurosilvacom (approved):

laurosilvacom:
@jsjoeio can't get builds to pass :/ going to open my own PR with your changes

PR #632, opened 2022-11-30
feat(tt): display quantity on team invoice

vojtaholik:
screenshot

gif


apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx
```diff
@@ -190,31 +220,45 @@ const Invoice: React.FC<
...
-                      ? `${charge.currency.toUpperCase()} 0.00`
-                      : `${charge.currency.toUpperCase()} ${amount}`}
-                  </td>
-                </tr>
+                {bulkCoupon ? (
+                  <tr className="table-row">
+                    <td>{product.name}</td>
+                    <td>
+                      {charge.currency.toUpperCase()}{' '}
+                      {formatUsd(charge.amount / 100 / bulkCoupon.maxUses)}
```
Comment (vojtaholik) [inline apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx:228]: couldn't think of better way of getting a price per seat 🤔

joelhooks (approved):

apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx
Comment (jbranchaud) [reply apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx:228]: This will work for the first bulk purchase, but if they add more seats on to an existing team, the numbers will get weird since `maxUses` will equal `originalPurchaseQuantity + newPurchaseQuantity`.

You could get the exact quantity for this purchase from Stripe. The commerce server does this.

https://github.com/skillrecordings/products/blob/3b813606e4ca720e422139f53e381e5eca7911d6/packages/commerce-server/src/record-new-purchase.ts#L46

CleanShot 2022-11-30 at 09 38 30@2x

PR #633, opened 2022-11-30
refactor(tt): use existing `getPurchaseForStripeCharge` function

jbranchaud:
Some refactoring based on my comments in https://github.com/skillrecordings/products/pull/632

- feat: serialize objects recursively for Next
- refactor(tt): use existing SDK function
- refactor(tt): eliminate unnecessary query

refa

### DIFF
--- a/apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx
+++ b/apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx
@@ -37,23 +37,13 @@ export const getServerSideProps: GetServerSideProps = async ({
     if (merchantCharge && merchantCharge.identifier) {
       const charge = await stripe.charges.retrieve(merchantCharge.identifier)
 
-      const merchantProductId = merchantCharge.merchantProductId
-      const merchantProduct = await prisma.merchantProduct.findUnique({
-        where: {
-          id: merchantProductId as string,
-        },
-        select: {
-          productId: true,
-        },
-      })
       const purchase = await getPurchaseForStripeCharge(
         merchantCharge.identifier,
       )
       const bulkCoupon = purchase && purchase.bulkCoupon
 
-      const productId = merchantProduct?.productId
       const product = await getProduct({
-        where: {id: productId},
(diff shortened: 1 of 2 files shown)


apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx
```diff
@@ -37,38 +37,13 @@ export const getServerSideProps: GetServerSideProps = async ({
...
-        : null
+      const purchase = await getPurchaseForStripeCharge(
+        merchantCharge.identifier,
+      )
+      const bulkCoupon = purchase && purchase.bulkCoupon
 
-      const productId = merchantProduct?.productId
       const product = await getProduct({
-        where: {id: productId},
+        where: {id: purchase?.productId},
```
Comment (jbranchaud) [inline apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx:46]: No need to Prisma query for the `merchantProduct` above, we can grab the `productId` off the `purchase` that we already queried for.

packages/commerce-server/src/prisma-next-serializer.ts
```diff
@@ -8,6 +8,8 @@ export function convertToSerializeForNextResponse(result: any) {
       result[resultKey] = result[resultKey].toISOString()
     } else if (result[resultKey]?.constructor?.name === 'Decimal') {
       result[resultKey] = result[resultKey].toNumber()
+    } else if (result[resultKey] instanceof Object) {
+      result[resultKey] = convertToSerializeForNextResponse(result[resultKey])
     }
```
jbranchaud, line 13: This allows the serializer to work with "nested" prisma objects. Prisma queries that do an include of a related record are going to potentially have things like `datetime` values nested below that will also need to be serialized.

e.g.:

```
// purchase
{
  id: '123',
  productId: '345',
  createdAt: <DateTime Obj>,
  // ...
  bulkCoupon: {
    id: '678',
    createdAt: <DateTimeObj>, // <== this will error unless we serialize recursively
  },
}
```

joelhooks (approved):
great one!

PR #634, opened 2022-11-30
feat(tt): team owner must claim a seat to view content

vojtaholik:
cat gif

### DIFF
--- a/apps/total-typescript/src/ability/ability.ts
+++ b/apps/total-typescript/src/ability/ability.ts
@@ -8,6 +8,9 @@ import {Exercise} from '../lib/exercises'
 import {SanityDocument} from '@sanity/client'
 import z from 'zod'
 import {hasAvailableSeats, hasBulkPurchase} from '@skillrecordings/ability'
+import get from 'lodash/get'
+import isEmpty from 'lodash/isEmpty'
+import last from 'lodash/last'
 
 export const UserSchema = z.object({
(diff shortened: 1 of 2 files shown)


apps/total-typescript/src/ability/ability.ts
```diff
@@ -52,8 +55,19 @@ const canViewWorkshop = ({user, module, lesson}: ViewerAbilityInput) => {
   //  available to the user (see below)
   const userHasPurchases = Boolean(user && user.purchases.length > 0)
   const hasVideo = Boolean(lesson?.muxPlaybackId)
-
-  return contentIsWorkshop && userHasPurchases && hasVideo
+  const userHasBulkPurchase = Boolean(
+    userHasPurchases && !isEmpty(get(last(user?.purchases), 'bulkCoupon')),
```
vojtaholik, line 59: I don't feel very confident about getting the last item from purchases array tbh.

jbranchaud: what about checking all of the user's purchases to see if at least one of them has a `bulkCoupon`, e.g. `_.some(user?.purchases || [], (purchase) => !isEmpty(purchase.bulkCoupon))`

vojtaholik:
I realized that I got it all wrong — using `usedCount` from a coupon to determine whether they have claimed a license for themselves doesn’t really make sense!

PR #639, opened 2022-12-01
refactor(tt): experiment with better page types

jbranchaud:
Use Next.js type helpers to get better typing across the
`getServerSideProps` and rendered component boundaries. It is not
perfect and will get more out of `tRPC` as we continue to expand how we
are using that. Nevertheless, we have a lot of pages that us
`getServerSideProps` and this can help improve our typings for those
situations.

[getServerSideProps with TypeScript](https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props#getserversideprops-with-typescript)

props

### DIFF
--- a/apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx
+++ b/apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx
@@ -2,24 +2,36 @@ import * as React from 'react'
 import {DownloadIcon} from '@heroicons/react/outline'
 import {convertToSerializeForNextResponse} from '@skillrecordings/commerce-server'
 import {useLocalStorage} from 'react-use'
-import {GetServerSideProps} from 'next'
+import {GetServerSideProps, InferGetServerSidePropsType} from 'next'
 import {stripe} from '@skillrecordings/commerce-server'
 import {Coupon, getSdk} from '@skillrecordings/database'
 import {Stripe} from 'stripe'
 import fromUnixTime from 'date-fns/fromUnixTime'
 import Layout from 'components/app/layout'
 import format from 'date-fns/format'
-import {prisma} from '@skillrecordings/database'
+// combine this line with the same-package import above
+import {prisma, Product} from '@skillrecordings/database'
 import {getCurrentAbility} from '@skillrecordings/ability'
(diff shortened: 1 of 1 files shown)


apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx
```diff
@@ -2,24 +2,36 @@ import * as React from 'react'
...
 import {getToken} from 'next-auth/jwt'
+import {isArray} from 'lodash'
 
-export const getServerSideProps: GetServerSideProps = async ({
+type InvoiceProps = {
+  charge: Stripe.Charge
+  product: Product
+  merchantChargeId: string
+  bulkCoupon: Coupon | null
+}
```
Comment (jbranchaud) [inline apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx:23]: Define the types in one place at the top of the page, then use them in `getServerSideProps` with `GetServerSideProps<InvoiceProps>`. Then the page component can infer the props it is receiving by doing `InferGetServerSidePropsType<typeof getServerSideProps>`.

apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx
```diff
@@ -46,14 +58,26 @@ export const getServerSideProps: GetServerSideProps = async ({
...
+      }
+
       return {
-        props: {
-          charge,
-          product: convertToSerializeForNextResponse(product),
-          merchantChargeId,
-          bulkCoupon: convertToSerializeForNextResponse(bulkCoupon),
-        },
+        props,
```
Comment (jbranchaud) [inline apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx:80]: If I declare `props` like this (with its explicit type as `InvoiceProps`) and then return it, I get type warnings right here for the specific values that have type issues. If I assemble and return props inline without explicitly referencing the type, then the type errors show up on the function and are an indecipherable bundle.

joelhooks (commented):
I don't personally like this syntax and don't think we get a lot of benefit at the cost of general readability in the code. The props were already typed in the component and we don't gain much in the way of safety in the server rendering side.

what am I missing?

joelhooks:
https://trpc.io/docs/ssg#fetch-data-in-getstaticprops

here's this technique used in an example

PR #643, opened 2022-12-02
feat(tt,commerce-server) support cross-purchase tiered pricing

jbranchaud:
@garrettdimon, @Creeland, and I worked on adding support for tiered pricing that accounts for multiple bulk purchases. If you have already bought some seats and now you are buying more, enough to surpass a tiered pricing threshold, then the appropriate discount should be applied to the new purchase.

- feat(tt): add cross-purchase tiered pricing
- refactor(tt): pull seat count logic into function
- test(tt): test cross-purchase tiered discounts

tier

### DIFF
--- a/packages/commerce-server/src/format-prices-for-product.test.ts
+++ b/packages/commerce-server/src/format-prices-for-product.test.ts
@@ -113,6 +113,26 @@ test('product with quantity 5 calculatedPrice to have discount applied', async (
   expect(expectedPrice).toBe(product?.calculatedPrice)
 })
 
+test('multiple purchases meeting quantity threshold have bulk discount applied', async () => {
+  const userId = 'user-123'
+  mockCtx.prisma.merchantCoupon.findMany.mockResolvedValue([
+    getMockCoupon('bulk', 0.05),
+  ])
+  mockCtx.prisma.purchase.findMany.mockResolvedValue([
+    getMockExistingBulkPurchase(userId, DEFAULT_PRODUCT_ID, 3),
+  ])
+  const quantity = 2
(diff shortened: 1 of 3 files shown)


packages/commerce-server/src/format-prices-for-product.test.ts
```diff
@@ -113,6 +113,26 @@ test('product with quantity 5 calculatedPrice to have discount applied', async (
...
 })
 
+test('multiple purchases meeting quantity threshold have bulk discount applied', async () => {
+  const userId = 'user-123'
+  mockCtx.prisma.merchantCoupon.findMany.mockResolvedValue([
+    getMockCoupon('bulk', 0.05),
+  ])
+  mockCtx.prisma.purchase.findMany.mockResolvedValue([
+    getMockExistingBulkPurchase(userId, DEFAULT_PRODUCT_ID, 3),
+  ])
```
jbranchaud, line 123: the test is failing on the `purchase.findMany` call, it seems that it isn't getting properly mocked so it is instead trying to call the DB for which it doesn't have a valid connection string. I'm mocking this call the same way other calls are being mocked though 🤔 

CleanShot 2022-12-02 at 14 29 56@2x

jbranchaud: The issue was that `ctx` (which is mocked in the test) was not being passed to the instance of `getSdk()` that was being used by the code exercised in this test. Passing `ctx` down and replacing `getSdk()` with `getSdk({ ctx })` fixed it.

joelhooks (approved):
⭐️

PR #653, opened 2022-12-06
feat(tt): share pricing display logic across app

jbranchaud:
The Buy page has a component with some custom styling for display the
formatted price along with any discount being applied. This attempts to
extract that as a component that can be used in the 'Buy More Seats'
component as well.

WIP: while this is pretty close to what we need -- I think the logic / mechanics are all in place -- the styles don't work well in both places.

@vojtaholik any ideas for how to get these styles working in both places? Or is this where the component sharing breaks down for this example and the styles need to be separated?

CleanShot 2022-12-06 at 16 16 16@2x

seat


joelhooks:
Screen Shot 2022-12-06 at 14 40 20

I pushed this, but it could still use some more love probably, though it looks decent!

joelhooks (approved):

PR #662, opened 2022-12-07
feat(commerce): put qty and checkout session in DB

jbranchaud:
_Pitch for this work: https://github.com/skillrecordings/products/discussions/656_

Record `quantity` and `checkoutSessionId` when creating a Purchase.

This way each Purchase record has a cached value of the quantity
purchased and a reference to the Stripe checkout session via ID to be
able to quickly look up other data related to that Purchase's checkout
session in Stripe.

Before this can go out, the database changes need to be applied to the TA and TT databases in Planetscale. I'll post links to those deploy requests once they are up.

- TA (deployed): https://app.planetscale.com/skill-recordings/testing-accessibility/deploy-requests/15
- TT (deployed): https://app.planetscale.com/skill-recordings/total-typescript/deploy-requests/4

Expand and Contract details for applying these changes to planetscale as deploy requests are documented here: https://roamresearch.com/#/app/egghead/page/KwQNqwYHS

### Next Steps

- Consider scripting a backfill process to add `quantity` and `checkoutSessionId` to all existing Purchases. This should just need to happen in the TA database.

perch

### DIFF
--- a/packages/database/prisma/schema.prisma
+++ b/packages/database/prisma/schema.prisma
@@ -202,6 +202,8 @@ model Purchase {
   redeemedBulkCouponId String?
   productId            String
   merchantChargeId     String?
+  quantity             Int?
+  checkoutSessionId    String?
   upgradedFromId       String?         @unique
   upgradedFrom         Purchase?       @relation("UpgradedToPurchase", fields: [upgradedFromId], references: [id], onDelete: NoAction, onUpdate: NoAction)
   upgradedTo           Purchase?       @relation("UpgradedToPurchase")
--- a/packages/commerce-server/src/format-prices-for-product.test.ts
+++ b/packages/commerce-server/src/format-prices-for-product.test.ts
@@ -369,6 +369,7 @@ function getMockExistingBulkPurchase(
     couponId: null,
     redeemedBulkCouponId: null,
     merchantChargeId: 'ch-123',
+    checkoutSessionId: 'cs-123',
     upgradedFromId: null,
     status: 'Valid',
     bulkCouponId: 'coupon-123',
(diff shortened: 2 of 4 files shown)


joelhooks (commented):
if we want to keep the same pattern I think we need a MerchantPurchase analog that stores sessions with MerchantAccount and identifier

we are denormalized in the hopes that alternative payment providers could be leveraged, and while it makes access a layer deeper, prisma handles that well and it doesn't muddy the shape of the patterns

packages/database/prisma/schema.prisma
```diff
@@ -202,6 +202,8 @@ model Purchase {
   redeemedBulkCouponId String?
   productId            String
   merchantChargeId     String?
+  quantity             Int?
+  checkoutSessionId    String?
```
joelhooks, line 206: this should be `identifier` in new `MerchantPurchase` table to fit with existing modeling

I think quantity might belong there as well, unless we can make a strong case for leaving it at this layer.

joelhooks:
@joelhooks I added a `bulk` property to metadata because we are allowing a "bulk" purchase of 1. Since we are dealing with quantity, didn't know if that would affect this functionality

joelhooks:
is this still valid?

jbranchaud:
> is this still valid?

It's valid in that this is still a feature worth doing to have better invoices for bulk/team purchases.

However it is not valid as is, it needs to be reworked a little to put the `checkoutSessionId` on a `MerchantPurchase` table to better cohere to the existing schema design. https://github.com/skillrecordings/products/pull/662#discussion_r1043683711

jbranchaud:
Closing this PR, created a linear issue for this with updated details https://linear.app/skillrecordings/issue/TT-172/improve-teambulk-invoices-with-more-purchase-metadata

PR #663, opened 2022-12-08
feat(tt): adds stronger drm

joelhooks:
using a trpc media loading scheme to chuck out better DRM so that open internet can't scrape all the videos 

the layer after this is using signed media urls

https://www.mux.com/blog/securing-video-content-with-signed-urls

locked down


apps/total-typescript/src/ability/ability.ts
```diff
@@ -45,17 +46,15 @@ const canViewTutorial = ({user, subscriber, module}: ViewerAbilityInput) => {
...
-const canViewWorkshop = ({user, module, lesson}: ViewerAbilityInput) => {
+const canViewWorkshop = ({user, module}: ViewerAbilityInput) => {
   const contentIsWorkshop = module?.moduleType === 'workshop'
 
   const purchases = user?.purchases || []
   const userHasPurchaseWithAccess = Boolean(
     purchases.find((purchase) => purchase.bulkCouponId === null),
   )
 
-  const hasVideo = Boolean(lesson?.muxPlaybackId)
```
joelhooks, line 56: I was scratching my head at why "has video" was needed here but I don't think it matters. We'd want a different state entirely for that situation I reckon.

apps/total-typescript/src/components/exercise-overlay.tsx
```diff
@@ -114,9 +114,9 @@ const Actions = () => {
 }
 
 const ExerciseOverlay = () => {
-  const {lesson, module} = useMuxPlayer()
+  const {lesson, module, lessonMedia} = useMuxPlayer()
```
joelhooks, line 117: the introduction of `lessonMedia` seperates playback ids and stackblitz from the static data that is loaded at buildtime

apps/total-typescript/src/components/exercise-overlay.tsx
```diff
@@ -481,15 +481,13 @@ const BlockedOverlay: React.FC = () => {
 type LoadingOverlayProps = {}
 
 const LoadingOverlay: React.FC<LoadingOverlayProps> = () => {
-  const {
-    muxPlayerProps: {playbackId},
```
joelhooks, line 485: musPlayerProps can be `null` now

apps/total-typescript/src/hooks/use-mux-player.tsx
```diff
@@ -69,6 +70,17 @@ export const VideoProvider: React.FC<
...
+  const {data: lessonMedia, status: lessonMediaLoadingStatus} = trpc.useQuery([
+    'exercise.media',
+    {
+      moduleSlug: module.slug.current,
+      moduleType: module.moduleType,
+      lessonSlug: exerciseSlug,
+      sectionSlug: section?.slug,
+      isSolution: lesson._type === 'solution',
+    },
+  ])
```
joelhooks, line 82: similar to an abilities call

apps/total-typescript/src/server/routers/exercise.ts
```diff
@@ -2,17 +2,77 @@ import {createRouter} from 'server/createRouter'
...
+        isSolution,
+        muxPlaybackId,
+      })
+
+      if (lessonSlug && ability.can('view', 'Content')) {
+        const exerciseMedia: ExerciseMedia = await getExerciseMedia(lessonSlug)
+        return ExerciseMediaSchema.parse(exerciseMedia)
+      }
+    },
+  })
```
joelhooks, line 78: check for media ability server side and do not serve if they can't watch

the above is a little hamfisted with the data loading and ability checks, might have perf issues, but it'll cache in teh client and we might be able to initiate some sort of preload scheme since trpc does batching really well basically dropping an effect in the mix and doing something like this:

https://trpc.io/docs/v9/links#request-batching

apps/total-typescript/src/lib/exercises.ts
```diff
@@ -2,6 +2,23 @@ import {sanityClient} from '../utils/sanity-client'
...
+    z
+      .object({
+        slug: z.string(),
+        body: z.any().array().optional(),
+        stackblitz: z.nullable(z.string()).optional(),
+        muxPlaybackId: z.nullable(z.string()).optional(),
+      })
+      .optional(),
+  ),
+})
```
jbranchaud, line 20: Zod 🙌

apps/total-typescript/src/server/get-content-rules.ts
```diff
@@ -0,0 +1,58 @@
...
+import {defineRulesForPurchases, UserSchema} from '../ability/ability'
+
+export const ContentRulesSchema = z.object({
+  moduleSlug: z.string().optional(),
+  moduleType: z.string().optional(),
+  lessonSlug: z.string().optional(),
+  sectionSlug: z.string().optional(),
+  isSolution: z.boolean().optional(),
+  muxPlaybackId: z.nullable(z.string().optional()),
+})
```
jbranchaud, line 18: since all the properties are `optional`, another way to write this is with `partial()`:

```suggestion
export const ContentRulesSchema = z.object({
  moduleSlug: z.string(),
  moduleType: z.string(),
  lessonSlug: z.string(),
  sectionSlug: z.string(),
  isSolution: z.boolean(),
  muxPlaybackId: z.string().nullable(),
}).partial()
```

apps/total-typescript/src/server/routers/exercise.ts
```diff
@@ -2,17 +2,41 @@ import {createRouter} from 'server/createRouter'
...
+    async resolve({ctx, input}) {
+      const rules = await getContentRules(ctx.req, input)
+      const ability = createAppAbility(rules)
+
+      if (input.lessonSlug && ability.can('view', 'Content')) {
+        const exerciseMedia: ExerciseMedia = await getExerciseMedia(
+          input.lessonSlug,
+        )
+        return ExerciseMediaSchema.parse(exerciseMedia)
+      }
```
jbranchaud, line 40: this is a great solution to getting media URLs behind server-side ability checks 💯

apps/total-typescript/src/templates/exercise-template.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react'
 import ModuleLessonListHeader from 'components/module-lesson-list-header'
 import Navigation from 'components/app/navigation'
 import Layout from 'components/app/layout'
-import {VideoProvider} from 'hooks/use-mux-player'
+import {useMuxPlayer, VideoProvider} from 'hooks/use-mux-player'
```
jbranchaud, line 5: imported, but not used
```suggestion
import {VideoProvider} from 'hooks/use-mux-player'
```

jbranchaud (approved):
🚀

joelhooks:
see SKS-6

joelhooks:
https://linear.app/skillrecordings/issue/SKS-6/resolve-type-diff-in-exercise-across-pro-tailwind-and-total-typescript

PR #670, opened 2022-12-11
fix(tt) purchase cta disappears on home page

joelhooks:
fixes TT-3


joelhooks:
my working theory is that this commit https://github.com/skillrecordings/products/commit/9b7b78020428d1934c0e3a939f162dc947ddb398 removed the caching but because of The Edge™ there is still some caching which is why clearing cookies fixed it. It was routing based on `level`

PR #674, opened 2022-12-13
[TA] Add boss letter and update faqs

tayiorbeii:
* Adds "boss" letter
* Updates FAQ section

set it right


joelhooks (approved):
looks great @tayiorbeii thanks!

PR #679, opened 2022-12-14
feat(cs): consider all scenarios for Purchase Type

jbranchaud:
There were some gaps in the `determinePurchaseType` logic in the `commerce-server`. The value returned from this function is used on the post-purchase page to switch on what info and CTAs we show the user.

This logic has now been updated to account for the following things:

- a user can have multiple 'individual' purchases: standard individual, upgrade, redemption
- some apps, TT for instance, will eventually have multiple products, the logic should have affordances for that
- any uncovered scenario should fall through to `NEW_INDIVIDUAL_PURCHASE` so that the default to some messaging rather than no/sparse messaging

When this issue first came up, we thought it would be as simple as removing the _is there an individual purchase_ check and then falling through to `NEW_INDIVIDUAL_PURCHASE` for anything that didn't catch. However, that check was used all throughout the logic. It turned out there wasn't a quick fix. This was a little more involved, but sets us up for handling future scenarios while simplifying the logic of current scenarios.

scenario


packages/commerce-server/src/determine-purchase-type.ts
```diff
@@ -41,81 +133,79 @@ type DeterminePurchaseTypeOptions = {
...
+    // this checkout session
+    if (!userId || !checkoutSessionPurchase?.id)
+      throw new Error('Missing userId or purchase for checkout session')
+
+    const allUserPurchases = await getPurchasesForUser(userId)
+
+    const {type, data} = summarizePurchases(
+      checkoutSessionPurchase,
+      allUserPurchases,
+    )
```
jbranchaud, line 171: off load the logic of summarizing the purchases to this function, which helps delineate `BULK` and `INDIVIDUAL` as a union type. The union type makes the rest of this logic cleaner.

packages/commerce-server/src/determine-purchase-type.ts
```diff
@@ -41,81 +133,79 @@ type DeterminePurchaseTypeOptions = {
...
+    // there are more specific types of individual purchases:
+    // - standard individual
+    // - upgrade individual
+    // - redemption
+    //
+    // but we'll just treat all of these as NEW_INDIVIDUAL_PURCHASE for now
+    return NEW_INDIVIDUAL_PURCHASE
+  } catch (e) {
+    // Report this to Sentry
     return NEW_INDIVIDUAL_PURCHASE
```
jbranchaud, line 209: Everything falls through to `NEW_INDIVIDUAL_PURCHASE` if we didn't identify a specific bulk purchase scenario

packages/commerce-server/src/determine-purchase-type.ts
```diff
@@ -41,81 +133,79 @@ type DeterminePurchaseTypeOptions = {
...
+          // this is the first purchase, it is a new bulk coupon purchase
+          return NEW_BULK_COUPON
+        } else {
+          // they made an individual purchase before, but are now purchasing
+          // a bulk coupon, that's an upgrade to bulk
+          return INDIVIDUAL_TO_BULK_UPGRADE
+        }
+      } else {
+        // anything else at this point is a purchase of additional bulk seats
+        return EXISTING_BULK_COUPON
```
jbranchaud, line 194: Able to handle the 3 bulk scenarios based on the summarized purchase data from `summarizePurchases`.

packages/commerce-server/src/determine-purchase-type.ts
```diff
@@ -41,81 +133,79 @@ type DeterminePurchaseTypeOptions = {
...
+    // Find the purchase record associated with this Stripe Checkout Session
+    // (via the `stripeChargeId`).
+    const checkoutSessionPurchase = await getPurchaseForStripeCharge(
+      stripeChargeId,
+    )
+
+    // return early if we don't have a userId or a purchase corresponding to
+    // this checkout session
+    if (!userId || !checkoutSessionPurchase?.id)
+      throw new Error('Missing userId or purchase for checkout session')
```
jbranchaud, line 164: this used to be a `return null`, but I'm having it throw instead which gets caught and returns as `NEW_INDIVIDUAL_PURCHASE` -- this narrows the type of the function to just the 4 purchase types, no `null` values.

packages/commerce-server/src/determine-purchase-type.ts
```diff
@@ -41,81 +133,79 @@ type DeterminePurchaseTypeOptions = {
...
+    const {email, stripeChargeId} = purchaseInfo
+
+    const user = !!email && (await getUserByEmail(email))
+    const {id: userId} = user || {}
+
+    // Find the purchase record associated with this Stripe Checkout Session
+    // (via the `stripeChargeId`).
+    const checkoutSessionPurchase = await getPurchaseForStripeCharge(
+      stripeChargeId,
+    )
```
jbranchaud, line 159: nothing changed here, this all just got indented by the `try`.

packages/commerce-server/src/determine-purchase-type.ts
```diff
@@ -41,81 +133,79 @@ type DeterminePurchaseTypeOptions = {
 
 export async function determinePurchaseType(
   options: DeterminePurchaseTypeOptions,
-): Promise<PurchaseType | null> {
+): Promise<PurchaseType> {
```
jbranchaud, line 136: 🎉

packages/commerce-server/src/determine-purchase-type.ts
```diff
@@ -33,6 +33,98 @@ const match = (
...
+    otherIndividualPurchases: Purchase[]
+  }
+}
+
+const summarizePurchases = (
+  focalPurchase: PurchaseForStripeCharge,
+  allUserPurchases: Purchase[],
+):
+  | {type: 'BULK'; data: BulkPurchaseData}
+  | {type: 'INDIVIDUAL'; data: IndividualPurchaseData} => {
```
jbranchaud, line 66: the breakthrough for me in this implementation was separating the logic that figures out the important/relevant pieces of info about a user's purchases **FROM** the logic that determines and returns the purchase type

packages/commerce-server/src/determine-purchase-type.ts
```diff
@@ -33,6 +33,98 @@ const match = (
...
+  // summarize only purchases whose productId's match the focal
+  // purchase's productId
+  const focalProductId = focalPurchase.productId
+
+  const purchaseIsBulk = Boolean(focalPurchase.bulkCoupon?.id)
+
+  // Summarize the purchases for this product ID
+  const focalProductPurchases = allUserPurchases.filter(
+    (purchase) => purchase.productId === focalProductId,
+  )
```
jbranchaud, line 76: affordances for an app that may have multiple products that can be purchased

for now, we just focus on the product/`productId` that matches the current purchase's product and summarize based on that. We can easily extend this to summarize details of purchases for other products.

jbranchaud: need to rethink this part slightly -- this won't quite work with something like TA where there are 3 different products that correspond to different purchase tiers of the TA product.

Creeland:
Should we expand the tests in determine-purchase-type.test.ts to cover bulk purchases?

Creeland (approved):

PR #680, opened 2022-12-15
fix(tt): move purchase objects to new/updated user

joelhooks:
shell game


apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx
```diff
@@ -158,9 +161,9 @@ const Invoice: React.FC<
                   Invoice For
                 </h2>
                 <div>
-                  {charge.billing_details.name}
+                  {customer.name}
```
Comment (joelhooks) [inline apps/total-typescript/src/pages/invoices/[merchantChargeId].tsx:164]: billing details are at time of purchase so using customer here because it can update

packages/database/src/prisma-api.ts
```diff
@@ -484,6 +484,11 @@ export function getSdk(
         user = await ctx.prisma.user.create({
           data: {email, name},
         })
+      } else if (name && user.name !== name) {
+        user = await ctx.prisma.user.update({
+          where: {id: user.id},
+          data: {name},
+        })
```
joelhooks, line 491: if we pass a name in, use it for updates too

packages/database/src/prisma-api.ts
```diff
@@ -541,5 +546,40 @@ export function getSdk(
...
+            in: chargesToUpdate.map((c) => c.id),
+          },
+        },
+        data: {
+          userId: userId,
+        },
+      })
+
+      return await ctx.prisma.$transaction([chargeUpdates, purchaseUpdates])
+    },
```
joelhooks, line 583: main logic, would like to test but these would need to be integration?

packages/skill-api/src/core/services/process-stripe-webhook.ts
```diff
@@ -92,14 +96,12 @@ export async function processStripeWebhooks({
...
           const currentEmail = user.email
           const {email, name} = event.data.object
-          await prisma.user.update({
-            where: {
-              id: user.id,
-            },
-            data: {
-              email,
-              name,
-            },
```
joelhooks, line 102: the issue was that in Total TypeScript you can have an account using the target email already where in Testing accessibility you couldn't have an account without a purchase. This TRANSFERS the purchase instead of updating the user.

packages/database/src/prisma-api.ts
jbranchaud: yeah, I would love to have some integration (unit test integrated with database) tests for these kinds of things. The tests we typically write in Rails, for instance, strike a great balance IMO of acting mostly as unit-level tests while integrating the database. This allows you to meaningfully cover business logic without having to jump into e2e mode (which are higher-effort, slower running tests).

Not sure how we'd set up database-integrating tests in this stack, but it'd be worth a spike at some point.

packages/skill-api/src/core/services/process-stripe-webhook.ts
jbranchaud: The individual pieces of this look good. I'm getting hung up on understanding the scenario this is for. I'll try to spell out what I think the scenario is, though it is mostly questions checking for understanding.

This is to solve an issue with users and purchases in Total TypeScript, right? (as opposed to TA)

A person ends up with two separate `User` records? And we need to transfer any `Purchase` and `MerchantCharge` records from one to the other? When would that happen?

It seems like because of line 100, we are either looking up the one `User` record for this person's email or creating that initial `User` record. How is it that they already have purchases at this point and from what `User` record are we transferring them?

Or is it that there are existing `Purchase` and `MerchantCharge` records for which `userId` is `null`?

packages/database/src/prisma-api.ts
joelhooks: I'd make them a separate suite and just use the docker container probably.

This feels like a spot where it makes a lot of sense to invest in a solution that exercises the database for real, but I'd probably keep stripe portion mocked initially

packages/skill-api/src/core/services/process-stripe-webhook.ts
joelhooks: @jbranchaud with TT we allow you to login and create an account. This means that when I do the email change in stripe the previous logic failed because it can't update the purchasing user's email to the new email.

TA does NOT allow login without purchase, but this would work fine there as well.

Downside: progress is lost.

Instead of a mutation, we treat user's email addresses as immutable and instead transfer the purchase to a new user.

This will be similar to the self-serve "transfer purchase" functionality that we will be adding, and could probably be directly used for that.

vojtaholik (approved):

PR #682, opened 2022-12-15
feat(tt): add Buy More Seats to Buy page

jbranchaud:
This PR adds a link to the Buy page, if you've already purchased, that links you to the Buy More Seats page. This way we don't dead-end users on the Buy page when they want to upgrade to a team or get more seats for their team.

CleanShot 2022-12-15 at 12 29 13@2x

bu


joelhooks (commented):
I added analytics back

joelhooks (approved):

PR #683, opened 2022-12-15
feat: improve 'out of seats' state of Team page

jbranchaud:
This gets rid of the old 'buy more seats' button that conditionally
displays when there are no more seats. We can instead rely on the 'Buy
More Seats' widget below this that is always visible.

It also adds a disabled attribute and disabled styles to the various
team buttons, instead of hiding them, to make it clearer what is
available if you get more seats.

## Total TypeScript

https://user-images.githubusercontent.com/694063/207927412-83204784-8824-40d3-ae63-cc07ebffc711.mp4

## Testing Accessibility

https://user-images.githubusercontent.com/694063/207927470-4b4ee89a-0d68-4ff5-b0d9-4d2e11c3a496.mp4

out of seats

### DIFF
--- a/apps/total-typescript/src/team/index.tsx
+++ b/apps/total-typescript/src/team/index.tsx
@@ -69,12 +69,9 @@ const InviteTeam: React.FC<React.PropsWithChildren<InviteTeamProps>> = ({
         {hasRedemptionsLeft &&
           bulkCouponId &&
           'Send the invite link below to your colleagues to get started:'}
+        {usedCount > 0 &&
+          `Your team has already redeemed ${usedCount} of ${maxUses} seats.`}
       </p>
-      {usedCount > 0 && (
-        <p className="pb-3 text-xs">
-          Your team has already redeemed {usedCount} of {maxUses} seats.
-        </p>
-      )}
       {bulkCouponId && (
         <>
           <div className="w-full ">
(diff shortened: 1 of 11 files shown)


apps/total-typescript/src/team/index.tsx
```diff
@@ -94,20 +95,12 @@ const InviteTeam: React.FC<React.PropsWithChildren<InviteTeamProps>> = ({
...
       )}
-      {!hasRedemptionsLeft && (
-        <div className="mt-5 flex items-center justify-between border-t border-gray-100 pt-5">
-          <Link href="/#buy">
-            <a className="flex-shrink-0 rounded-md bg-cyan-500 px-4 py-2 font-semibold text-white transition hover:bg-cyan-600">
-              Buy more seats
-            </a>
-          </Link>
-        </div>
-      )}
```
jbranchaud, line 110: Same: https://github.com/skillrecordings/products/pull/683/files#r1049953467

packages/react/src/components/button/index.tsx
```diff
@@ -72,7 +72,11 @@ const Button: React.FC<React.PropsWithChildren<ButtonProps>> = ({
   const contentProps = {rightIcon, leftIcon, children}
 
   return (
-    <button data-sr-button {...rest} disabled={isLoading || rest.disabled}>
+    <button
+      data-sr-button
+      {...rest}
+      disabled={isLoading || isDisabled || rest.disabled}
+    >
```
jbranchaud, line 79: This shared react package button was not taking the `isDisabled` prop/option into account, this adds it as one of the `disabled` conditions.

vojtaholik: wasn't it getting passed as part of `rest` props? I can see `rest.disabled` in there.

vojtaholik (approved):

PR #693, opened 2022-12-20
feat(tt): custom hook for formatted price logic

jbranchaud:
I'm going to go a different direction on this --> tRPC query instead

but I thought it would be interesting to share what I came up with here. This `useFormattedPrice` hook could be used across TT and TA to replace 4 individual `useQuery` calls that are all effectively the same.

hooked


jbranchaud:
superseded by https://github.com/skillrecordings/products/pull/695

PR #695, opened 2022-12-21
feat(tt): move price formatting to tRPC query

jbranchaud:
This is the first step in a process to move this kind of query into the
skill-api package so that it can be shared by other apps. That process
will eventually lead to upgrading tRPC to v10 as well.

format

### DIFF
--- a/apps/total-typescript/src/path-to-purchase-react/pricing.tsx
+++ b/apps/total-typescript/src/path-to-purchase-react/pricing.tsx
@@ -26,6 +26,7 @@ import {track} from '../utils/analytics'
 import {useRouter} from 'next/router'
 import * as Switch from '@radix-ui/react-switch'
 import Link from 'next/link'
+import {trpc} from 'utils/trpc'
 
 function getFirstPPPCoupon(availableCoupons: any[] = []) {
   return find(availableCoupons, (coupon) => coupon.type === 'ppp') || false
@@ -82,29 +83,21 @@ export const Pricing: React.FC<React.PropsWithChildren<PricingProps>> = ({
   const {subscriber, loadingSubscriber} = useConvertkit()
   const router = useRouter()
 
-  const {data: formattedPrice, status} = useQuery<FormattedPrice>(
-    ['pricing', merchantCoupon, debouncedQuantity, productId, userId, couponId],
-    () =>
-      fetch('/api/skill/prices', {
-        method: 'POST',
-        body: JSON.stringify({
-          productId,
-          ...(merchantCoupon && {merchantCouponId: merchantCoupon.id}),
-          quantity,
-          purchases,
-          userId,
-          siteCouponId: couponId,
-        }),
-        headers: {
-          'Content-Type': 'application/json',
-        },
-      })
-        .then((res) => res.json())
-        .then((formattedPrice: FormattedPrice) => {
-          addPrice(formattedPrice, productId)
-          return formattedPrice
-        }),
-  )
+  const {data: formattedPrice, status} = trpc.useQuery([
+    'pricing.formatted',
+    {
+      productId,
+      userId,
+      quantity: debouncedQuantity,
+      couponId,
+      merchantCoupon,
+      purchases,
+    },
+  ])
+
+  if (formattedPrice) {
+    addPrice(formattedPrice, productId)
+  }
 
   const defaultCoupon = formattedPrice?.defaultCoupon
   const appliedMerchantCoupon = formattedPrice?.appliedMerchantCoupon
@@ -116,7 +109,7 @@ export const Pricing: React.FC<React.PropsWithChildren<PricingProps>> = ({
   // do not show the box if purchased
   // do not show the box if it's a downgrade
   const showPPPBox =
-    (pppCoupon || merchantCoupon?.type === 'ppp') &&
+    Boolean(pppCoupon || merchantCoupon?.type === 'ppp') &&
     !purchased &&
     !isDowngrade(formattedPrice)
 
--- a/apps/total-typescript/src/path-to-purchase-react/pricing.tsx
+++ b/apps/total-typescript/src/path-to-purchase-react/pricing.tsx
@@ -83,21 +83,24 @@ export const Pricing: React.FC<React.PropsWithChildren<PricingProps>> = ({
   const {subscriber, loadingSubscriber} = useConvertkit()
   const router = useRouter()
 
-  const {data: formattedPrice, status} = trpc.useQuery([
-    'pricing.formatted',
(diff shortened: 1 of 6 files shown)


apps/total-typescript/src/team/buy-more-seats.tsx
```diff
@@ -30,33 +29,15 @@ const BuyMoreSeats = (props: BuyMoreSeatsProps) => {
...
-    .parse(formattedPrice)
+  const {data: formattedPrice, status} = trpc.useQuery([
+    'pricing.formatted',
+    {
+      productId,
+      userId,
+      quantity: debouncedQuantity,
+      purchases: [],
+    },
+  ])
```
jbranchaud, line 40: This `trpc` call on the 'Buy More Seats' page works great.

apps/total-typescript/src/path-to-purchase-react/pricing.tsx
```diff
@@ -82,29 +83,21 @@ export const Pricing: React.FC<React.PropsWithChildren<PricingProps>> = ({
...
-          'Content-Type': 'application/json',
-        },
-      })
-        .then((res) => res.json())
-        .then((formattedPrice: FormattedPrice) => {
-          addPrice(formattedPrice, productId)
-          return formattedPrice
-        }),
-  )
+  const {data: formattedPrice, status} = trpc.useQuery([
```
jbranchaud, line 86: Since this is the primary change to the 'Buy' component, I have to assume this is related to the issue I am seeing.

The issue being: when I try to toggle from 'Individual' to 'Team', it doesn't actually toggle and the page locks up.

When I change `quantity: debouncedQuantity` to just `quantity`, it allows me to toggle it once, but then when I click the toggle again, the page locks up.

This `trpc` endpoint is working well for the other page I put it on, so I'm having a hard time understanding what it is getting caught on here.

I guess the main different between this query and the one on the other page are the options going into it:

```javascript
    {
      productId,
      userId,
      quantity: debouncedQuantity,
      purchases: [],
    },
```

I'm running out of ideas for what the issue could be or how to debug it. Any insights @joelhooks ?

jbranchaud: Here is a video of the behavior.

https://user-images.githubusercontent.com/694063/208967930-288bae13-77e3-4337-a922-04bb0b35d6f3.mp4

A detail I just noticed this time is that a `/api/trpc/pricing.formatted` web request isn't showing up in the network log for that click, so it may not even be getting to call that in the first place.

apps/total-typescript/src/server/routers/pricing.ts
```diff
@@ -0,0 +1,143 @@
...
+      percentageDiscount: defaultCoupon.percentageDiscount.toString(),
+    }
+
+    return {activeMerchantCoupon, defaultCoupon: minimalDefaultCoupon}
+  }
+}
+
+export const pricing = createRouter().query('formatted', {
+  input: PricingFormattedInputSchema,
+  async resolve({ctx, input}): Promise<FormattedPrice> {
```
jbranchaud, line 98: Here is where the `pricing.formatted` trpc query lives -- I love this shift. This is a better spot for all of this logic and the built-in Zod input parsing is great.

jbranchaud: I'm imaging this query will get promoted to a skill-api query that can be shared with TA and other apps, but wanted to get the first pass working within just TT.

packages/commerce-server/src/get-active-merchant-coupon.ts
```diff
@@ -6,9 +6,9 @@ export async function getActiveMerchantCoupon({
   siteCouponId,
   code,
 }: {
-  productId: string
-  siteCouponId: string
-  code: string
+  productId: string | undefined
+  siteCouponId: string | undefined
+  code: string | undefined
```
jbranchaud, line 11: The downstream uses of these values are each `string | undefined`, so I loosened this type signature so that I could use it more cleanly in the trpc query.

apps/total-typescript/src/path-to-purchase-react/pricing.tsx
jbranchaud: Turns out this was due to some weird rendering logic with the `addPrice` hook. Moving `addPrice` into an `onSuccess` handler on the tRPC query did the trick.

image

apps/total-typescript/src/server/routers/pricing.ts
```diff
@@ -0,0 +1,143 @@
...
+    productId,
+  )
+
+  return find(purchases, (purchase) => {
+    const upgradeProductIds = availableUpgrades.map(
+      (upgrade) => upgrade.upgradableFrom.id,
+    )
+    return upgradeProductIds.includes(purchase.productId)
+  })?.id
+}
```
joelhooks, line 53: these methods likely belong in the `commerce-server` package

that could be in a follow on PR for sure, but that's where I see them

apps/total-typescript/src/server/routers/pricing.ts
joelhooks: Thinking about the `skill-api` I think it makes sense for webhooks/callbacks to some extent, but wonder how far we could take trpc and reduce the number of options we have for how we do things.

packages/commerce-server/src/get-active-merchant-coupon.ts
joelhooks: making them optional is effectively the same result but less verbose with `?` so wondering if this was required/preferred?

joelhooks (approved):
🎉

apps/total-typescript/src/server/routers/pricing.ts
jbranchaud: the `purchases` option, for instance, is something that we could look up on the fly in a tRPC query instead of passing it all over. that would have the added benefit of ensuring the value is _fresh_.

packages/commerce-server/src/get-active-merchant-coupon.ts
jbranchaud: They are the same as far as I can tell and `code?: string`, etc. would have been better. Just typed it out as such probably right after looking at a type overlay that displayed as `string | undefined`.

apps/total-typescript/src/server/routers/pricing.ts
joelhooks: absolutely agree

jbranchaud: 🎉  https://github.com/skillrecordings/products/pull/698

PR #701, opened 2022-12-26
use a less restrictive base type for loading lesson metadata

joelhooks:
we are also using trpc to load various media async

metadata


apps/total-typescript/src/ability/ability.ts
```diff
@@ -32,7 +32,7 @@ export const createAppAbility = createMongoAbility as CreateAbility<AppAbility>
 type ViewerAbilityInput = {
   user?: User
   subscriber?: any
-  lesson?: Exercise
+  lesson?: LessonResource
```
joelhooks, line 35: `LessonResource` is considerably less opinionated at the type level with no concerns regarding the underlying **media** (stackblitz, mux, etc)

apps/total-typescript/src/components/exercise-overlay.tsx
```diff
@@ -114,10 +113,14 @@ const Actions = () => {
 
 const ExerciseOverlay = () => {
   const {lesson, module} = useMuxPlayer()
+  const router = useRouter()
+  const {data: stackblitz, status} = trpc.stackblitz.byExerciseSlug.useQuery({
+    slug: router.query.exercise as string,
+    type: lesson._type,
+  })
```
joelhooks, line 120: the router gives us out exercise slug and then we can load stackblitz by type

this component isn't universally portable (stackblitz). We send `_type` so we can switch on `solution`

apps/total-typescript/src/components/lesson-list.tsx
```diff
@@ -239,16 +239,18 @@ const SolutionLink = ({
...
   section?: SanityDocument
-  exercise: Exercise
+  exercise: LessonResource
   path: string
 }) => {
   const router = useRouter()
-  const solution = exercise.solution
+  const {data: solution, status} = trpc.solutions.getSolution.useQuery({
+    exerciseSlug: router.query.exercise as string,
+  })
```
joelhooks, line 248: a `LessonResource` doesn't specifically have a `solution` (an Exercise is a superset) so we load them if we can async

apps/total-typescript/src/hooks/use-mux-player.tsx
```diff
@@ -46,6 +44,22 @@ type VideoProviderProps = {
...
+) => {
+  const router = useRouter()
+  const {data: nextExercise} = trpc.lessons.getNextLesson.useQuery({
+    type: lesson._type,
+    slug: router.query.exercise as string,
+    module: module.slug.current,
+    section: section?.slug.current,
+  })
+  return nextExercise
 }
```
joelhooks, line 63: this hook lets us load the next lesson async since we don't have the `solution` on the LessonResource

apps/total-typescript/src/hooks/use-mux-player.tsx
```diff
@@ -59,15 +73,12 @@ export const VideoProvider: React.FC<
   onEnded = async () => {},
   section,
   exerciseSlug,
+  videoThumbId,
```
joelhooks, line 76: this is the mux playback id as well and we just pass it around instead of rooting into the "exercise" for it

apps/total-typescript/src/hooks/use-mux-player.tsx
```diff
@@ -142,35 +151,56 @@ export const VideoProvider: React.FC<
...
+      playbackRate: currentPlaybackRate,
+    })
+  }, [currentPlaybackRate, setPlayerPrefs])
+
+  const setDisplayOverlayCallback = React.useCallback(
+    (value: boolean) => {
+      setDisplayOverlay(value)
+    },
+    [setDisplayOverlay],
+  )
```
joelhooks, line 195: these were all callbacks so should be defined as such

apps/total-typescript/src/lib/exercises.ts
```diff
@@ -1,36 +1,28 @@
...
+          _key: z.string(),
+          stackblitz: z.nullable(z.string()).optional(),
+          muxPlaybackId: z.nullable(z.string()).optional(),
+          transcript: z.nullable(z.any().array()).optional(),
+        })
+        .merge(BaseLessonResourceSchema)
+        .optional(),
+    ),
+  })
+  .merge(BaseLessonResourceSchema)
```
joelhooks, line 25: an exercise and solution set uses the lesson resource schema

apps/total-typescript/src/pages/tutorials/[module]/[exercise]/index.tsx
```diff
@@ -12,7 +12,12 @@ export const getStaticProps: GetStaticProps = async (context) => {
   const exercise = await getExercise(exerciseSlug)
 
   return {
-    props: {exercise, module},
+    props: {
+      exercise,
+      module,
+      videoThumbId: exercise.muxPlaybackId,
```
Comment (joelhooks) [inline apps/total-typescript/src/pages/tutorials/[module]/[exercise]/index.tsx:18]: exercises still exist and we can load them at build time and utilize the full set before treating it as a LessonResource at the page/component level

apps/total-typescript/src/pages/tutorials/[module]/[exercise]/index.tsx
```diff
@@ -37,8 +42,20 @@ export const getStaticPaths: GetStaticPaths = async (context) => {
...
+  module,
+  videoThumbId,
+  transcript,
+}) => {
+  return (
+    <ExerciseTemplate
+      exercise={exercise}
+      module={module}
+      videoThumbId={videoThumbId}
+      transcript={transcript}
```
Comment (joelhooks) [inline apps/total-typescript/src/pages/tutorials/[module]/[exercise]/index.tsx:56]: LessonResource also doesn't have a transcript but we want those at build time for seo etc

apps/total-typescript/src/schemas/base-lesson-resource.ts
```diff
@@ -0,0 +1,10 @@
+import z from 'zod'
+
+export const BaseLessonResourceSchema = z.object({
+  _type: z.string(),
+  _updatedAt: z.string().optional(),
+  title: z.string(),
+  slug: z.string(),
+  description: z.nullable(z.string()).optional(),
+  body: z.nullable(z.any().array().optional()),
+})
```
joelhooks, line 10: basic resources look like this

apps/total-typescript/src/server/routers/lessons.ts
```diff
@@ -0,0 +1,39 @@
...
+
+      const exerciseForSolution = exercises.find((resource: SanityDocument) => {
+        return resource.solution?._key === lesson._key
+      })
+
+      const current = find(exercises, {_id: exerciseForSolution._id})
+      const nextExerciseIndex = indexOf(exercises, current) + 1
+      return exercises[nextExerciseIndex]
+    }),
+})
```
joelhooks, line 39: next lesson logic is now serverside/async

apps/total-typescript/src/server/routers/solutions.ts
```diff
@@ -0,0 +1,16 @@
...
+  getSolution: publicProcedure
+    .input(
+      z.object({
+        exerciseSlug: z.string(),
+      }),
+    )
+    .query(async ({ctx, input}) => {
+      const lesson = await getExercise(input.exerciseSlug)
+      return lesson.solution
+    }),
```
joelhooks, line 15: we can grab an exercise solution async as needed

apps/total-typescript/src/server/routers/stackblitz-resource.ts
```diff
@@ -0,0 +1,23 @@
...
+    )
+    .query(async ({ctx, input}) => {
+      const lesson = await getExercise(input.slug)
+
+      const stackblitz =
+        input.type === 'solution'
+          ? lesson.solution?.stackblitz
+          : lesson.stackblitz
+
+      return stackblitz
```
joelhooks, line 21: grab the stackblitz async

apps/total-typescript/src/hooks/use-mux-player.tsx
joelhooks: i think i'm going to switch to the videoResourceId and create an api proxy endpoint for the video thumbnail and then this "mux" hook will be something like useVideoResource 🤔

apps/total-typescript/src/hooks/use-mux-player.tsx
```diff
@@ -46,6 +44,22 @@ type VideoProviderProps = {
...
+
+const useNextLesson = (
+  lesson: VideoResource,
+  module: SanityDocument,
+  section?: SanityDocument,
+) => {
+  const router = useRouter()
+  const {data: nextExercise} = trpc.lessons.getNextLesson.useQuery({
+    type: lesson._type,
+    slug: router.query.exercise as string,
```
joelhooks, line 58: the route contains data and we can use that which is super useful for determining if we want the exercise or the solution as the current lesson

this will also work if we use /ide as well to present the stackblitz or other interactive step

apps/total-typescript/src/pages/tutorials/[module]/[exercise]/index.tsx
Comment (joelhooks) [reply apps/total-typescript/src/pages/tutorials/[module]/[exercise]/index.tsx:56]: the `exercise` is "cast down" in the `exercise-template` and "loses" data when it become a  `LessonResource` which is why we explicitly separate the transcript as a "media resource"

apps/total-typescript/src/ability/ability.ts
joelhooks: we have an existing LessonResource

https://github.com/skillrecordings/products/blob/main/packages/types/src/index.ts#L50

It's very strictly tied to the egghead oriented sellable data model

that entire package could probably be deprecated

PR #702, opened 2022-12-27
load video resources instead of attaching playback id to lesson

joelhooks:
playback

### DIFF
--- a/apps/total-typescript/src/templates/workshop-template.tsx
+++ b/apps/total-typescript/src/templates/workshop-template.tsx
@@ -150,9 +150,7 @@ const Header: React.FC<{workshop: SanityDocument}> = ({workshop}) => {
 const WorkshopSectionNavigator: React.FC<{workshop: SanityDocument}> = ({
   workshop,
 }) => {
-  const {slug, sections, _type} = workshop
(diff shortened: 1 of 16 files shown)


apps/total-typescript/src/templates/workshop-template.tsx
```diff
@@ -201,68 +198,90 @@ const WorkshopSectionNavigator: React.FC<{workshop: SanityDocument}> = ({
   )
 }
 
-const WorkshopSectionExerciseNavigator: React.FC<{
+const LessonListItem = ({
```
joelhooks, line 201: the list item let's us load data as needed

apps/total-typescript/src/video/use-video-resource.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const context = {
+    videoResource,
+    loadingVideoResource: status === 'loading',
+  }
+  return (
+    <VideoResourceContext.Provider value={context}>
+      {children}
+    </VideoResourceContext.Provider>
+  )
+}
```
joelhooks, line 35: simple context to give us access to the loaded video resource for the page/template/section

PR #705, opened 2022-12-28
moves all workshop video core into single namespace

joelhooks:
Alright, this moves all the core stuff into a `video` folder (name isn't final just what i started with. suggestions welcome). The remaining issue is how we use trpc if some of this moves into a package. 

Is it all or nothing?

The alternative would be to **swap out trpc for react query and skill-api.**

This PR is kind of wacky, but makes it relatively easier to think about versus having everything spread around the app. 

there are only for trpc calls in this folder which has me leaning towards skill-api implementation

suspect this will be a factor for pricing related stuff as well?

core


joelhooks:
I'm leaning towards `skill-api` instead of trpc...

apps/total-typescript/src/video/exercise-description.tsx
```diff
@@ -54,7 +56,7 @@ export const ExerciseDescription: React.FC<{exercise: LessonResource}> = ({
               <a>Total TypeScript Core Volume</a>
             </Link>{' '}
             and can be unlocked immediately after purchase. Already purchased?{' '}
-            <Link href="/login">
+            <Link href="/apps/total-typescript/src/pages/login">
```
joelhooks, line 59:
```suggestion
            <Link href="/login">
```

joelhooks, line 59:
```suggestion
            <Link href="/login">
```

joelhooks, line 59:
```suggestion
            <Link href="/login">
```

PR #706, opened 2022-12-28
use react-query instead of trpc

joelhooks:
Having a look at what using the `skill-api` would look like instead of local trpc because trpc wasn't obvious in the monorepo shared package setup.

query


apps/total-typescript/src/pages/api/skill/modules/verify-access.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const VerifyModuleAccessArgsSchema = z.object({
+  moduleSlug: z.string().optional(),
+  moduleType: z.string().optional(),
+  lessonSlug: z.string().optional(),
+  sectionSlug: z.string().optional(),
+  isSolution: z.boolean().default(false).optional(),
+})
+
+const verifyModuleAccess = async (
```
joelhooks, line 19: same story here in terms of portability, but we've settled on the module/section/lesson structure and we could think about what that looks like in a portable sense more generically since we don't need the full details of a module to understand if the learner has access or not

in our current scheme for Total TypeScript we are giving folks access to the first lesson, so there is logic for that, but we might be able to be more explicit about that marking specific lessons as free to avoid the need for so many arguments and we strictly check if a learner has access to the module itself

with that in mind we can use the module id, convertkit subscriber, and the logged in user to determine current access levels like `can('view', 'Module', module)` and the rules contain a list of modules the learner has access to based on their combined `subscriber` + `user` state in the databases

joelhooks (commented):
the skill-api approach feels pretty good to me

apps/total-typescript/src/pages/api/skill/modules/verify-access.ts
joelhooks: https://linear.app/skillrecordings/issue/TT-70/verify-access-at-the-module-level-instead-of-lesson-level

packages/skill-lesson/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@skillrecordings/skill-lesson",
```
joelhooks, line 2: this moves all of the "logic bits" into `skill-lesson` package leaving the UI components behind

packages/skill-lesson/hooks/use-mux-player.tsx
```diff
@@ -71,14 +70,24 @@ export const VideoProvider: React.FC<
...
+          moduleType: module.moduleType,
+          lessonSlug: exerciseSlug,
+          sectionSlug: section?.slug,
+          isSolution: lesson._type === 'solution',
+        }),
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }).then((response) => response.json())
+    },
```
joelhooks, line 89: todo: base ability on module access instead of this full stack

the ability needs all this info because we allow the first lesson of a module/section to be viewed freely, but that requires a ton of information.

packages/skill-lesson/hooks/use-next-lesson.ts
```diff
@@ -0,0 +1,32 @@
...
+          slug: router.query.exercise as string,
+          module: module.slug.current,
+          section: section?.slug,
+        }),
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }).then((r) => r.json())
+    },
+  )
```
joelhooks, line 30: the api routes are local to total typescript for now but can migrate to `skill-api`

also having those routes independent might give us the ability to perform different logic per product if we wanted/needed to

joelhooks (commented):
https://github.com/trpc/trpc/tree/main/examples/soa

PR #707, opened 2022-12-29
sets up the skill-lesson package and replaces use-convertkit

joelhooks:
first step towards having shared logic for the video lesson aspects of both pro tailwind and total typescript 🎉

tailwind


joelhooks (commented):
had to upgrade to trpc 10 after react-query update

PR #711, opened 2022-12-30
use a shared trpc server context between TT and PT

joelhooks:
sharing


apps/protailwind/src/server/routers/_app.ts
```diff
@@ -1,21 +1,15 @@
...
-  progress: progressRouter,
-  convertkit: convertkitRouter,
-  resources: lessonResourcesRouter,
-})
+export const appRouter = mergeRouters(
+  router({
+    resources: lessonResourcesRouter,
+  }),
+  skillLessonRouter,
+)
```
joelhooks, line 13: `mergeRouters` gives you a flat router from two sources

apps/total-typescript/package.json
```diff
@@ -105,7 +105,7 @@
     "react-use-clipboard": "^1.0.7",
     "refractor": "^4.5.0",
     "speakingurl": "^14.0.1",
-    "superjson": "^1.9.1",
+    "superjson": "^1.11.0",
```
joelhooks, line 108: version mismatches cause all sorts of issues in monorepo 😭

packages/skill-lesson/server/trpc.ts
```diff
@@ -8,9 +8,9 @@
  * @see https://trpc.io/docs/v10/procedures
  */
 
-import {Context} from './context'
 import {initTRPC} from '@trpc/server'
 import superjson from 'superjson'
+import {Context} from './context'
```
joelhooks, line 13: probably makes sense to move the context and this file to a dedicated `skill-trpc` package?

apps/total-typescript/package.json
KATT: There's a package called manypkg that can help

joelhooks: ya it's great! it's installed in this repo, just sometimes forget/neglect and get bitten (again) ☠️

packages/skill-lesson/hooks/use-video-resource.tsx
```diff
@@ -24,15 +21,8 @@ export const VideoResourceProvider: React.FC<VideoResourceProviderProps> = ({
...
-    ['videoResource', videoResourceId],
-    async () => {
-      const resource = await fetch(
-        `/api/skill/video-resources/${videoResourceId}`,
-      ).then((response) => response.json())
-      return VideoResourceSchema.parse(resource)
-    },
-  )
+  const {data: videoResource, status} =
+    trpcSkillLessons.videoResource.byId.useQuery({id: videoResourceId})
```
joelhooks, line 25: using trpc instead of local API

packages/skill-lesson/utils/trpc-skill-lessons.ts
```diff
@@ -0,0 +1,34 @@
...
+       * @link https://react-query.tanstack.com/reference/QueryClient
+       */
+      // queryClientConfig: { defaultOptions: { queries: { staleTime: 60 } } },
+    }
+  },
+  ssr: false,
+})
+
+export type RouterInput = inferRouterInputs<SkillLessonRouter>
+export type RouterOutput = inferRouterOutputs<SkillLessonRouter>
```
joelhooks, line 34: this is another instance of the router, so technically we are calling this factory twice:

1. in the app using `AppRouter`
2. here in the `skill-lessons` package

This appears to work fine in practice. @KATT would be curious what you think 😅

apps/total-typescript/src/video/lesson-list.tsx
```diff
@@ -246,18 +243,13 @@ const SolutionLink = ({
   path: string
 }) => {
   const router = useRouter()
-  const {data: solution} = useQuery(['solution', exercise.slug], async () => {
-    const fullExercise = await fetch(
-      `/api/skill/exercises/${exercise.slug}`,
-    ).then((res) => res.json())
-    return SolutionSchema.parse(fullExercise.solution)
-  })
```
joelhooks, line 254: was migrating this to trpc and noticed it was only called here and didn't think it was needed based on the logic below so I removed it

packages/skill-lesson/lib/lesson-resource.ts
```diff
@@ -0,0 +1,73 @@
...
+    solution: z.nullable(
+      z
+        .object({
+          _key: z.string(),
+        })
+        .merge(BaseLessonResourceSchema)
+        .optional(),
+    ),
+  })
+  .merge(BaseLessonResourceSchema)
```
joelhooks, line 19: I don't love all the variants of this type of object and the lesson/lesson resource/exercise stuff.

joelhooks: technically this is an extremely stripped down generic exercise shape, works for now but definitely something to flag and thing about

packages/skill-lesson/lib/modules.ts
```diff
@@ -0,0 +1,76 @@
...
+const modulesQuery = groq`*[_type == "module" && state == 'published'] | order(_createdAt desc) {
+  _id,
+  _type,
+  title,
+  slug,
+  "image": image.asset->url,
+  _updatedAt,
+  _createdAt,
+  description,
+  "exercises": resources[@->._type in ['exercise', 'explainer']]->{
```
joelhooks, line 13: this would be better named 'lessons' I think

packages/skill-lesson/server/routers/modules.ts
```diff
@@ -0,0 +1,48 @@
...
+      return defineRulesForPurchases({
+        ...(token && {user: UserSchema.parse(token)}),
+        ...(convertkitSubscriber && {
+          subscriber: convertkitSubscriber,
+        }),
+        module,
+        lesson,
+        section,
+        isSolution: input.isSolution,
+      }) as any[]
```
joelhooks, line 46: copped out with an `any` because it complained about the CASL types. It returns an array of rules, which should be fine here.

PR #717, opened 2023-01-02
feat(tt): exit fullscreen on video ended

vojtaholik:
this prevents content from being non-interactive when finishing a video in fullscreen mode by proactively exiting it. 

likely not ideal, but I couldn't figure out how to prevent default muxplayer's behavior and target its container as fullscreen wrapper to make it so video overlays could persist with fullscreen mode turned on. I tried listening for `onfullscreenchange` event on muxplayer's ref in combination with `e.preventDefault()` but that didn't work (weird).

gif


joelhooks:
did we check with mux support? we pay them!

joelhooks (approved):

PR #718, opened 2023-01-02
fix(tt): record bulk and non-bulk coupon relation

jbranchaud:
When either a bulk coupon (single-seat) or a non-bulk coupon is
redeemed, we need to tie the $0 purchase record to that coupon.

coupe

### DIFF
--- a/packages/skill-api/src/core/services/redeem-golden-ticket.ts
+++ b/packages/skill-api/src/core/services/redeem-golden-ticket.ts
@@ -39,7 +39,7 @@ export async function redeemGoldenTicket({
       req,
       options: {nextAuthOptions},
     } = params
-    const {findOrCreateUser, getCoupon} = getSdk()
+    const {findOrCreateUser, getCouponWithBulkPurchases} = getSdk()
 
     const {email: baseEmail, couponId, sendEmail = true} = req.body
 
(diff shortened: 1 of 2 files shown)


packages/skill-api/src/core/services/redeem-golden-ticket.ts
```diff
@@ -48,14 +48,16 @@ export async function redeemGoldenTicket({
...
+    const coupon = await getCouponWithBulkPurchases(couponId)
 
     const couponValidation = validateCoupon(coupon)
 
     if (coupon && couponValidation.isRedeemable) {
-      const bulkCouponRedemption = coupon.maxUses > 1
+      // if the Coupon is the Bulk Coupon of a Bulk Purchase,
+      // then a bulk coupon is being redeemed
+      const bulkCouponRedemption = Boolean(
+        coupon?.bulkCouponPurchases[0].bulkCouponId,
```
jbranchaud, line 59: This probably needs to be more like:

```
coupon?.bulkCouponPurchases?.[0]?.bulkCouponId
```

packages/skill-api/src/core/services/redeem-golden-ticket.ts
```diff
@@ -99,6 +101,8 @@ export async function redeemGoldenTicket({
         data: {
           userId: user.id,
           redeemedBulkCouponId: bulkCouponRedemption ? coupon.id : null,
+          // TODO: rename this to `couponUsedId` for non-bulk redemption papertrail
+          couponId: bulkCouponRedemption ? null : coupon.id,
```
jbranchaud, line 105: This way, either the `redeemedBulkCouponId` or the `couponId` get set.

packages/skill-api/src/core/services/redeem-golden-ticket.ts
joelhooks: `first` from lodash is generally my goto and it handles all that under the hood

joelhooks (approved):

PR #721, opened 2023-01-03
fix(tt): bring back subscribe form on tutorials

vojtaholik:
gif


packages/skill-lesson/server/routers/modules.ts
```diff
@@ -24,7 +24,7 @@ export const modulesRouter = router({
     )
     .query(async ({ctx, input}) => {
       const token = await getToken({req: ctx.req})
-      const convertkitSubscriber = getSubscriberFromCookie(ctx.req)
+      const convertkitSubscriber = await getSubscriberFromCookie(ctx.req)
```
vojtaholik, line 27: this is the culprit.

PR #722, opened 2023-01-03
feat(tt): process email signup w/o page refresh

vojtaholik:
https://user-images.githubusercontent.com/25487857/210367822-bcdbcd4e-5c53-443d-8290-2dfbd32ca995.mp4

gif


packages/skill-api/src/core/services/convertkit/subscribe-to-convertkit.ts
```diff
@@ -36,17 +36,16 @@ export async function subscribeToConvertkit({
...
-      )
-      await setConvertkitSubscriberFields(subscriberWithFields, fields)
+      await setConvertkitSubscriberFields(fullSubscriber, fields)
     }
 
     return {
       status: 200,
       body: subscriber,
-      cookies: getConvertkitSubscriberCookie(subscriber),
+      cookies: getConvertkitSubscriberCookie(fullSubscriber),
```
vojtaholik, line 48: we weren't storing a full subscriber right away. one could think about it as making it act as if they had confirmed their subscription. seems fine?

packages/skill-lesson/hooks/use-mux-player.tsx
```diff
@@ -63,14 +64,18 @@ export const VideoProvider: React.FC<
...
-      moduleType: module.moduleType,
-      lessonSlug: exerciseSlug,
-      sectionSlug: section?.slug,
-      isSolution: lesson._type === 'solution',
-    })
+  const {
+    data: abilityRules,
+    status: abilityRulesStatus,
+    refetch: refetchAbilityRules,
+    isRefetching,
```
vojtaholik, line 71: with full subscriber stored in a cookie we get to call `refetchAbilityRules` which reflects new abilities without having to refresh the page.

packages/skill-lesson/hooks/use-mux-player.tsx
```diff
@@ -31,6 +31,7 @@ type VideoContextType = {
   canShowVideo: boolean
   loadingUserStatus: boolean
   ability: AppAbility
+  refetchAbilityRules: () => {}
```
vojtaholik, line 34: how does one type this thing? seemed pretty complex so I avoided it but I might be missing something.

joelhooks: one way this would "just work" is to add the subscriber to the input of the existing trpc call so then when it mutates so do abilities?

joelhooks (approved):

PR #724, opened 2023-01-03
feat(tt): 'Buy More Seats' is aware of productId

jbranchaud:
If the URL query params include a `productId` we look up the product for the 'Buy More Seats' component that way. Otherwise we fallback to the user's bulk or individual purchase for a product ID.

Based on looking up the product by `productId`, we are able to get and display the product's name.

CleanShot 2023-01-03 at 16 24 53@2x

aware

### DIFF
--- a/apps/total-typescript/src/pages/team/buy-more-seats.tsx
+++ b/apps/total-typescript/src/pages/team/buy-more-seats.tsx
@@ -1,38 +1,50 @@
 import React from 'react'
 import Layout from 'components/app/layout'
-import {convertToSerializeForNextResponse} from '@skillrecordings/commerce-server'
 import {GetServerSideProps} from 'next'
 import BuyMoreSeats from 'team/buy-more-seats'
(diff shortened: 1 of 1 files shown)


apps/total-typescript/src/pages/team/buy-more-seats.tsx
```diff
@@ -66,11 +74,14 @@ const BuyMoreSeatsPage: React.FC<
       className="bg-noise text-gray-900"
     >
       <main className="mx-auto flex w-full max-w-xl flex-grow flex-col items-center justify-center gap-3 p-5 py-16">
+        <h2 className="mx-auto max-w-lg py-5 font-text text-3xl font-bold text-white lg:text-4xl">
+          {product.name}
+        </h2>
```
jbranchaud, line 79: Display the product name right about the 'Buy More Seats' component.

joelhooks (approved):

PR #741, opened 2023-01-09
feat(pt): basic sanity structure for commerce

vojtaholik:
Respecting following structure: **bundle > module > section > lesson**

Modules can have a product resource attached to them instead of product being a wrapper around module. If a module has a product associated with it, it can be purchased.
module

Products have a `productId` connecting it with a product in our database. 
product object

Bundle is a type of a product that can have multiple modules. It has a `productId`.
bundle

gif


apps/protailwind/studio/schemas/documents/module.js
```diff
@@ -76,6 +76,19 @@ export default {
...
+      description: 'Sellable products associated with this module',
+      type: 'array',
+      of: [
+        {
+          title: 'Exercise, Section, or Explainer',
+          type: 'reference',
+          to: [{title: 'Product', type: 'product'}],
+        },
+      ],
+    },
```
joelhooks, line 91: I don't think that we need to have a reference link like this and an individual module is likely just an association with a `productId` - We already have `Product` in the Prisma Schema, so imo this should generally be a reference to that.

Another way to look at this would be to have the Product as a `resource` object (similar to `solution` in Exercise) that can be filtered for.

What's the advantage to having a top level `Product` document? I think the module has most of the metadata (image, title, etc and features is probably not a fit here)?

vojtaholik: I like the resource object approach and updated my PR with it. 👍

PR #743, opened 2023-01-09
enhance(trpc): use a top level trpc folder for trpc stuff

joelhooks:
moves stuff out of /server and /utils and renames the client/config to `trpc.client.ts`

consolidate


jbranchaud (approved):
🚀  love having this all in the same top-level dir

PR #748, opened 2023-01-11
some lessons aren't exercises so show progress

joelhooks:
make progress

### DIFF
--- a/apps/protailwind/src/templates/workshop-template.tsx
+++ b/apps/protailwind/src/templates/workshop-template.tsx
@@ -350,8 +350,9 @@ const LessonListItem = ({
   const {data: userProgress} = trpc.progress.get.useQuery()
 
   const isExerciseCompleted =
-    isArray(userProgress) &&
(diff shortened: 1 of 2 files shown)


apps/protailwind/src/templates/workshop-template.tsx
```diff
@@ -350,8 +350,9 @@ const LessonListItem = ({
   const {data: userProgress} = trpc.progress.get.useQuery()
 
   const isExerciseCompleted =
-    isArray(userProgress) &&
-    find(userProgress, ({lessonSlug}) => lessonSlug === solution?.slug)
+    isArray(userProgress) && lessonResource._type === 'exercise'
+      ? find(userProgress, ({lessonSlug}) => lessonSlug === solution?.slug)
+      : find(userProgress, ({lessonSlug}) => lessonSlug === lessonResource.slug)
```
joelhooks, line 355: this is silly, but works

fundamentally I think we should be recording the EXERCISE (lesson) slug as progress/completion event and NOT the solution. that would require a migration of existing progress I think?

PR #752, opened 2023-01-14
bootstrap inngest in monorepo

joelhooks:
inngest is an open source workflow engine (coupled with a cloud service) that allows for async event driven "step function" workflows

https://www.inngest.com/docs/quick-start

- handling survey responses
- webhooks for content/media
- e-commerce webhooks
- email automation

ingest


apps/total-typescript/src/pages/api/inngest.ts
```diff
@@ -0,0 +1,6 @@
+import {serve} from 'inngest/next'
+import {convertkitSurveyAnswered} from '@skillrecordings/inngest'
+
+export default serve(process.env.NEXT_PUBLIC_SITE_TITLE, [
+  convertkitSurveyAnswered,
+])
```
joelhooks, line 6: inngest operates as an api endpoint within your app, so as events occur, they trigger configured functions. in this case only the `convertkitSurveyAnswered` function is present

packages/inngest/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@skillrecordings/inngest",
```
joelhooks, line 2: so we can use inngest across packages/apps it is a package

functions could also exist locally in apps, or even exported from other packages.

packages/inngest/src/convertkit.inngest.ts
```diff
@@ -0,0 +1,10 @@
+import {inngest} from './inngest.server'
+
+export const convertkitSurveyAnswered = inngest.createStepFunction(
+  'ConvertKit Survey Answered',
+  'convertkit/survey-answered',
+  ({event, tools}) => {
+    tools.sleep('1 second')
+    return {event, body: 'now you can do stuff!'}
+  },
```
joelhooks, line 9: this function doesn't actually do anything.

packages/skill-lesson/lib/convertkit.ts
```diff
@@ -86,9 +88,18 @@ export const answerSurvey = async ({
...
 
-    return await response.json()
+    await inngest.send({
+      name: 'convertkit/survey-answered',
+      data: {
+        subscriber,
+        question,
+        answer,
+      },
+    })
```
joelhooks, line 100: sending an event that will be handled by the configured function

PR #754, opened 2023-01-16
chore(deps): use manypkg to upgrade framer-motion

jbranchaud:
The Sanity v3 Studio that gets generated by the create-skill-app package is not able to run a development server. When you start it up, it immediately fails because of a bad import statement in a mismatched dependency.

I tracked this mismatched dependency down. Because of having several different versions of framer-motion throughout the various packages and apps, we get into a situation where motionone/dom 10.11.0 is being mixed with motionon/animation 10.14.0. The difference in those minor versions is an import getting moved around.

CleanShot 2023-01-16 at 12 20 25@2x

To fix this, we can run manypkg to update framer-motion. E.g.:

```
$ pnpm manypkg update framer-motion
```

This depends on `manypkg` having been added as a script in the top-level `package.json`.

many

### DIFF
--- a/apps/badass/package.json
+++ b/apps/badass/package.json
@@ -59,7 +59,7 @@
     "dotenv-flow": "^3.2.0",
     "focus-visible": "^5.2.0",
     "formik": "2.2.9",
-    "framer-motion": "^7.6.1",
+    "framer-motion": "^8.4.3",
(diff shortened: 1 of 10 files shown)


vojtaholik (approved):
⭐ thank you!!

PR #757, opened 2023-01-18
fix: fix next lesson logic for explainers

joelhooks:
explainers

### DIFF
--- a/packages/skill-lesson/trpc/routers/lessons.ts
+++ b/packages/skill-lesson/trpc/routers/lessons.ts
@@ -31,22 +31,27 @@ export const lessonsRouter = router({
 
       const lessons = section ? section.lessons : module.lessons
 
-      const exerciseForSolution = lessons.find((resource: SanityDocument) => {
-        return resource.solution?._key === lesson.solution?._key
-      })
+      let currentLesson
 
-      // TODO: this is likely overloading the lesson type just because it
-      //   exists and is different that the exercise type. We should probably
-      //   create a new type instead of `lesson` since `lesson` describes ALL
-      //   lessons, not just the ones that are not exercises.
-      const current =
-        input.type === 'lesson'
-          ? find(lessons, {slug: input.slug})
-          : find(lessons, {_id: exerciseForSolution._id})
-      const nextExerciseIndex = indexOf(lessons, current) + 1
+      switch (input.type) {
+        case 'exercise':
+          //exercises have solutions, that's what makes them exercises
+          //so we match the solution key to identify the exercise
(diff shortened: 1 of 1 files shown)


packages/skill-lesson/trpc/routers/lessons.ts
```diff
@@ -31,22 +31,27 @@ export const lessonsRouter = router({
...
-      const exerciseForSolution = lessons.find((resource: SanityDocument) => {
-        return resource.solution?._key === lesson.solution?._key
-      })
-
-      // TODO: this is likely overloading the lesson type just because it
-      //   exists and is different that the exercise type. We should probably
-      //   create a new type instead of `lesson` since `lesson` describes ALL
-      //   lessons, not just the ones that are not exercises.
-      const current =
-        input.type === 'lesson'
```
joelhooks, line 43: we were looking for a special type here with `lesson` but that wasn't catching `explainer`

packages/skill-lesson/trpc/routers/lessons.ts
```diff
@@ -31,22 +31,27 @@ export const lessonsRouter = router({
...
+        case 'exercise':
+          //exercises have solutions, that's what makes them exercises
+          //so we match the solution key to identify the exercise
+          const exerciseForSolution = lessons.find(
+            (resource: SanityDocument) => {
+              return resource.solution?._key === lesson.solution?._key
+            },
+          )
+          currentLesson = find(lessons, {_id: exerciseForSolution._id})
+          break
```
joelhooks, line 46: an `exercise` is special because it has a `solution` and we need the `exercise` for a `solution` when that is the case. other types shouldn't require this so instead of looking for `lesson` and `explainer` we check for `exercise` and do the special business

this whole shit needs tests

packages/skill-lesson/trpc/routers/lessons.ts
```diff
@@ -31,22 +31,27 @@ export const lessonsRouter = router({
...
-        ? LessonResourceSchema.parse(lessons[nextExerciseIndex])
+      let currentLesson
+
+      switch (input.type) {
+        case 'exercise':
+          //exercises have solutions, that's what makes them exercises
+          //so we match the solution key to identify the exercise
+          const exerciseForSolution = lessons.find(
+            (resource: SanityDocument) => {
+              return resource.solution?._key === lesson.solution?._key
```
joelhooks, line 42: fun fact! when both of these values are `undefined` this returns a match 🎉 (that's what was going on here, so two explainers in a row was making it sad)

PR #758, opened 2023-01-19
feat(pt): fixing seo errors to improve health score for the product

nicollguarnizo:
This PR attempts to fix different SEO errors we have in Pro Tailwind. 

1. Description problems (too short) 
2. 3XX redirect in sitemap because we were using the wrong URL 
3. Solving a 500 page error 

fire


apps/protailwind/.env.development
```diff
@@ -8,7 +8,7 @@ NEXT_PUBLIC_PARTNER_FIRST_NAME=Simon
 NEXT_PUBLIC_PARTNER_LAST_NAME=Vrachliotis
 NEXT_PUBLIC_PARTNER_TWITTER=simonswiss
 NEXT_PUBLIC_SEO_KEYWORDS="tailwind, css, frontend"
-NEXT_PUBLIC_PRODUCT_DESCRIPTION="Increase development velocity and craft sustainable systems for your team"
+NEXT_PUBLIC_PRODUCT_DESCRIPTION="Increase development velocity and craft sustainable systems for your team by digging into the techniques that will level up your Tailwind skills."
```
nicollguarnizo, line 11: the description was too short

apps/protailwind/public/sitemap.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
         <sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
-        <sitemap><loc>https://protailwind.com/sitemap-0.xml</loc></sitemap>
-<sitemap><loc>https://protailwind.com/server-sitemap.xml</loc></sitemap>
+        <sitemap><loc>https://www.protailwind.com/sitemap-0.xml</loc></sitemap>
```
nicollguarnizo, line 3: we are having `3XX redirect in sitemap` errors since it goes to a permanent redirect, adding the `www` fixes the error

apps/protailwind/src/video/lesson-list.tsx
```diff
@@ -140,12 +138,14 @@ export const LessonList: React.FC<{
                         Exercise
                       </Link>
                     </li>
-                    <SolutionLink
-                      module={module}
-                      exercise={exercise}
-                      section={section}
-                      path={path}
-                    />
+                    {exercise.solution != null && (
```
nicollguarnizo, line 141: I'm having a TS problem, and I'm probably thinking too hard about this, but I was trying to filter if the `exercise` has a `solution`. However, since the type comes from `import {type LessonResource} from '@skillrecordings/skill-lesson/schemas/lesson-resource'` and it doesn't seem to be able to access `solution` even tho its part of the object. 

I went down the road and added `solution` to the schema from `skill-lesson`, but this breaks builds for other apps, such as total typescript. 

Any ideas  how can access `solution` without breaking other apps?

packages/skill-lesson/schemas/base-lesson-resource.ts
```diff
@@ -7,4 +7,20 @@ export const BaseLessonResourceSchema = z.object({
...
+        transcript: z.nullable(z.any().array()).optional(),
+        github: z
+          .object({
+            url: z.string(),
+          })
+          .optional()
+          .nullable(),
+      }),
+    )
+    .optional(),
```
joelhooks, line 25: this can't happen. the entire purpose of this schema is to prevent this sort of application specific type.

apps/protailwind/src/video/lesson-list.tsx
joelhooks: **all exercises have solutions** or they should be a different type. struggling with an application fix for what is a data problem.

joelhooks: an exercise without a solution is an explainer!

joelhooks (approved):

joelhooks (approved):

PR #760, opened 2023-01-19
feat(tt): sort user progress by latest

vojtaholik:
makes it so userProgress returns a sorted array with first item being latest one updated. should make it easier to setup "continue watching" action.

gif


vojtaholik:
could this be done inside our sdk? https://github.com/skillrecordings/products/blob/main/packages/database/src/prisma-api.ts#L278-L285 

that's where I was looking at first but couldn't get `orderBy` to lookup by `progress.updatedAt`

packages/skill-lesson/trpc/routers/progress.ts
```diff
@@ -59,7 +60,12 @@ export const progressRouter = router({
         const lessonProgress = await getLessonProgressForUser(
           token.id as string,
         )
-        return lessonProgress
+        const sortedLessonProgressByLatest = sortBy(
+          lessonProgress,
+          ({updatedAt}) => updatedAt,
+        ).reverse()
```
joelhooks, line 66: instead of reversing a lodash sort, I'd recommend sorting them at the query level since this is likely how we will always want them: https://www.prisma.io/docs/concepts/components/prisma-client/filtering-and-sorting#sorting

https://github.com/skillrecordings/products/blob/330a47ddfa7fd51195f560e18e16fd9eb34f525c/packages/database/src/prisma-api.ts#L278-L286

PR #765, opened 2023-01-23
fix(pt): fixing images broken from crawling pages

nicollguarnizo:
The report from `ahrefs` showed images broken, this PR calls the MUX Api for the thumbnails instead of using `getBaseUrl`, which doesn't optimize the images and throws a 400

fix tape


apps/protailwind/src/lib/tips.ts
```diff
@@ -23,6 +23,7 @@ export const TipSchema = z.object({
     .optional()
     .nullable(),
   videoResourceId: z.nullable(z.string()).optional(),
+  muxPlaybackId: z.nullable(z.string()).optional(),
```
joelhooks, line 26: this is actually intentionally excluded! exposing it like this exposes the videos for public scraping

https://github.com/skillrecordings/products/blob/93f289a6ae58bbb315da06f5d49f4dd7b251a3fc/apps/total-typescript/src/pages/api/video-thumb.ts

https://github.com/skillrecordings/products/blob/5211e80c1ba210aabe2e30ab10cfdf3fa3b0bba0/apps/total-typescript/src/templates/tip-template.tsx#L57-L60

the question is why isn't this happening with other sites.

apps/protailwind/src/video/exercise-overlay.tsx
```diff
@@ -438,16 +437,6 @@ const BlockedOverlay = () => {
...
-          {videoResourceId && (
-            <Image
-              src={thumbnail}
-              layout="fill"
-              alt=""
-              aria-hidden="true"
-              objectFit="cover"
-              className="opacity-50 blur-sm brightness-50 contrast-125"
-            />
-          )}
```
joelhooks, line 450: why is this being removed?

joelhooks (changes requested):

apps/protailwind/src/video/exercise-overlay.tsx
vojtaholik: it's being used within the blocked overlay
image

apps/protailwind/src/lib/tips.ts
nicollguarnizo: I made all the changes based on how TT is set up.

https://github.com/skillrecordings/products/blob/93f289a6ae58bbb315da06f5d49f4dd7b251a3fc/apps/total-typescript/src/lib/tips.ts#L15

 If I understand correctly, the templates called the thumbnail using the `getBaseUrl`, which automatically crawls `_next` folder for the images, since the thumbnails come from MUX API, it keeps throwing the error. 

https://github.com/skillrecordings/products/blob/5211e80c1ba210aabe2e30ab10cfdf3fa3b0bba0/apps/total-typescript/src/templates/tip-template.tsx#L97-L99

There are no errors in the network for the pages for any image, but according to the `ahrefs` report, every page that has a video and loads a thumbnail, it's broken.

apps/protailwind/src/video/exercise-overlay.tsx
nicollguarnizo: I will add it back! didn't realize that part

apps/protailwind/src/lib/tips.ts
vojtaholik: the url can be replaced with `getBaseUrl()` version. it just won't work locally and I haven't looked into why yet. should work fine in prod.

nicollguarnizo: it does work fine in prod, it loads the image perfectly. The problem is when crawling the pages.

joelhooks: it's not crawling the _next folder, my guess is that Next generated image urls are being cached between CI and prod which is why the urls in _next have a dot vercel url

getBaseUrl() is a simple string replacement to switch between a deployed URL and localhost 

this is likely not a code problem

nicollguarnizo: Ok, will close this PR and keep searching for why this is happening.

PR #766, opened 2023-01-24
using lesson id and not solution slugs for progress

joelhooks:
slugs

### DIFF
--- a/apps/total-typescript/src/hooks/use-progress.ts
+++ b/apps/total-typescript/src/hooks/use-progress.ts
@@ -1,90 +1,60 @@
 import React from 'react'
-import {
-  first,
-  filter,
-  find,
-  flatMap,
-  get,
-  indexOf,
-  isEmpty,
-  map,
-  isArray,
-} from 'lodash'
+import {first, filter, find, flatMap, get, indexOf, isEmpty} from 'lodash'
 import {getNextSection} from '@skillrecordings/skill-lesson/utils/get-next-section'
 import {Lesson} from '@skillrecordings/skill-lesson/schemas/lesson'
 import {LessonProgress} from '@skillrecordings/database'
-import {Exercise} from '@skillrecordings/skill-lesson/schemas/exercise'
 import {Section} from '@skillrecordings/skill-lesson/schemas/section'
 import {Module} from '@skillrecordings/skill-lesson/schemas/module'
 import {trpcSkillLessons} from '@skillrecordings/skill-lesson/utils/trpc-skill-lessons'
 
-export function extractExercisesAndResource(sections?: Section[] | null) {
-  const exercises = flatMap<Section, Exercise>(
+export function getAllLessons(sections?: Section[] | null) {
+  const lessons = flatMap<Section, Lesson>(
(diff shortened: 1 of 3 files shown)


apps/total-typescript/src/hooks/use-progress.ts
```diff
@@ -1,90 +1,60 @@
...
 import {Lesson} from '@skillrecordings/skill-lesson/schemas/lesson'
 import {LessonProgress} from '@skillrecordings/database'
-import {Exercise} from '@skillrecordings/skill-lesson/schemas/exercise'
 import {Section} from '@skillrecordings/skill-lesson/schemas/section'
 import {Module} from '@skillrecordings/skill-lesson/schemas/module'
 import {trpcSkillLessons} from '@skillrecordings/skill-lesson/utils/trpc-skill-lessons'
 
-export function extractExercisesAndResource(sections?: Section[] | null) {
-  const exercises = flatMap<Section, Exercise>(
+export function getAllLessons(sections?: Section[] | null) {
```
joelhooks, line 10: I tore this hook down a lot and converted any mention of "exercise" to "lesson"

core philosophy here: it's ALL lessons

shouldn't need to determine if it's a solution, explainer, whatever

this means that we need to determine progress based on lesson IDs instead of storing solution slugs, that was broken approach, added loads of complexity

apps/total-typescript/src/hooks/use-progress.ts
```diff
@@ -99,206 +69,65 @@ export const getNextSectionForProgress = ({
...
-      )
-    }
-  }, [
-    exercises,
-    hasProgress,
-    lastCompletedSolution,
-    nextExercise?.slug,
-    nextSection,
-    userProgressStatus,
-  ])
```
joelhooks, line 204: once I started removing stuff this felt like it was not needed, but it may still be just wanted to reset a bit and rethink

apps/total-typescript/src/hooks/use-progress.ts
```diff
@@ -99,206 +69,65 @@ export const getNextSectionForProgress = ({
...
-    isCompleted,
-    percentCompleted,
-    completedLessons: completedResources,
-    status: nextExerciseStatus,
+    // used to determine if we have ANY completed lessons
+    isModuleStarted: !isEmpty(completedLessons),
+    isProgressLoaded:
+      nextLessonStatus === 'success' && userProgressStatus === 'success',
+    // used to determine if a specific lesson is completed
+    completedLessons: completedLessons,
```
joelhooks, line 109: consolidated the stuff logically and labeled it as actually used

apps/total-typescript/src/hooks/use-progress.ts
vojtaholik: I was thinking that a better approach would be to have `getNextLessonForProgress` trpc query which would always give us a correct "next up" lesson in correlation with user progress by skipping already completed ones in sequence.

vojtaholik:
works great! I'm not gonna approve just yet so learners don't loose checkmarks for too long.

PR #768, opened 2023-01-24
build a progress object on the server

joelhooks:
progress

PR #770, opened 2023-01-25
feat(tt): use the new moduleProgress query

vojtaholik:
this puts the new moduleProgress query (introduced in #768) in use on workshop pages to see how it feels before extracting it into a hook (do we even want to?) works great!

gif


apps/total-typescript/src/templates/workshop-template.tsx
```diff
@@ -164,16 +184,19 @@ const Header: React.FC<{workshop: Module}> = ({workshop}) => {
   )
 }
 
-const WorkshopSectionNavigator: React.FC<{workshop: Module}> = ({workshop}) => {
+const WorkshopSectionNavigator: React.FC<{
+  workshop: Module
```
joelhooks, line 188: instead, just calling again. no need to pass in as props, you can call the router as many times as needed

apps/total-typescript/src/templates/workshop-template.tsx
```diff
@@ -214,8 +244,13 @@ const WorkshopSectionNavigator: React.FC<{workshop: Module}> = ({workshop}) => {
 const SectionItem: React.FC<{
   section: Section
   workshop: Module
-}> = ({section, workshop}) => {
-  const {isCompleted, percentCompleted} = useSectionProgress({section: section})
+  moduleProgress: ModuleProgress | null | undefined
+}> = ({section, workshop, moduleProgress}) => {
```
joelhooks, line 248: same here, instead of props just call the router. it caches and batches so it will optimize under the hood

vojtaholik: very cool! will fix

joelhooks (approved):

PR #772, opened 2023-01-26
fixes build and types

joelhooks:
fixing

made sure that the library versions were all aligned and explicitly set a type for the trpc context in skill-lesson

I created an entire type for the `createContext` but to appease the type checker it actually only required that `Session` was imported in trpc-context.ts

This is beyond my understanding but it often works for this type of error along with aligning versions


packages/skill-lesson/trpc/trpc-context.ts
```diff
@@ -1,8 +1,19 @@
 import {PrismaClient} from '@skillrecordings/database'
 import * as trpc from '@trpc/server'
 import * as trpcNext from '@trpc/server/adapters/next'
+import {NextApiRequest, NextApiResponse} from 'next'
+import {Session} from 'next-auth'
```
joelhooks, line 5: this is what actually solves the issue, but explicitly defining the type below is less mystery meat

PR #775, opened 2023-01-27
feat(skill-api): get ck tag id per product

vojtaholik:
screenshot

gif


packages/skill-api/src/core/services/convertkit/convertkit-tag-purchase.ts
```diff
@@ -4,16 +4,24 @@ import {
...
       throw new Error('‼️ set convertkit purchase tag id')
     }
-    const subscriber = await tagSubscriber(
-      email,
-      process.env.CONVERTKIT_PURCHASED_TAG_ID,
-    )
+
+    const subscriber = await tagSubscriber(email, convertkitPurchasedTagId)
 
     const purchasedOnFieldName =
```
vojtaholik, line 26: TODO: investigate why setting a custom field isn't working
image

joelhooks (approved):
We can also great tags on the fly per product, which is interesting and would mean we wouldn't have to manage them in Sanity (I actually think this should probably be in Prisma since it's commerce related and not content)

PR #771, opened 2023-01-25
Replace `testing-javascript` with app that includes Sanity and tRPC - TJS-13

jbranchaud:
- feat(tjs): use more modern CSA gen for app
- feat(tjs): add Sanity (and missing trpc) files
- feat(tjs): add Sanity config vars, import scripts

The bulk of the files in this PR are the ones coming from the `create-skill-app` setup of Sanity, all the documents and objects that support products with modules with lessons with resources.

This PR includes a couple scripts for importing all Testing JavaScript data into Sanity. I ran it against the `production` dataset for the `testing-javascript` Sanity project. Here is what the studio looks like with all that data (note: still need to deploy the Studio).

CleanShot 2023-01-25 at 12 54 16@2x

bansh

PR #776, opened 2023-01-30
feat(csa): remove comment from abilities template

jbranchaud:
This comment has already been addressed a couple times and shouldn't be
part of a generated skill app template. https://github.com/skillrecordings/products/pull/771#discussion_r1087527414

white


joelhooks (approved):
I think this can be deleted because it redundantly sets the cookie it just read, but was disagreeing about the appropriateness of a trpc router as a spot to set a cookie

PR #779, opened 2023-02-01
email change and transfer request

joelhooks:
There are some "edges" here that can be smoothed down, but this should pave the way for this functionality. Only implemented for TT so far, but is transportable.

https://www.loom.com/share/808ff48faf8e419b88727f0756f660a5

transfer


packages/commerce-server/src/record-new-purchase.ts
```diff
@@ -142,6 +143,11 @@ export async function recordNewPurchase(checkoutSessionId: string): Promise<{
     purchaseType = NEW_INDIVIDUAL_PURCHASE
   }
 
+  await createPurchaseUserTransfer({
+    purchaseId: purchase.id,
+    sourceUserId: user.id,
+  })
```
jbranchaud, line 149: You could push this up into `createMerchantChargeAndPurchase` in `prisma-api.ts`. Then it is closer to the actual purchase creation and you can bundle it into the database transaction that the other purchase concerns are wrapped in:

CleanShot 2023-02-01 at 13 45 15@2x

apps/total-typescript/src/trpc/routers/purchaseUserTransfer.ts
```diff
@@ -0,0 +1,396 @@
...
+      })
+
+      return await prisma.purchaseUserTransfer.create({
+        data: {
+          purchaseId: purchaseUserTransfer.purchaseId,
+          transferState: 'AVAILABLE',
+          expiresAt: purchaseUserTransfer.expiresAt,
+          sourceUserId: purchaseUserTransfer.sourceUserId,
+        },
+      })
```
jbranchaud, line 146: these could be wrapped in a transaction

apps/total-typescript/src/purchase-transfer/purchase-transfer.tsx
```diff
@@ -0,0 +1,145 @@
...
+      email: data.email,
+    })
+  }
+
+  return (
+    <div>
+      <form
+        className="flex w-full flex-col gap-2 text-left md:flex-row"
+        onSubmit={handleSubmit(onSubmit)}
+      >
```
jbranchaud, line 52: pretty awesome the way trpc trims down the logic/complexity in the component layer of this feature

jbranchaud (approved):
Looks great, I appreciate the supplemental loom feature walkthrough. I left a couple non-blocking comments.

PR #785, opened 2023-02-06
update(pt): display transfer purchase ui on purchase page and polish its ui

vojtaholik:
- display transfer purchase UI on `purchases/[purchaseId]` pages
- move `purchases` route templates into separate directory inside `src`
- add ability to style transfer purchase ui separately (theming)
- polish styles

screenshot

gif


joelhooks (approved):
we should apply these changes to Total TypeScript as well by moving the components into `skill-lesson`

PR #799, opened 2023-02-14
feat: forward invoice functionality

vojtaholik:
screenshots

gif

### DIFF
--- a/apps/protailwind/src/pages/invoices/[merchantChargeId].tsx
+++ b/apps/protailwind/src/pages/invoices/[merchantChargeId].tsx
@@ -10,8 +10,6 @@ import fromUnixTime from 'date-fns/fromUnixTime'
 import Layout from 'components/layout'
 import format from 'date-fns/format'
 import {prisma} from '@skillrecordings/database'
-import {getCurrentAbility} from '@skillrecordings/ability'
-import {getToken} from 'next-auth/jwt'
 import {trpc} from '../../trpc/trpc.client'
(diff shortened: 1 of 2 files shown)


jbranchaud:
I just check out this branch to try this out. The invoice show URL when used by someone who isn't authenticated (or who doesn't have an account) is in accessible because we short-circuit the user back to the home page.

Maybe we can special case the Invoice Show page as part of the auth-checking redirect logic?

jbranchaud:
> I just checked out this branch to try this out. The invoice show URL when used by someone who isn't authenticated (or who doesn't have an account) is in accessible because we short-circuit the user back to the home page.
> 
> Maybe we can special case the Invoice Show page as part of the auth-checking redirect logic?

Just looked a little closer at these page components. The Invoice Show page does an ability check and if the user is not allowed to view the invoice, then they are redirected to the Invoices Index page.

https://github.com/skillrecordings/products/blob/main/apps/total-typescript/src/pages/invoices/%5BmerchantChargeId%5D.tsx#L67-L72

And the Invoices Index page does the same thing, ability check and redirect user to home page if they don't have permission.

https://github.com/skillrecordings/products/blob/main/apps/total-typescript/src/pages/invoices/index.tsx#L19-L32

vojtaholik:
@jbranchaud great catch! I made it visible to anyone https://github.com/skillrecordings/products/pull/799/commits/afbe5afb434f5aace5195b93bf499257bb20880c

jbranchaud (approved):
💯

PR #800, opened 2023-02-14
jbranchaud/tt 106 make sure buy more seats route create a

jbranchaud:
### feat(tt): handle another bulk purchase scenario

If a user has already purchased a product and then while not
authenticated they go back through the purchase flow, we should be able
to recognize that they have already purchased and create a bulk coupon
for their new purchase that they can share with their team.

Here is the post-purchase screen that they get now if they are making a subsequent individual purchase of the product.

CleanShot 2023-02-14 at 11 30 01@2x

doritos

### DIFF
--- a/packages/database/src/prisma-api.ts
+++ b/packages/database/src/prisma-api.ts
@@ -394,6 +394,26 @@ export function getSdk(
         },
       })
 
+      // if this user has already purchased this product, then an additional
+      // purchase (even for only 1 seat) should be processed as a bulk
+      // purchase so that they can distribute the seat to someone else.
+      //
+      // if an existingPurchase is found, but no existingBulkCoupon is found,
+      // then the logic below will account for that and create a new bulk
+      // coupon for the requested quantity.
+      const existingPurchase = await ctx.prisma.purchase.findFirst({
+        where: {
+          productId,
+          userId,
+          status: 'Valid',
+        },
+      })
+
+      // TODO: This doesn't seem to be looking up the bulk coupon based
+      // on the product ID which will become an issue when any particular
+      // app has more than one product.
+      // E.g. should probably account for `restrictedToProductId`
+      //
       // Check if this user has already purchased a bulk coupon, in which
       // case, we'll be able to treat this purchase as adding seats.
       //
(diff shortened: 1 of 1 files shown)


packages/database/src/prisma-api.ts
```diff
@@ -394,6 +394,26 @@ export function getSdk(
...
+      // if an existingPurchase is found, but no existingBulkCoupon is found,
+      // then the logic below will account for that and create a new bulk
+      // coupon for the requested quantity.
+      const existingPurchase = await ctx.prisma.purchase.findFirst({
+        where: {
+          productId,
+          userId,
+          status: 'Valid',
+        },
+      })
```
jbranchaud, line 410: determine if there is an existing purchase of this product which we factor in below to determine if this new purchase should be considered a _bulk_ purchase

packages/database/src/prisma-api.ts
```diff
@@ -394,6 +394,26 @@ export function getSdk(
...
+
+      // TODO: This doesn't seem to be looking up the bulk coupon based
+      // on the product ID which will become an issue when any particular
+      // app has more than one product.
+      // E.g. should probably account for `restrictedToProductId`
+      //
```
jbranchaud, line 416: I think we have a gap here that hasn't been an issue for single-product apps, but we need to be a little more specific with the query. I'll deal with this in a separate PR.

packages/database/src/prisma-api.ts
```diff
@@ -458,11 +481,12 @@ export function getSdk(
         },
       })
 
+      const oneWeekInMilliseconds = 1000 * 60 * 60 * 24 * 7
       const purchaseUserTransfer = ctx.prisma.purchaseUserTransfer.create({
         data: {
           sourceUserId: userId,
           purchaseId: purchaseId,
-          expiresAt: new Date(Date.now() + 1000 * 60 * 60 * 24 * 7),
+          expiresAt: new Date(Date.now() + oneWeekInMilliseconds),
```
jbranchaud, line 489: I stared at this expression long enough that I decided it was a good idea to extract it as a descriptive variable name.

packages/database/src/prisma-api.ts
joelhooks: 👍

packages/database/src/prisma-api.ts
joelhooks: 🤡

joelhooks (approved):

PR #801, opened 2023-02-14
feat(commerce): more accurate quantity on invoices

jbranchaud:
(_please review and leave comments, no merge until schema changes have been deployed to planetscale_)

The only value you we had to go off of when generating invoices for a user's bulk purchase was the `Coupon.maxUses` value. While this works fine for a user who has made a single bulk purchase, it quickly results in inaccurate invoices for anyone who has done 2 or more bulk purchases. That `maxUses` values keeps getting added to, so the latest invoice as well as older ones will display a quantity that is the sum of all seats ever purchased by this user.

The fix is to add a `MerchantPurchase` table that ties to a Purchase and stores the Stripe `checkoutSessionId`. This checkout session ID can be used to look up the exact quantity involved in the purchase. We can use this value in the invoice line item instead of `maxUses`.

To do:
- [x] Before this PR can be merged, the schema changes need to be deployed to the various skill-stack planetscale databases.
- [ ] After this PR is merged, we can optionally go through and backfill `MerchantPurchase` records for existing purchases. I say _optionally_ because I think it would be reasonable to start recording this going forward and not go to a lot of effort to back fill for existing purchases.

Here is an example of multiple bulk purchases (5 and then 3 seats) totaling up to 8 maxUses for a bulk coupon:

CleanShot 2023-02-14 at 16 51 57@2x

And here is the invoice for the purchase of 5 seats:

CleanShot 2023-02-14 at 16 52 22@2x

nachos


apps/protailwind/src/pages/invoices/[merchantChargeId].tsx
```diff
@@ -43,6 +43,18 @@ export const getServerSideProps: GetServerSideProps = async ({
...
+      if (purchase?.merchantPurchase?.checkoutSessionId) {
+        const checkoutSession = await stripe.checkout.sessions.retrieve(
+          purchase?.merchantPurchase?.checkoutSessionId,
+          {expand: ['line_items']},
+        )
+
+        quantity = checkoutSession.line_items?.data[0].quantity || 1
+      } else if (bulkCoupon) {
+        quantity = bulkCoupon.maxUses
+      }
```
Comment (jbranchaud) [inline apps/protailwind/src/pages/invoices/[merchantChargeId].tsx:56]: determine (server-side) the `quantity` to be used in the invoice (rendered below)

We attempt to look up the quantity from Stripe via the checkout session ID. If that is unavailable, then we fallback to the `bulkCoupon.maxUses` which is going to work for most existing purchases anyway.

packages/database/prisma/schema.prisma
```diff
@@ -154,6 +155,14 @@ model MerchantProduct {
   merchantPrices    MerchantPrice[]
 }
 
+model MerchantPurchase {
+  id                String          @id @default(uuid())
+  checkoutSessionId String
+  merchantAccountId String
+  merchantAccount   MerchantAccount @relation(fields: [merchantAccountId], references: [id], onDelete: Cascade)
+  purchase          Purchase[]
+}
```
jbranchaud, line 164: a new table for storing the `checkoutSessionId` related to a `Purchase`.

packages/database/src/prisma-api.ts
```diff
@@ -116,6 +116,11 @@ export function getSdk(
               bulkCouponPurchases: true,
             },
           },
+          merchantPurchase: {
+            select: {
+              checkoutSessionId: true,
+            },
+          },
```
jbranchaud, line 123: this SDK function is what the invoice show page uses to look up the purchase, we lookup and include the `checkoutSessionId` in the result.

packages/database/src/prisma-api.ts
```diff
@@ -447,6 +454,16 @@ export function getSdk(
...
 
+      const merchantPurchaseId = v4()
+
+      const merchantPurchase = ctx.prisma.merchantPurchase.create({
+        data: {
+          id: merchantPurchaseId,
+          checkoutSessionId,
+          merchantAccountId,
+        },
+      })
```
jbranchaud, line 465: Create a `MerchantPurchase` anytime a new purchase is made going forward. This SDK function isn't called for $0 coupon redemption purchases, so we'll only record these when there is a Stripe check session to tie it to.

apps/protailwind/src/pages/invoices/[merchantChargeId].tsx
```diff
@@ -43,6 +43,18 @@ export const getServerSideProps: GetServerSideProps = async ({
       )
       const bulkCoupon = purchase && purchase.bulkCoupon
 
+      let quantity = 1
+      if (purchase?.merchantPurchase?.checkoutSessionId) {
+        const checkoutSession = await stripe.checkout.sessions.retrieve(
+          purchase?.merchantPurchase?.checkoutSessionId,
```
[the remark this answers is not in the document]
Comment (joelhooks) [reply apps/protailwind/src/pages/invoices/[merchantChargeId].tsx:49]: oh the conditional access is redundant because of the if above

packages/database/prisma/schema.prisma
joelhooks: we've gone to pretty great lengths to not encode stripe specific data in the context of "merchant" in this schema

IMO probably need a table for this and not a simple field. 

MerchantSession or similar would fit the pattern with an "identifier"

Hard to answer the "what about paypal" question, but "checkoutSessionId" is definitely very stripe

jbranchaud: that makes a lot of sense. looking more closely at the comment you made on an earlier PR, I see you had suggested `identifier` there as well. https://github.com/skillrecordings/products/pull/662#discussion_r1043683711

CleanShot 2023-02-15 at 12 10 40@2x

So I guess how this could work then is: if I have a `Purchase` that is tied to a `MerchantSession` whose `MerchantAccount` is `stripe`, then the `MerchantSession.identifier` is expected to be a `checkoutSessionId` and I can use it as such.

apps/protailwind/src/pages/invoices/[merchantChargeId].tsx
```diff
@@ -3,91 +3,69 @@ import {DownloadIcon} from '@heroicons/react/outline'
...
-      })
-
-      const purchase = await getPurchaseForStripeCharge(
-        merchantCharge.identifier,
-      )
-      const bulkCoupon = purchase && purchase.bulkCoupon
-
-      const product = await getProduct({
-        where: {id: purchase?.productId},
-      })
```
Comment (jbranchaud) [inline apps/protailwind/src/pages/invoices/[merchantChargeId].tsx:48]: rather than having to copy this logic into each app's invoice page, I moved it to a skill-lesson tRPC query

packages/database/prisma/schema.prisma
```diff
@@ -154,6 +155,14 @@ model MerchantProduct {
   merchantPrices    MerchantPrice[]
 }
 
+model MerchantSession {
+  id                String          @id @default(uuid())
+  identifier        String
+  merchantAccountId String
+  merchantAccount   MerchantAccount @relation(fields: [merchantAccountId], references: [id], onDelete: Cascade)
+  purchase          Purchase[]
+}
```
jbranchaud, line 164: This is called `MerchantSession` now and it has an `identifier` which in the case of Stripe merchant purchases will be a `checkoutSessionId`.

packages/skill-lesson/trpc/routers/invoices.ts
```diff
@@ -0,0 +1,67 @@
+import {z} from 'zod'
+import {publicProcedure, router} from '../trpc.server'
+import {getSdk} from '@skillrecordings/database'
+import {stripe} from '@skillrecordings/commerce-server'
+
+export const invoicesRouter = router({
```
jbranchaud, line 6: tRPC router for the invoice page, it fetches all the relevant records (charge, purchase, bulk coupon) and then determines the quantity with logic for when to fall back to the bulk coupon's quantity.

jbranchaud:
Planetscale schema changes have been deployed:
- Total TypeScript https://app.planetscale.com/skill-recordings/total-typescript/deploy-requests/10
- Testing Accessibility https://app.planetscale.com/skill-recordings/testing-accessibility/deploy-requests/19
- Pro Tailwind https://app.planetscale.com/skill-recordings/pro-tailwind/deploy-requests/5
- TypeScript Course https://app.planetscale.com/skill-recordings/typescript-course/deploy-requests/3

apps/protailwind/src/pages/invoices/[merchantChargeId].tsx
Comment (joelhooks) [reply apps/protailwind/src/pages/invoices/[merchantChargeId].tsx:48]: this is great ⭐️

PR #804, opened 2023-02-15
jbranchaud/tt 176 adding seats to single seat bulk

jbranchaud:
- feat: bulk coupons can have *1* or more seats
- feat: filter existing bulk coupon by product ID
- cleanup: remove TODO comments leading to this PR

hbo industry


packages/database/src/prisma-api.ts
```diff
@@ -409,22 +409,12 @@ export function getSdk(
...
-      // least one `bulkCouponPurchase` attached to this Coupon, then it is a
-      // bulk coupon for this user.
+      // Check if this user has already purchased a bulk coupon for this
+      // product, in which case, we'll be able to treat this purchase as
+      // adding seats.
       const existingBulkCoupon = await ctx.prisma.coupon.findFirst({
         where: {
-          maxUses: {
-            gt: 1,
-          },
```
jbranchaud, line 427: This logic was from when we considered bulk purchase to be more than 1 seat. Since then, we allow a user to make a bulk purchase starting at 1 seat.

packages/database/src/prisma-api.ts
```diff
@@ -409,22 +409,12 @@ export function getSdk(
...
+          restrictedToProductId: productId,
```
jbranchaud, line 417: For apps that will eventually need to support multiple products, we'll need this `restrictedToProductId` filter in here so that we are looking up the correct bulk coupon.

joelhooks (approved):

PR #814, opened 2023-02-22
use a callback for when a module ends

joelhooks:
runaway


vojtaholik (approved):
One detail I've noticed is that green checkmark in lesson list wouldn't appear since `invalidate` doesn't trigger a re-render.

PR #818, opened 2023-02-23
invalidate queries when mutation is settled

joelhooks:
had to use the query client because tRPC invalidate wasn't causing a rerender 🤔

also update react-query and trpc

invalid


apps/total-typescript/src/templates/exercise-template.tsx
```diff
@@ -50,9 +51,7 @@ const ExerciseTemplate: React.FC<{
           {lessonSlug: router.query.lesson as string},
           {
             onSettled: () => {
-              utils.moduleProgress.bySlug.invalidate({
-                slug: module.slug.current,
-              })
+              client.invalidateQueries()
```
joelhooks, line 54: invalidation via the query client functions as expected and re-renders the with updated progress across the page

apps/total-typescript/src/trpc/trpc.client.ts
```diff
@@ -8,6 +8,24 @@ import type {AppRouter} from 'trpc/routers/_app'
...
+         * The order here allows route changes in `onSuccess` without
+         * having a flash of content change whilst redirecting.
+         **/
+        // Calls the `onSuccess` defined in the `useQuery()`-options:
+        await opts.originalFn()
+        // Invalidate all queries in the react-query cache:
+        await opts.queryClient.invalidateQueries()
+      },
+    },
+  },
```
joelhooks, line 28: EVERY mutation invalidates the cache and reloads

apps/total-typescript/src/templates/exercise-template.tsx
```diff
@@ -38,24 +38,14 @@ const ExerciseTemplate: React.FC<{
...
-        addProgressMutation.mutate(
-          {lessonSlug: router.query.lesson as string},
-          {
-            onSettled: () => {
-              utils.moduleProgress.bySlug.invalidate({
-                slug: module.slug.current,
-              })
-            },
-          },
-        )
```
joelhooks, line 58: we can remove all of these

vojtaholik (approved):
works great 👍

PR #819, opened 2023-02-24
update(tt): lesson list to close prev section

vojtaholik:
testing this out to see how it feels.

gif


joelhooks (approved):
felt great to me

PR #828, opened 2023-03-01
feat(tt): if a Stripe coupon is used, record it

jbranchaud:
More details in Roam (https://roamresearch.com/#/app/egghead/page/R83NFE69j) and the linear issue.

Essentially, there was nothing in place before this to attempt to tie a purchase's `Coupon` record to a corresponding Stripe `MerchantCoupon`.

I just tested this out with a 5% off bulk purchase locally and it is now tying those records together as expected.

the office kevin

### DIFF
--- a/packages/commerce-server/src/record-new-purchase.ts
+++ b/packages/commerce-server/src/record-new-purchase.ts
@@ -52,7 +52,7 @@ export async function stripeData(options: StripeDataOptions) {
 
   // extract MerchantCoupon identifier if used for purchase
   const discount = first(lineItem.discounts)
-  let stripeCouponId = discount?.discount.coupon.id
+  const stripeCouponId = discount?.discount.coupon.id
 
   return {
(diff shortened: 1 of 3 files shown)


packages/commerce-server/src/record-new-purchase.ts
```diff
@@ -50,13 +50,18 @@ export async function stripeData(options: StripeDataOptions) {
   const stripeChargeId = stripeCharge?.id as string
   const stripeChargeAmount = stripeCharge?.amount || 0
 
+  // extract MerchantCoupon identifier if used for purchase
+  const discount = first(lineItem.discounts)
+  let stripeCouponId = discount?.discount.coupon.id
```
jbranchaud, line 55:
```suggestion
  const stripeCouponId = discount?.discount.coupon.id
```

joelhooks (approved):

PR #829, opened 2023-03-01
feat(pt): new footer bio image + text tweaks

simonswiss:
One can never bring too much corgi to the Pro Tailwind brand 🐾

CleanShot 2023-03-02 at 09 20 04@2x


joelhooks (approved):
🐶

PR #835, opened 2023-03-06
hotfix(pt): ability to view content

vojtaholik:
gif

### DIFF
--- a/packages/skill-lesson/lib/products.ts
+++ b/packages/skill-lesson/lib/products.ts
@@ -34,8 +34,8 @@ export const getActiveProducts = async () =>
     ],
   }
 
-export const getProductModules = async (productIds: string[]) => {
-  const product = await sanityClient.fetch(
+export const getProducts = async (productIds: string[]) => {
+  const products = await sanityClient.fetch(
     groq`*[_type == 'product' && productId in $productIds][]{
     _id,
     productId,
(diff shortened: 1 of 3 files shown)


packages/skill-lesson/lib/products.ts
```diff
@@ -47,7 +47,26 @@ export const getProductModules = async (productIds: string[]) => {
...
+    {
+      productIds,
+    },
+  )
+  return products.map((product: any) => {
+    // product can also be a module with product resource
+    // TODO: adjust logic in here: https://github.com/skillrecordings/products/blob/main/packages/skill-lesson/utils/ability.ts#L72
+    // so that this object doesn't have to be reshaped
+    return {...product, modules: [product]}
+  })
```
vojtaholik, line 69: this query will get workshops with product resource which is what we have on PT. I had to reshape the object a little bit because of the logic here which is expecting `modules` to be present: https://github.com/skillrecordings/products/blob/main/packages/skill-lesson/utils/ability.ts#L72

should likely be improved, but works fine and will enable users to view content again.

packages/skill-lesson/trpc/routers/modules.ts
```diff
@@ -42,7 +42,9 @@ export const modulesRouter = router({
         const user = UserSchema.parse(token)
         const productsPurchased =
           user.purchases?.map((purchase) => purchase.productId) || []
-        purchasedModules = await getProductModules(productsPurchased)
+        const products = await getProducts(productsPurchased)
+        const moduleProducts = await getModuleProducts(productsPurchased)
+        purchasedModules.push(...products, ...moduleProducts)
```
vojtaholik, line 47: I'm joining results of both queries so that it works for all products.

PR #833, opened 2023-03-04
add a `Restricted` status to purchase

joelhooks:
pre-cursor to enforcing regional purchases only allowed to playback in the region purchased for

restricted

PR #836, opened 2023-03-07
Address my comments on the 'Restricted' purchase PR

jbranchaud:
I looking over the 'Restricted' purchase PR (https://github.com/skillrecordings/products/pull/833) and left some comments. This PR addresses the comments that I thought were worth getting into the codebase.

Highlights:
- Fix a couple boolean logic precedence issues in places that are checking if a Purchase is either `Valid` or `Restricted`.
- Rework "Don't create PPP promo code for bulk purchase" logic
- Add Zod and query parsing to the Stripe Checkout service to get better types flowing throughout that function

mac


packages/ability/src/ability.ts
```diff
@@ -77,8 +77,8 @@ export function defineRulesForPurchases(purchases: any[]) {
       productId: {
         $in: purchases?.map(
           (purchase: any) =>
-            (purchase.productId && purchase.status === 'Valid') ||
-            purchase.status === 'Restricted',
+            purchase.productId &&
+            ['Valid', 'Restricted'].includes(purchase.status),
```
jbranchaud, line 81: The parentheses were throwing off the precedence of this check. I thought it would be simpler to do an `includes` check for the `purchase.status` instead.

packages/skill-api/package.json
```diff
@@ -33,27 +33,28 @@
...
     "micro": "^9.3.4",
+    "mjml": "^4.12.0",
     "node-html-markdown": "^1.2.0",
+    "pluralize": "^8.0.0",
     "preact": "^10.7.2",
     "preact-render-to-string": "^5.2.0",
     "sanitize-html": "^2.7.0",
     "uuid": "^8.3.2",
-    "pluralize": "^8.0.0"
+    "zod": "^3.21.2"
```
jbranchaud, line 57: Adding `Zod` to this package. Looks like pnpm rearranged some of the deps while doing that.

packages/skill-api/src/core/services/stripe-checkout.ts
```diff
@@ -155,22 +177,21 @@ export async function stripeCheckout({
           coupon: coupon.id,
         })
       } else if (merchantCoupon && merchantCoupon.identifier) {
-        // no ppp for bulk purchases
-        const isNotPPP = merchantCoupon.type !== 'ppp'
-        if (isNotPPP || quantity === 1) {
-          const {id} = await stripe.promotionCodes.create({
```
jbranchaud, line 161: I kept trying to reason through this conditional, but found it confusing. What I was understanding it distilled down to was, "if they are purchasing multiple seats and the coupon code is PPP, then don't apply that coupon." So I added some logic earlier up in the function that nulls-out the `merchantCoupon` variable if this is a multi-seat purchase. That simplifies things downstream.

packages/skill-api/src/core/services/stripe-checkout.ts
```diff
@@ -119,7 +141,7 @@ export async function stripeCheckout({
           upgradeFromPurchase,
       )
 
-      const TWELVE_FOUR_HOURS_FROM_NOW = Math.floor(
+      const TWELVE_HOURS_FROM_NOW = Math.floor(
```
jbranchaud, line 144: I noticed this variable renaming from `TWENTY_FOUR_HOURS_FROM_NOW` to `TWELVE_FOUR_HOURS_FROM_NOW` and fixed it up throughout this file.

packages/skill-api/src/core/services/stripe-checkout.ts
```diff
@@ -32,33 +33,46 @@ export async function stripeCheckout({
...
+          quantity: z.number().optional(),
+          bulk: z.boolean().default(false),
+        })
+        .transform(({quantity, userId, ...rest}) => {
+          return {
+            userId: userId || token?.sub,
+            quantity: quantity || 1,
+            ...rest,
+          }
+        })
```
jbranchaud, line 52: Added `Zod` parsing of the incoming query params which also sets defaults in the `transform`. This means we have typed values throughout this function instead of `any`s, so we no longer have to stick ` as string` all over the place.

packages/skill-api/src/core/services/stripe-checkout.ts
```diff
@@ -94,13 +108,21 @@ export async function stripeCheckout({
...
+
+      // Null out the `merchantCoupon` if it is a PPP coupon when the
+      // quantity is greater than 1. You cannot apply PPP to a purchase
+      // of multiple seats.
+      if (merchantCoupon?.type === 'ppp' && quantity > 1) {
+        // TODO: Let support know that we may have showed a discounted
+        // price to a user when we didn't mean to.
+
+        merchantCoupon = null
+      }
```
jbranchaud, line 125: Here is where the coupon logic is handled - it basically clears the `merchantCoupon` if it is PPP and the quantity is greater than 1.

I'm also thinking that if this service is handling a scenario where we have a PPP coupon and a multi-seat purchase, then it is possible one of the client apps has a bug and that we displayed the wrong price to a user. In that case, we should notify support so that they are aware this is happening. Would that make sense as some kind of `Sentry` call?

packages/skill-api/src/core/services/stripe-checkout.ts
```diff
@@ -155,22 +177,21 @@ export async function stripeCheckout({
...
+        })
       }
 
+      // TODO: Could we throw this error much earlier in the function and
+      // avoid generating promo codes and making other queries when we are
+      // eventually going to throw anyway?
       if (!loadedProduct) {
-        throw new CheckoutError('No product was found', productId as string)
+        throw new CheckoutError('No product was found', productId)
       }
```
jbranchaud, line 195: I think it would make sense to push this all up to where `loadedProduct` gets queried so that we can circumvent a bunch of logic and queries that are going to get thrown away anyway.

Thoughts?

joelhooks (approved):

PR #839, opened 2023-03-10
feat(db): make customer id field optional

jbranchaud:
The prisma relation was already marked as optional, so the id field
itself should be nullable as well.

I need this for the Testing JavaScript data import because there are tons of `MerchantCharge` records that do not have a `MerchantCustomer` because of how Stripe purchases used to be done.

- [ ] Once these changes get the greenlight, I need to set up schema Deploy Requests across all the planetscale databases.

CleanShot 2023-03-10 at 12 06 25@2x


jbranchaud:
going a different direction with this

PR #843, opened 2023-03-13
feat(tt): module testimonials

vojtaholik:
screenshot

gif


joelhooks:
Looks great! Any way that we could get these inline or otherwise higher in the page as a "callout" versus a "reviews" section down in the trenches?

vojtaholik:
> Looks great! Any way that we could get these inline or otherwise higher in the page as a "callout" versus a "reviews" section down in the trenches?

inline testimonial

PR #846, opened 2023-03-15
Allow for tutorial module sections

Creeland:
This sets up our client so that tutorials can support sections. 

It was mostly recreating what has already been done to modules. But, the tricky bit was setting up the redirects so that we don't break any links that point to exercises without sections.

I'm not sure what the best order of operations is for migrating the data and merging this PR. Sections break the existing build but would break this one. Personally, I would probably migrate the data and then merge but I definitely want some input on this. @vojtaholik @jbranchaud 

juggling


joelhooks (commented):
nothing here stood out to me and it looks like it should work. it just needs to be timed right so there isn't a lot of downtime.

PR #850, opened 2023-03-17
feat(sks): abstract module lesson list components

vojtaholik:
this abstracts `lesson-list` components while making it themeable. I didn't make them shared yet (there are handful more components in `video` dir awaiting similar treatment), but it's all prepared for that since both PT and TT are now using identical components with styles driven by unique stylesheets. this doesn't include a default stylesheet with sensible fallback styles, that shall come in later.

file-tree

screenshots

gif

### DIFF
--- a/apps/protailwind/src/video/module-lesson-list/lesson-list.tsx
+++ b/apps/protailwind/src/video/module-lesson-list/lesson-list.tsx
@@ -359,36 +359,7 @@ const Lesson = React.forwardRef<
         />
         {isExpanded && (
           <ul>
-            {lesson._type === 'exercise' && (
-              <>
-                {lessonResourceRenderer ? (
-                  lessonResourceRenderer(path, module, lesson, section)
-                ) : (
-                  <>
-                    <ProblemLink
-                      module={module}
-                      exercise={lesson}
-                      section={section}
-                      path={path}
-                    />
-                    <SolutionLink
-                      module={module}
-                      lesson={lesson}
-                      section={section}
-                      path={path}
-                    />
-                  </>
-                )}
-              </>
-            )}
-            {lesson._type === 'explainer' && (
-              <ExplainerLink
-                lesson={lesson}
-                module={module}
-                section={section}
-                path={path}
-              />
-            )}
+            <>{lessonResourceRenderer(path, module, lesson, section)}</>
           </ul>
         )}
       </li>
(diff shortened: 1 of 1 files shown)


apps/total-typescript/src/trpc/routers/stackblitz-resource.ts
```diff
@@ -1,23 +0,0 @@
-import {publicProcedure, router} from '@skillrecordings/skill-lesson'
-import {z} from 'zod'
-import {getExercise} from '../../lib/exercises'
-
-export const stackblitzResourceRouter = router({
```
vojtaholik, line 5: I replaced this with `resourcesRouter` which we used on PT and implies more generic usage.

apps/protailwind/src/lib/exercises.ts
```diff
@@ -42,6 +43,7 @@ export const ExerciseSchema = z
         _key: z.string(),
         videoResourceId: z.nullable(z.string()).optional(),
         transcript: z.nullable(z.any().array()).optional(),
+        stackblitz: z.nullable(z.string()).optional(),
```
joelhooks, line 46: my approach to this, instead of polluting our entire space with resource types that are never used is to treat resource loading as a separate concern from an "exercise" (lesson, generally) and to the extent possible focus on the lesson type

because of the nature of Lesson resources they can be loaded clientsode (trpc) like this: 

https://github.com/skillrecordings/products/blob/main/apps/total-typescript/src/trpc/routers/stackblitz-resource.ts

Now, we can load a given Lesson's resources with the lesson ID and the type/implementation can be LOCAL to the project and we don't have to do stuff like the zod changes here polluting this project with stuff that will never be used

apps/total-typescript/src/video/module-lesson-list/large-screen-module-lesson-list.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const LargeScreenModuleLessonList: React.FC<{
+  exerciseResourcesRenderer: (
+    path: string,
+    module: Module,
+    lesson: Lesson,
+    section?: Section,
+  ) => void
+  module: Module
+  path: string
+}> = ({module, path, exerciseResourcesRenderer}) => {
```
joelhooks, line 17: the renderer should have a default that displays minimum information (title?)

joelhooks (approved):
lessonResourceRenderer still probably needs a default component but this looks good to me

apps/total-typescript/src/video/module-lesson-list/large-screen-module-lesson-list.tsx
vojtaholik: there will always be a lesson title rendered: https://github.com/skillrecordings/products/blob/main/apps/protailwind/src/video/module-lesson-list/lesson-list.tsx#L353
or am I missing something?

PR #851, opened 2023-03-17
Remove duplicate Workshops from Testing JavaScript Sanity Import

jbranchaud:
I already used this updated script to re-import the production Sanity data. This PR is to get those changes under version control.

- feat(kcd): ignore specific duplicate TJS courses
- feat(kcd): scrub WEBVTT from top of SRT text

workshop

### DIFF
--- a/apps/testing-javascript/src/data/sanity-import.ts
+++ b/apps/testing-javascript/src/data/sanity-import.ts
@@ -283,16 +283,39 @@ const importCourseData = async () => {
   const uniqueLessons: {[lessonSlug: string]: LessonData} = {}
 
   for (const product of products) {
+    // they have already been parsed, no reason to re-parse them individually
+    // here
     const parsedProduct = productSchema.parse(product)
 
     const {slug: productSlug} = parsedProduct
 
+    // ignore courses with the following slugs because they are the Series
+    // duplicates. The up-to-date courses are Playlist records with the other
+    // slugs.
+    // Series.where(id: [214, 406, 223, 407, 409, 408, 246, 410]).pluck(:slug)
+    const ignoredCourseSlugs = [
+      'static-analysis-testing-javascript-applications-71c1',
+      'test-node-js-backends',
+      'use-dom-testing-library-to-test-any-js-framework',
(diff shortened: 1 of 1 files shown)


apps/testing-javascript/src/data/sanity-import.ts
```diff
@@ -283,16 +302,39 @@ const importCourseData = async () => {
...
 
     const {slug: productSlug} = parsedProduct
 
+    // ignore courses with the following slugs because they are the Series
+    // duplicates. The up-to-date courses are Playlist records with the other
+    // slugs.
+    // Series.where(id: [214, 406, 223, 407, 409, 408, 246, 410]).pluck(:slug)
+    const ignoredCourseSlugs = [
+      'static-analysis-testing-javascript-applications-71c1',
+      'test-node-js-backends',
```
jbranchaud, line 317: I decided to hard-code this list of slugs that I pulled from `egghead-rails` rather than doing anything fancy. This ensures that only the 8 `Playlist` versions of the courses get pulled in as Workshop records.

apps/testing-javascript/src/data/sanity-import.ts
```diff
@@ -71,6 +71,25 @@ const CastingWordsSchema = z
...
   .nullish()
+  .transform((data) => {
+    if (data === null || data === undefined) {
+      return data
+    }
+
+    const {srt, ...rest} = data
+
+    if (srt.startsWith('WEBVTT\n\n')) {
+      const updatedSrt = srt.replace(/^WEBVTT\n\n/, '')
```
jbranchaud, line 82: I originally wrote a separate script back when we realized that the `WEBVTT` needed to be scrubbed from the SRT text. It seemed worthwhile to pull those changes back into this script to make it a one-step process if this ever needs to get run again.

apps/testing-javascript/src/data/sanity-import.ts
joelhooks: ah ya, this is way better. We don't need heavy reusability.

joelhooks (approved):

PR #853, opened 2023-03-21
feat: coupon loading more dynamic

joelhooks:
dynamic


packages/commerce-server/src/props-for-commerce.ts
```diff
@@ -24,9 +24,11 @@ export async function propsForCommerce({
     ? await getPurchasesForUser(token.sub as string)
     : false
 
+  console.log({products})
```
Creeland, line 27: Do we want to keep this console log?

packages/commerce-server/src/get-active-merchant-coupon.ts
```diff
@@ -14,7 +14,9 @@ export async function getActiveMerchantCoupon({
 
   let activeMerchantCoupon = null
 
-  const defaultCoupons = await getDefaultCoupon(productId)
+  const defaultCoupons = productId
+    ? await getDefaultCoupon([productId])
```
Creeland, line 18: Is getDefaultCoupon capable of accepting an array as an argument? [async getDefaultCoupon(productId?: string) {](https://github.com/skillrecordings/products/blob/d4c420f10094235a96a923ec2d3afc547276acd5/packages/database/src/prisma-api.ts#L644-L667)

Creeland (commented):
So, if I'm understanding the changes correctly. This now checks if one of several coupons

packages/commerce-server/src/get-active-merchant-coupon.ts
joelhooks: as part of this PR is is now

packages/commerce-server/src/props-for-commerce.ts
joelhooks, line 27:
```suggestion
```

vojtaholik (approved):

PR #854, opened 2023-03-21
Fixes PT-116: fix layout shift for purchased product

DrShpongle:
<details>
  <summary>Before:</summary>
  
</details>
<details>
  <summary>After:</summary>
  
</details>

Also updated seed data with new `productId`.

jump

### DIFF
--- a/apps/protailwind/src/templates/workshop-template.tsx
+++ b/apps/protailwind/src/templates/workshop-template.tsx
@@ -41,7 +41,8 @@ const WorkshopTemplate: React.FC<{
     product: SanityProduct
   }
   commerceProps?: CommerceProps
-}> = ({workshop, commerceProps}) => {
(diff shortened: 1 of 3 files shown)


apps/protailwind/src/templates/workshop-template.tsx
```diff
@@ -89,45 +90,52 @@ const WorkshopTemplate: React.FC<{
           )}
         </div>
 
-        {hasPurchased ? (
+        {!commersePropsLoading && (
```
vojtaholik, line 93: @DrShpongle would it be possible to avoid layout shift while still showing the loading skeleton? I think it's better to indicate that something is loading within the sidebar area instead of leaving it empty.

vojtaholik (changes requested):

PR #865, opened 2023-04-05
do not restart the video on tab focus

joelhooks:
restart

### DIFF
--- a/apps/protailwind/src/video/video.tsx
+++ b/apps/protailwind/src/video/video.tsx
@@ -82,7 +82,7 @@ export const Video: React.FC<
             hidden: displayOverlay,
           })}
         >
-          {canShowVideo && !loadingVideoResource ? (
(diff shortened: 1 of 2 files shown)


apps/protailwind/src/video/video.tsx
```diff
@@ -82,7 +82,7 @@ export const Video: React.FC<
             hidden: displayOverlay,
           })}
         >
-          {canShowVideo && !loadingVideoResource ? (
+          {canShowVideo && videoResource?.muxPlaybackId ? (
```
joelhooks, line 85: we want to make sure the data is available and do not care about the loading status of the resource

trpc (react-query) is `loading` on focus, but the video `muxPlaybackId` should be stable and present when reloading.

vojtaholik (approved):
🎉

PR #900, opened 2023-04-14
add the tutorial routes

joelhooks:
learning


apps/scriptkit/src/certificate/certificate-handler.server.tsx
```diff
@@ -0,0 +1,208 @@
+// @ts-nocheck
```
joelhooks, line 1: I had to do this in both places `@vercel/og` is used because the `tw` prop is a type error and I couldn't figure out the right switch to flip 😭

apps/scriptkit/src/components/navigation.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react'
 import Link from 'components/link'
 // import Logo from 'images/logo.svg'
 import {Disclosure} from '@headlessui/react'
-import {Bars3Icon, XMarkIcon} from '@heroicons/react/20/solid'
+import {ViewBoardsIcon, XIcon} from '@heroicons/react/solid'
```
joelhooks, line 5: @vojtaholik all the other projects are a version behind with the icons so I did a best guess here.

apps/scriptkit/src/data/next-seo.json
```diff
@@ -0,0 +1,36 @@
+{
```
joelhooks, line 1: I had to move these data files into `src` because of ts config

apps/scriptkit/src/pages/[user]/[script].tsx
```diff
@@ -145,7 +145,7 @@ export async function getStaticProps(context: any) {
 }
 
 export async function getStaticPaths() {
-  return await getScriptPaths()
+  return (await getScriptPaths()) as any
```
Comment (joelhooks) [inline apps/scriptkit/src/pages/[user]/[script].tsx:148]: this was a type error that I didn't take the time to unwind, needs a proper fix

johnlindquist (approved):

PR #906, opened 2023-04-19
feat(kcd): products should reference DB productId

jbranchaud:
I've already manually updated the Testing JavaScript Sanity `product` entities with these three `productId` values, but thought it would also be worth it to quickly fix/note the correctly `productId` values in the script.

note


apps/testing-javascript/src/data/sanity-import.ts
```diff
@@ -647,10 +647,23 @@ const importCourseData = async () => {
...
+      case 'Pro Testing':
+        productId = 'kcd_4f0b26ee-d61d-4245-a204-26f5774355a5'
+        break
+    }
+
     const product: Product = {
       _id: productSanityId,
       _type: 'product',
-      productId: productSanityId,
+      productId,
```
jbranchaud, line 666: I shouldn't be using this generated `productSanityId` value, I should be using the `kcd_uuid` values that already exist in the database. Accordingly, I created a switch statement to get those values properly set for each product.

joelhooks (approved):

PR #919, opened 2023-04-27
Build static assets first

johnlindquist:
Just changing order in the "build" step so we get the assets first.

first


johnlindquist:
Confirmed fixed the community scripts page: https://scriptkit-191omyy0y-skillrecordings.vercel.app/scripts

> It was broken before because it was loading the out-dated test data from the repo rather than the "latest" from the build scripts

zacjones93 (approved):
LGTM!

PR #923, opened 2023-05-02
React's most useful type helper

mattpocock:


vojtaholik (approved):
https://www.totaltypescript.com/react-component-props-type-helper

PR #940, opened 2023-05-12
feat(cs): adding tips

nicollguarnizo:
working


apps/colt-steele/src/utils/dexie.ts
```diff
@@ -0,0 +1,22 @@
+import Dexie, {Table} from 'dexie'
```
vojtaholik, line 1: this can be imported from `@skillrecordings/skill-lesson` package

apps/colt-steele/src/components/portable-text/index.tsx
```diff
@@ -0,0 +1,231 @@
+import React from 'react'
```
vojtaholik, line 1: this can be imported from `@skillrecordings/skill-lesson` package

packages/skill-lesson/video/video-transcript.tsx
```diff
@@ -22,10 +22,15 @@ export const VideoTranscript: React.FC<{
...
-      <h2 data-title="">Transcript</h2>
+      <h2 className="text-2xl font-bold" data-title="">
+        Transcript
+      </h2>
       <div data-transcript="">
         {typeof transcript === 'string' ? (
-          <ReactMarkdown components={markdownComponents}>
+          <ReactMarkdown
+            className="prose max-w-none pt-4"
+            components={markdownComponents}
```
vojtaholik, line 32: these should be styled via `data-` attribute in external stylesheet.

vojtaholik: https://github.com/skillrecordings/products/blob/main/packages/create-skill-app/templates/next/src/styles/globals.css#L9-L16

vojtaholik (approved):

PR #950, opened 2023-05-18
fix(tt): Fix Total TypeScript build

mattpocock:


vojtaholik (approved):
🙏

apps/total-typescript/globals.d.ts
```diff
@@ -0,0 +1,5 @@
+namespace React {
+  interface HTMLAttributes<T> {
+    tw?: string
+  }
+}
```
vojtaholik, line 5: @mattpocock I was wondering if you happen to know if this could be defined globally from within our [@skillrecordings/tsconfig](https://github.com/skillrecordings/products/tree/main/packages/tsconfig)? 😄

PR #952, opened 2023-05-18
feat(tt): added concept page

mattpocock:


apps/total-typescript/schemas/objects/body.tsx
```diff
@@ -64,6 +64,22 @@ export default defineField({
...
+          //   fields: [
+          //     {
+          //       name: 'concept',
+          //       type: 'reference',
+          //       title: 'Concept',
+          //       to: [{type: 'skosConcept'}],
+          //       validation: (Rule) => Rule.required(),
+          //     },
+          //   ],
+          // }),
```
mattpocock, line 82: Interested in the idea of having links to these concept pages available as a mark for the body text in exercises. This would potentially let us pull in summaries for the concept that were available on hover.

Ofc, the easier version of this would be to just have them as links and maintain the links ourselves if the slugs change.

apps/total-typescript/schemas/documents/conceptPage.ts
```diff
@@ -0,0 +1,49 @@
+import {MdOutlineArticle} from 'react-icons/md'
+
+export default {
+  name: 'conceptPage',
```
mattpocock, line 4: Looks similar to an article, but doesn't show up on the articles list and is hard-linked to a concept.

PR #954, opened 2023-05-18
upgrade typescript to 5.1.1-rc

joelhooks:
prepping for server components

build house


apps/total-typescript/src/app/concepts/page.tsx
```diff
@@ -0,0 +1,29 @@
...
+    </>
+  )
+}
+
+export default async function ConceptListPage() {
+  return (
+    <div>
+      <h1>concepts</h1>
+      <Suspense fallback={<p>Loading feed...</p>}>
+        {/* @ts-expect-error Server Component */}
```
joelhooks, line 24: waiting for TS 5.1.1 to drop officially and react typings to get updated

apps/total-typescript/src/certificate/workshop-certificate.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import {Module} from '@skillrecordings/skill-lesson/schemas/module'
 import {LockClosedIcon} from '@heroicons/react/solid'
 import Image from 'next/image'
-import {trpc} from 'trpc/trpc.client'
+import {trpc} from '@/trpc/trpc.client'
```
joelhooks, line 5: next added a line to the tsconfig that called for files inside of the `.next` dir and TS hated it, so I had to bail on our config since it decalered a `rootDir` which is what made this possible. This use of an alias is more standard.

PR #956, opened 2023-05-19
chore(pt): standardise page headline, subheadline and button links

simonswiss:
This PR also fixes a duplicated `key` on the `/buy` page since both products had the same `product.name`.

Oh, and adds the little love heart hover effect on the signature in the article pages!

Running puppy


vojtaholik (approved):
🎉

PR #965, opened 2023-05-23
Added IIMT article

mattpocock:


vojtaholik (approved):
https://www.totaltypescript.com/immediately-indexed-mapped-type

PR #967, opened 2023-05-23
Where to put your types article

mattpocock:


vojtaholik (approved):
https://www.totaltypescript.com/where-to-put-your-types-in-application-code

PR #975, opened 2023-05-26
jb/er 28 rework Lesson Progress import script to be fault tolerant

jbranchaud:
This script synchronously transfers records from a local LessonProgress
table to a remote Planetscale LessonProgress table. Because of
consistent connection failures during the multi-threaded streaming
approach, I decided to go another direction with this. With this script,
we always know how far we've processed through the table of records, so
if the connection gets closed, we can record where we are, and pick up
where we left off. This makes the script idempotent which means that it
can be run outside of the cutover processes after the site has been
brought back up. The original import script will still need to be run
first in order to load those records up into a local Docker database.

If a connection gets closed, which happens quite often with planetscale,
then this script will note where it left off and re-run itself from that
point. In this way, the script doesn't need constant babysitting to
constantly start the script when it exits early. However, if it is
failing over and over without making any progress, then the script will
notice that and exit with a message.

_How this script is intended to be used:_

1) assumes that `lesson-progress-import.ts` has been run to populate a local (docker) mysql DB (at 3309) with all the `LessonProgress` records that have been exported
2) after the rest of the cutover process has been handled and the site has been brought online, then this script can be run. The idea is that instead of incurring a bunch more downtime while this runs, we can instead restore the lesson progress data while the site is useable. There may be a few support requests, "my course progress isn't showing up!", but that is preferable to extended downtime.
3) run this script while the docker DB is connected at 3309 and a connection has been opened to planetscale KCD at 3399. It will take ~90 to 120 minutes to run

fault tolerant

### DIFF
--- a/apps/epic-react/src/data/transfer-lesson-progress-from-local-db-to-planetscale.ts
+++ b/apps/epic-react/src/data/transfer-lesson-progress-from-local-db-to-planetscale.ts
@@ -1,12 +1,22 @@
 // Running this script (Epic React):
 //
-// TODO: Update these details!!!
-// Ensure you have the `PURCHASE_DATA_IMPORT_FILE_PATH` env var (in
-// `.env.local`) set to the location of the Epic React purchase data JSON
-// export. If this file hasn't been exported yet, follow the instructions here:
-// https://github.com/skillrecordings/egghead-rails/blob/main/docs/export-kcd-products-data.md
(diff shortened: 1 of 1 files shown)


joelhooks:
one thing planetscale is good at is a lot of simultaneous connections so my approach in the past has been to just hammer them in parallel with new connections

writing this before looking at the actual code

joelhooks (approved):

PR #983, opened 2023-05-30
Added shiki-remote-plugin

mattpocock:


vojtaholik (approved):
very cool! would be awesome if we could also pass in theme options.

PR #986, opened 2023-05-31
Only call shiki-remote when necessary

mattpocock:


packages/skill-lesson/markdown/shiki-remote-plugin.ts
```diff
@@ -32,11 +33,39 @@ export interface ShikiRemotePluginOptions {
...
     await visitCodeNodes(ast, async (node) => {
       const code = node.value
 
+      if (!highlighter) {
+        highlighter = await createShikiHighlighter({
+          theme: opts.theme || 'dark-plus',
+        })
+      }
+
+      if (!node.meta?.includes('twoslash')) {
```
mattpocock, line 51: We only need to call the remote service when necessary: when 'twoslash' is specified in the code fence. Otherwise, we can just run `renderCodeToHTML` which just acts like a normal, non-powerful syntax highlighter.

vojtaholik (approved):
⭐

PR #989, opened 2023-06-01
chore(pt): tips page -> smaller thumbnails, order with newest first

simonswiss:
The auto-generated thumbnails are not that aesthetically pleasing, so I opted for a layout that de-emphasizes them and makes the content quicker to scan.

New tips ordered first, too 👍


vojtaholik (approved):
⭐

PR #992, opened 2023-06-01
fix: keep Next.js for TT at 13.4.3 for now

jbranchaud:
There is a SessionProvider error that is only reproduced on 13.4.4.

This also removes a duplicate instance of `next` as a dependency in TT's `package.json`.

image

bad patch


jbranchaud:
This was a `next-auth` version alignment issue that was fixed here: https://github.com/skillrecordings/products/commit/bc2ec3bf7cb36a58abfa1e3bb0d7c3e4b6f19c73

PR #996, opened 2023-06-06
fix(ta): certificates

vojtaholik:
gif


apps/testingaccessibility/src/pages/api/certificate/index.tsx
```diff
@@ -0,0 +1,148 @@
+import {ImageResponse} from '@vercel/og'
```
vojtaholik, line 1: I refactored the generation logic to use vercel/og™ and removed external service connection

PR #1007, opened 2023-06-09
Fixes PT-132: fix styles in popups

DrShpongle:
<details>
  <summary>Before:</summary>
  
</details>
<details>
  <summary>After:</summary>
  
</details>

Also added `Balancer` to the heading.

I couldn't reproduce the "The 'golden ticket' redeem dialog" issue described in related Linear issue.


packages/skill-lesson/video/video-overlays.tsx
```diff
@@ -403,7 +403,7 @@ const BuyProduct: React.FC<{product?: SanityProduct}> = ({product}) => {
         <h3 data-description="">
           <Balancer>
             {canViewRegionRestriction ? (
-              <div data-markdown="">
```
vojtaholik, line 406: wouldn't it be possible to use the markdown data attribute instead of adding a new one? this addition would need to be styled across all products.

apps/protailwind/src/styles/video/video-overlays.css
```diff
@@ -106,6 +106,9 @@
         }
         [data-description] {
           @apply w-full pb-10 pt-3 text-lg opacity-90;
+          [data-restriction-copy] {
+            @apply text-white;
+          }
           [data-markdown] {
             @apply prose w-full prose-p:text-gray-300;
```
vojtaholik, line 113: adding `text-gray-300` class here should work fine.

packages/skill-lesson/video/video-overlays.tsx
DrShpongle: evntually removed `data-markdown` at all, no additional styles needed.

vojtaholik (approved):

PR #1020, opened 2023-06-20
feat(epic-web): Navigate to branding on right click

kentcdodds:
Not as awesome as https://twitter.com/zenorocha/status/1670816067109924864 but still pretty great.


vojtaholik (approved):
I'll make it nicer once I finish updating the site with new design elements :)

PR #1024, opened 2023-06-21
cleanup: remove unused, duplicated purchase update

jbranchaud:
Related to my investigation of https://linear.app/skillrecordings/issue/SKS-52/valid-purchases-should-have-valid-status-not-restricted

We have this function duplicated across many of the apps, due to it showing up in the `create-skill-app` template. I'm removing it from the template and all the apps where it isn't being used. The app and package that are using this function are using the one that coms from `getSdk()` (as they should be).

This is some good clean up to do and greatly reduces the number of possible places that a `purchase` update can be happening.

clean up


vojtaholik (approved):
LGTM

PR #1040, opened 2023-07-05
feat(ew): spike product detail pages

vojtaholik:
screenshot

gif


vojtaholik:
I think I'd like to try to consolidate some of the data loading logic into a new trpc route specifically targeted for product detail pagea because it's becoming quite messy. any thoughts on this @joelhooks?

vojtaholik:
I think I somehow managed to completely bork the package-lock file, heh. Completely removing and reinstalling doesn't help either.

PR #1065, opened 2023-07-20
feat(commerce): spike upgrade for team seats

joelhooks:
pretty quick dubious quality spike to explore dynamic bulk purchase lookup

this was trickier than expected because of the way we recharge bulk coupons and we do NOT record quantity with individual purchases which would make this a lot more straightforward.

That said we can just load the session too. Might be worth updating the model so we have that data on our side alongside the purchase price?

upgrades

### DIFF
--- a/packages/commerce-server/src/format-prices-for-product.ts
+++ b/packages/commerce-server/src/format-prices-for-product.ts
@@ -5,6 +5,7 @@ import {Context, defaultContext, getSdk} from '@skillrecordings/database'
 import {FormattedPrice} from './@types'
 import {Purchase} from '@prisma/client'
 import {getStripeSdk} from '@skillrecordings/stripe-sdk'
+import * as console from 'console'
 
 // 10% premium for an upgrade
 // TODO: Display Coupon Errors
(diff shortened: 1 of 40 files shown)


packages/commerce-server/src/format-prices-for-product.ts
```diff
@@ -31,6 +33,60 @@ type FormatPricesForProductOptions = {
...
+          merchantSession: true,
+        },
+      })) as any
+    if (parentPurchase?.merchantSession?.identifier) {
+      const charge = await getCheckoutSession(
+        parentPurchase.merchantSession.identifier,
+      )
+      if (charge.line_items) {
+        const {quantity, amount_total} = charge.line_items.data[0]
+        unitPrice = quantity ? amount_total / 100 / quantity : 0
```
jbranchaud, line 81: Checking my understanding -- the way the _true_ branch of this ternary is working is that the `amount_total` might be something like `236000` and the `quantity = 4`, so then you do `236000 / 100 / 4 => 590` which tells you that each seat cost `590` (hence _unit price_).

And, then for a `total_amount` that involved a bulk discount, you'd have something like `531,000` (which comes from `(590 * 10 * 0.9) * 100`). So then the unit price you'd get here would be the bulk discounted unit price of `531`.

packages/commerce-server/src/format-prices-for-product.ts
```diff
@@ -31,6 +34,25 @@ type FormatPricesForProductOptions = {
...
+  if (upgradeProductId) {
+    const price = await getPrice({
+      where: {
+        productId: upgradeProductId,
+      },
+    })
+    return price?.unitAmount.toNumber() || 0
+  }
+  return 0
+}
```
joelhooks, line 54: this is heavily simplified this way

packages/commerce-server/src/format-prices-for-product.ts
```diff
@@ -137,7 +198,7 @@ export async function formatPricesForProduct(
     calculatedPrice: getCalculatedPriced({
       unitPrice: price.unitAmount.toNumber(),
       ...(upgradeFromPurchase && {
-        fixedDiscount: upgradeFromPurchase.totalAmount.toNumber(),
+        fixedDiscount: fixedDiscountForUpgrade,
       }),
```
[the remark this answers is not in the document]
joelhooks: I think we can defer the bulk upgrades. It's more complex for sure.

PR #1068, opened 2023-07-21
jbranchaud/tjs 39 dont run up epic webs stripe webhook error rate

jbranchaud:
- feat(error-reporting): move Sentry config to env
- feat: return 200 for unrecognized stripeProductId

This should address the issue where we are running up the error rate in Epic Web stripe because of unrecognized product ID instances from TJS/ER purchases.

sentry


packages/skill-api/src/core/services/process-stripe-webhook.ts
```diff
@@ -129,10 +134,22 @@ export async function processStripeWebhooks({
         }
       }
     } catch (err: any) {
-      console.error(err)
-      return {
-        status: 400,
-        body: `Webhook Error: ${err.message}`,
+      // To avoid making too much noise in Stripe, we are going to report a
+      // instance of No Associated Product to Sentry and then return a 200
+      // response from here.
```
joelhooks, line 139: this comment is a lie 🍰

packages/skill-api/src/core/services/process-stripe-webhook.ts
```diff
@@ -129,10 +134,22 @@ export async function processStripeWebhooks({
...
+      if (err.message === NO_ASSOCIATED_PRODUCT) {
+        // Sentry.captureException(err)
```
joelhooks, line 141: Do we want this to be enabled? Most of the PR is setting it up!

jbranchaud: just noticed that as well and already committed it back in

packages/skill-api/src/core/services/process-stripe-webhook.ts
jbranchaud: @joelhooks yeah, I originally thought the issue was just creating noise in stripe, but the truth is that too high of an error rate and Stripe kills your webhook, right?

joelhooks: I meant the Sentry bit because of the comment and was making a Portal joke 🤡

joelhooks (approved):

packages/skill-api/src/core/services/process-stripe-webhook.ts
jbranchaud: whoops, right over my head. Portal joke now received and appreciated 🙌

PR #1071, opened 2023-07-24
types: add type combining Product and Prices

jbranchaud:
Throwing my hat into the ring of making adjustments to the `FormattedPrice` type

throw hat into ring


joelhooks (approved):
I like it, thanks!

PR #1073, opened 2023-07-24
feat: skip CK tagging in development

jbranchaud:
I was getting this error about the CK purchase tag ID needing to be set. I don't think we want to be sending tagging events to CK in development, so I'm checking the `NODE_ENV` and exiting this function early if we are in `development`.

CleanShot 2023-07-24 at 11 48 03@2x

### DIFF
--- a/packages/skill-api/src/core/services/convertkit/convertkit-tag-purchase.ts
+++ b/packages/skill-api/src/core/services/convertkit/convertkit-tag-purchase.ts
@@ -41,6 +41,9 @@ async function getSanityProduct(productId: string) {
 }
 
 export async function convertkitTagPurchase(email: string, purchase: Purchase) {
+  // skip CK tagging in development
+  if (process.env.NODE_ENV === 'development') return
+
   try {
     const sanityProduct = await getSanityProduct(purchase.productId)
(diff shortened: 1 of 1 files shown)


joelhooks:
One thing is we want this to raise a flag in prod because it's easy to forget and a pita when it does

packages/skill-api/src/core/services/convertkit/convertkit-tag-purchase.ts
```diff
@@ -41,6 +41,9 @@ async function getSanityProduct(productId: string) {
 }
 
 export async function convertkitTagPurchase(email: string, purchase: Purchase) {
+  // skip CK tagging in development
+  if (process.env.NODE_ENV === 'development') return
```
joelhooks, line 45:
```suggestion
  if (process.env.SKIP_CK_TAGGING === 'true') return
```

I'd prefer if this was explicit because the broad strokes makes for weird debugging sometimes

jbranchaud:
@joelhooks I applied your change. Makes sense to do it more granularly like you said.

joelhooks (approved):

PR #1074, opened 2023-07-25
feat(commerce): PPP coupon not available for non-ppp purchasers

vojtaholik:
if regional pricing criteria are met, then PPP coupon is available for users:
- with no purchase
- that previously made a purchase using PPP coupon

I also tested a different solution which only disabled PPP for upgradeProduct (in this case a react bundle), but then the standalone product would be less expensive with PPP then the upgrade itself. So I decided to turn it off on all tiers.

screenshots

gif

### DIFF
--- a/packages/commerce-server/src/format-prices-for-product.ts
+++ b/packages/commerce-server/src/format-prices-for-product.ts
@@ -81,6 +81,7 @@ export async function formatPricesForProduct(
     couponForIdOrCode,
     getPrice,
     getPurchase,
+    getPurchasesForUser,
   } = getSdk({ctx})
 
   const upgradeFromPurchase = upgradeFromPurchaseId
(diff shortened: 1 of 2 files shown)


packages/commerce-server/src/format-prices-for-product.ts
```diff
@@ -166,8 +173,17 @@ export async function formatPricesForProduct(
     ? appliedMerchantCoupon.percentageDiscount.toNumber() < bulkCouponPercent
     : true
 
-  const pppAvailable =
-    quantity === 1 && pppDiscountPercent > 0 && appliedMerchantCouponLessThanPPP
```
jbranchaud, line 170: the change looks good, my only suggestion would be to keep a secondary variable like `pppConditionsMet` with these three conditions and then that can be used below:

```javascript
const pppAvailable = userPurchases.length > 0
  ? hasPurchaseWithPPP && pppConditionsMet
  : pppConditionsMet
```

thoughts?

jbranchaud (approved):

PR #1080, opened 2023-07-25
jbranchaud/tjs 39 dont run up epic webs stripe webhook error rate 3

jbranchaud:
This PR does two things:

1. returns 200 responses to Stripe when we don't recognize the incoming Stripe Product ID
2. reports the unrecognized Product ID to Sentry so that we have a paper trail of it happening that we can investigate

Note: currently investigating in staging why these changes soured in production and caused 500s for TT

sour


jbranchaud:
Closing this since we got the primary part of this fix in the other week and the Sentry stuff can come later.

PR #1081, opened 2023-07-25
corrects incorrect all region upgrade pricing

joelhooks:
regional


packages/skill-api/src/core/services/stripe-checkout.ts
```diff
@@ -219,6 +218,7 @@ export async function stripeCheckout({
           coupon: coupon.id,
         })
       } else if (merchantCoupon && merchantCoupon.identifier) {
+        console.log('BLAH')
```
joelhooks, line 221:
```suggestion
```

jbranchaud (approved):
🚀

PR #1086, opened 2023-07-26
add logic branch to price calculation

joelhooks:
I added a logic branch for PPP that looks to see if we are upgrading to a higher tier and **automatically applies PPP coupon** instead of requiring user intervention for this

### DIFF
--- a/packages/commerce-server/src/format-prices-for-product.ts
+++ b/packages/commerce-server/src/format-prices-for-product.ts
@@ -42,18 +42,14 @@ export async function getFixedDiscountForUpgrade({
   upgradeFromPurchase?: Purchase
   ctx?: Context
 }) {
-  if (upgradeFromPurchase?.status === 'Restricted') {
-    return upgradeFromPurchase?.totalAmount.toNumber() || 0
-  } else {
-    const {getPrice} = getSdk({ctx})
-    if (upgradeProductId) {
-      const price = await getPrice({
-        where: {
-          productId: upgradeProductId,
-        },
-      })
-      return price?.unitAmount.toNumber() || 0
-    }
+  const {getPrice} = getSdk({ctx})
+  if (upgradeProductId) {
+    const price = await getPrice({
+      where: {
+        productId: upgradeProductId,
+      },
+    })
+    return price?.unitAmount.toNumber() || 0
   }
   return 0
 }
@@ -132,10 +128,6 @@ export async function formatPricesForProduct(
 
   const userPurchases = await getPurchasesForUser(userId)
 
-  const hasPurchaseWithPPP = userPurchases.find(
-    (purchase) => purchase.status === 'Restricted',
-  )
-
   const product = await getProduct({
     where: {id: productId},
     include: {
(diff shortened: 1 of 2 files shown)


packages/commerce-server/src/format-prices-for-product.ts
```diff
@@ -184,12 +176,18 @@ export async function formatPricesForProduct(
   const pppConditionsMet =
     quantity === 1 && pppDiscountPercent > 0 && appliedMerchantCouponLessThanPPP
 
+  const hasPurchaseWithPPP = userPurchases.some(
+    (purchase) =>
+      purchase.status === 'Restricted' &&
+      upgradeFromPurchase &&
+      purchase.productId === upgradeFromPurchase?.productId,
+  )
```
joelhooks, line 184: this should really only be relevant if they are upgrading a product with PPP restrictions, but we could automatically apply it across the board if they have any PPP too. 

That would be making the assumption that they would want PPP applied to any/all purchases, which I feel is safe?

If that's the case we could remove the product id comparison.

packages/commerce-server/src/format-prices-for-product.ts
```diff
@@ -298,6 +296,35 @@ export async function formatPricesForProduct(
...
+          fixedDiscount: fixedDiscountForUpgrade,
+        }),
+      }),
+      appliedMerchantCoupon: merchantCouponWithoutIdentifier,
+      ...(upgradeFromPurchase && {
+        upgradeFromPurchaseId,
+        upgradeFromPurchase,
+        upgradedProduct,
+      }),
+    }
```
joelhooks, line 327: if they have any restricted/ppp purchase and are upgrading, just apply PPP instead of giving them a branching choice here. It's a safe assumption.

This probably leaves a hole where they want to upgrade to the unrestricted version to remove regional restriction, which DOES happen, but is rarer.

packages/skill-lesson/path-to-purchase/pricing.tsx
```diff
@@ -109,11 +109,14 @@ export const Pricing: React.FC<React.PropsWithChildren<PricingProps>> = ({
...
 
-  const isRestrictedUpgrade = purchaseToUpgrade?.status === 'Restricted'
-
   const defaultCoupon = formattedPrice?.defaultCoupon
   const appliedMerchantCoupon = formattedPrice?.appliedMerchantCoupon
 
+  const isRestrictedUpgrade =
+    purchaseToUpgrade?.status === 'Restricted' &&
+    appliedMerchantCoupon &&
+    appliedMerchantCoupon.type !== 'ppp'
```
joelhooks, line 118: this is for the "remove restricted access" scenario, which isn't in this PR yet because still noodling...

joelhooks (commented):
this definitely needs tests.

packages/commerce-server/src/format-prices-for-product.ts
```diff
@@ -181,15 +172,28 @@ export async function formatPricesForProduct(
...
+  const hasNonPPPPurchases = userPurchases.some(
+    (purchase) => purchase.status === 'Valid',
+  )
+
   const pppConditionsMet =
-    quantity === 1 && pppDiscountPercent > 0 && appliedMerchantCouponLessThanPPP
+    pppDiscountPercent > 0 &&
+    quantity === 1 &&
+    !hasNonPPPPurchases &&
+    appliedMerchantCouponLessThanPPP
```
joelhooks, line 183: if they've made any non-PPP purchases, they don't get PPP which seems fair, mixing and matching makes life hard and PPP is meant as a courtesy

PR #1087, opened 2023-07-26
Jb/fix bundled pricing display

jbranchaud:
There is a pricing display bug where without a site-wide coupon, we show the upgrade price along with the discount banner slashing out that same upgrade price. This is because of the logic that determines when certain pricing is a discount. It wasn't accounting for this kind of upgrade pricing.

This PR introduces a `fullPrice` concept in the `FormattedPrice` object which simplifies logic and removes the need to recompute values that were already determined in the commerce-server.

_Note: this doesn't eliminate all the edges in the pricing display logic. For instance, things are still displaying weird when you toggle over to a bulk discounted upgrade._

## Before

CleanShot 2023-07-26 at 13 52 06@2x

## After

CleanShot 2023-07-26 at 15 53 41@2x

upgrade

### DIFF
--- a/packages/commerce-server/src/format-prices-for-product.ts
+++ b/packages/commerce-server/src/format-prices-for-product.ts
@@ -198,10 +198,14 @@ export async function formatPricesForProduct(
   const bulkDiscountAvailable =
     bulkCouponPercent > 0 && appliedMerchantCouponLessThanBulk && !pppApplied
 
+  const unitPrice: number = price.unitAmount.toNumber()
+  const fullPrice: number = unitPrice * quantity - fixedDiscountForUpgrade
+
   let defaultPriceProduct: FormattedPrice = {
     ...product,
     quantity,
-    unitPrice: price.unitAmount.toNumber(),
+    unitPrice,
+    fullPrice,
     calculatedPrice: getCalculatedPriced({
       unitPrice: price.unitAmount.toNumber(),
       ...(upgradeFromPurchase && {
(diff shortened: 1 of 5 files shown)


packages/commerce-server/src/format-prices-for-product.ts
```diff
@@ -194,10 +194,14 @@ export async function formatPricesForProduct(
...
 
+  const unitPrice: number = price.unitAmount.toNumber()
+  const fullPrice: number = unitPrice * quantity - fixedDiscountForUpgrade
+
   let defaultPriceProduct: FormattedPrice = {
     ...product,
     quantity,
-    unitPrice: price.unitAmount.toNumber(),
+    unitPrice,
+    fullPrice,
```
jbranchaud, line 204: Compute `fullPrice` in the commerce-server, so that anyone getting a `FormattedPrice` object can make use of that value already computed server-side and not have to recompute these values.

Open to suggestions on the name, `fullPrice` seemed to be how downstream components were thinking of it, but there may be better options for what to call this.

packages/skill-lesson/path-to-purchase/pricing-check-context.tsx
```diff
@@ -75,7 +75,7 @@ export const PriceCheckProvider: React.FC<React.PropsWithChildren<any>> = ({
     if (!price) {
       return false
     }
-    return price.unitPrice > price.calculatedPrice
+    return price.fullPrice > price.calculatedPrice
```
jbranchaud, line 78: This fixes the `isDiscount` check for upgrade pricing.

joelhooks (approved):

jbranchaud:
Added a couple tests, moving this out of `NO MERGE`

PR #1089, opened 2023-07-27
feat: update .env files, add SKIP_CK_TAGGING

jbranchaud:
We recently added the `SKIP_CK_TAGGING` flag (https://github.com/skillrecordings/products/pull/1073) so that development can opt-in and out of subscribing and tagging a new user in CK.

I noticed that the email addresses I used for several recent local TT test checkout sessions were being included in real/prod CK email sequences. I believe that is because the CK tagging was still active in the post-checkout flow running locally. **I think it is a good default to skip this tagging with the flag.**

Anyone needing to enable it or debug something related to it can easily turn it back on in their own environment.

CleanShot 2023-07-27 at 15 35 45@2x

skip


joelhooks (approved):
i think this could go a step further and skip adding to ck at all if it doesnt

PR #1093, opened 2023-08-01
jb/refactor coupon application in pricing logic

jbranchaud:
_Update: Loom Walkthrough https://www.loom.com/share/ab73c4acfb0a479eabde80d776926d6c_

This PR represents a refactoring of the Price Formatter logic to make it easier to reason about and easier to make updates as pricing logic, coupon logic, and product offerings continue to evolve.

There was, for instance, lots of redundant logic scattered throughout the price formatter that makes it more error-prone to update because you have to know to update multiple spots.

There were conditional blocks that appear to be added independent of similar logic paths because it was hard to reason about in what ways they overlapped and were different.

This PR took the approach of moving _coupon determining_ logic into a separate module to be called by the price formatter. It separates out the logic for PPP, bulk, and special coupons and then funnels those results back to the price formatter in a way that means there is a single path through the formatter rather than hundreds of lines of `if/else` logic.

It's a little hard to demonstrate with the diff alone, but the core part of the Price Formatter module goes from ~300 lines to about 40.

CleanShot 2023-08-01 at 15 45 05@2x

The PR is comprised of the following series of commits. I took a very incremental approach, moving logic around primarily, with very little actual changes to the logic. Once these changes are reviewed and approved, these should be squashed before merging to make it trivial to roll this back. That said, I've done extensive manual testing against Total TypeScript locally, through all kinds of scenarios to check the logic and it is sound.

- feat(commerce): move pppApplied into helper
- feat(commerce): move hasPPP check to helper
- feat(commerce): return result instead of throwing
- feat(commerce): move more PPP logic into helper
- feat(commerce): consolidate PPP logic
- cleanup(commerce): remove user purchases lookup
- feat(commerce): combine PPP formatting scenarios
- feat(commerce): move bulk coupon logic into helper
- feat(commerce): try renaming incoming coupon
- feat(commerce): make coupon comparison more explicit
- feat(commerce): find available coupons in helper
- feat(commerce): lookup bulk coupon in helper
- cleanup(commerce): remove unused PPP percent const
- feat(commerce): apply bulk, combine conditions
- feat(commerce): tighten fixed discount logic
- feat(commerce): pull upgradeDetails logic up
- feat(commerce): collapse bulk logic into rest
- feat(skill-api): fix params to get fixed discount
- feat(commerce): collapse available PPP logic
- feat(commerce): collapse extraneous 'special' logic
- feat(commerce): remove commented error-throwing
- cleanup(commerce): clean up some comments
- cleanup(commerce): more coupon comment cleanup
- feat(commerce): reduce PPP details logic
- feat(commerce): remove unused hasPurchaseWithPPP
- feat(commerce): note future refactoring
- feat(commerce): get bulk coupon, instead of details
- feat(commerce): no need to return pppDetails
- feat(commerce): remove coupon `code` logic branch
- feat(commerce): remove last redundant price block
- fix(commerce): don't lookup coupon w/ undefined
- fix(commerce): ensure PPP does not leak through

simpler

### DIFF
--- a/packages/commerce-server/src/@types/index.ts
+++ b/packages/commerce-server/src/@types/index.ts
@@ -8,7 +8,7 @@ import type {PortableTextBlock} from '@portabletext/types'
 
 type MerchantCouponWithCountry = MerchantCoupon & {country?: string | undefined}
 
-type MinimalMerchantCoupon = Omit<
+export type MinimalMerchantCoupon = Omit<
   MerchantCouponWithCountry,
   'identifier' | 'merchantAccountId'
 >
--- a/packages/commerce-server/src/determine-coupon-to-apply.ts
+++ b/packages/commerce-server/src/determine-coupon-to-apply.ts
@@ -65,7 +65,7 @@ export const determineCouponToApply = async (
     prismaCtx,
   })
 
-  const bulkDiscountDetails = await getBulkCouponDetails({
+  const bulkCouponToBeApplied = await getBulkCouponDetails({
     prismaCtx,
     userId,
     productId,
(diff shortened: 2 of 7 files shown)


joelhooks (commented):
😮

jbranchaud:
> 😮

There is a lot happening, so here is a Loom walkthrough https://www.loom.com/share/ab73c4acfb0a479eabde80d776926d6c \cc @joelhooks

joelhooks (approved):

PR #1102, opened 2023-08-02
fix(commerce): fix how fixed discount is computed

jbranchaud:
There was an issue with how we were computing the fixed discount for upgrades, where we were differentiating between a "Core Product to Bundled Product" upgrade and a PPP (Region Restriction) Upgrade.

Irrespective of a site-wide discount, we were displaying a $0 price for a PPP Upgrade:

### Example 1 (without site-wide discount)

CleanShot 2023-08-02 at 09 37 32@2x

### Example 2 (with site-wide discount)

CleanShot 2023-08-02 at 09 36 09@2x

---

## The Fix

The logic for computing the fixed discount has now been updated to account for the difference between those two types of upgrades:

### PPP Upgrade (originally purchased at $276.50, now paying $513.50—difference of $790—to remove Region Restriction)

CleanShot 2023-08-02 at 11 51 53@2x

### PPP Upgrade even works with site-wide discount in effect

CleanShot 2023-08-02 at 12 02 07@2x

### Fixed Discount for Bundle Upgrades works as expected as well

CleanShot 2023-08-02 at 11 56 21@2x

super moon


packages/commerce-server/src/format-prices-for-product.ts
```diff
@@ -33,23 +34,52 @@ type FormatPricesForProductOptions = {
...
-  upgradeProductId,
-  upgradeFromPurchase,
+  purchaseToBeUpgraded,
+  productToBePurchased,
   ctx = defaultContext,
 }: {
-  upgradeProductId: string
-  upgradeFromPurchase?: Purchase
+  purchaseToBeUpgraded: Purchase | null
+  productToBePurchased: Product
```
jbranchaud, line 42: I reworked the incoming params to make the naming more explicit and to make it so that we can compare the Purchase To Be Upgraded with the Product To Be Purchased.

packages/commerce-server/src/format-prices-for-product.ts
```diff
@@ -103,7 +133,7 @@ export async function formatPricesForProduct(
           status: true,
         },
       })
-    : false
+    : null
```
jbranchaud, line 136: This makes the types easier -- `Purchase | null` instead of `Purchase | null | false`

packages/commerce-server/src/format-prices-for-product.ts
```diff
@@ -138,6 +158,14 @@ export async function formatPricesForProduct(
     throw new PriceFormattingError(`no-product-found`, noContextOptions)
   }
 
+  const fixedDiscountForUpgrade = await getFixedDiscountForUpgrade({
+    purchaseToBeUpgraded: upgradeFromPurchase,
+    productToBePurchased: product,
+    ctx,
+  })
```
jbranchaud, line 165: Update the call site

packages/skill-api/src/core/services/stripe-checkout.ts
```diff
@@ -180,12 +180,10 @@ export async function stripeCheckout({
...
-        const fixedDiscountForUpgrade = upgradeFromPurchase
-          ? await getFixedDiscountForUpgrade({
-              upgradeProductId: upgradeFromPurchase.productId,
-              upgradeFromPurchase,
-            })
-          : 0
+        const fixedDiscountForUpgrade = await getFixedDiscountForUpgrade({
+          purchaseToBeUpgraded: upgradeFromPurchase,
+          productToBePurchased: loadedProduct,
+        })
```
jbranchaud, line 186: Update the one other call site

packages/commerce-server/src/format-prices-for-product.test.ts
```diff
@@ -429,6 +430,14 @@ function mockDefaultAndUpgradeProduct() {
     mockPurchaseToBeUpgraded,
   )
 
+  mockCtx.prisma.upgradableProducts.findMany.mockResolvedValue([
+    {
+      // @ts-ignore
+      upgradableTo: {id: mockUpgradeProduct.id, name: mockUpgradeProduct.name},
+      upgradableFrom: {id: mockProduct.id, name: mockProduct.name},
+    },
+  ])
```
jbranchaud, line 439: This test setup now needs a mocked lookup of the `UpgradableProducts`

joelhooks (approved):

PR #1104, opened 2023-08-04
feat(skill-lesson): use authed userId for pricing

jbranchaud:
Rather than relying on the client-side call sites to pass in a `userId`,
we should just use the `userId` on the NextAuth JWT from the
server-side.

next auth


packages/skill-lesson/trpc/routers/pricing.ts
```diff
@@ -132,9 +130,11 @@ export const pricing = router({
 
       const token = await getToken({req: ctx.req})
 
+      const verifiedUserId = token?.sub
```
jbranchaud, line 133: I think it is generally going to be safer (and more accurate) to grab the `userId` off the authed JWT within the server-side code rather than relying on the client to pass in the relevant `userId`.

vojtaholik (approved):

PR #1107, opened 2023-08-08
fix(tt): pass applied bulk coupon to Stripe

jbranchaud:
This commit also includes a refactor of how we are building the Skill
API stripe checkout path for form submission. It consolidates the
parameter building, adds type-checking, and uses URLSearchParams.

Example of the bulk coupon not being passed to Stripe:

image

Now, it is being passed in:

CleanShot 2023-08-08 at 11 12 52@2x

bulk up

### DIFF
--- a/apps/total-typescript/src/templates/purchased-product-template.tsx
+++ b/apps/total-typescript/src/templates/purchased-product-template.tsx
@@ -27,6 +27,7 @@ import {track} from '@skillrecordings/skill-lesson/utils/analytics'
 import {usePriceCheck} from '@skillrecordings/skill-lesson/path-to-purchase/pricing-check-context'
 import {PriceDisplay} from '@skillrecordings/skill-lesson/path-to-purchase/pricing'
 import {QueryStatus} from '@tanstack/react-query'
+import {buildStripeCheckoutPath} from '@skillrecordings/skill-lesson/utils/build-stripe-checkout-path'
 
 const PurchasedProductTemplate: React.FC<ProductPageProps> = ({
   purchases = [],
@@ -195,14 +196,15 @@ const Upgrade: React.FC<{
   purchaseToUpgrade: any
   formattedPrice: FormattedPrice | undefined
   formattedPriceStatus: QueryStatus
-}> = ({
-  purchaseToUpgrade,
-  formattedPrice,
-  product,
-  userId,
-  formattedPriceStatus,
-}) => {
-  const appliedMerchantCoupon = formattedPrice?.appliedMerchantCoupon
+}> = ({formattedPrice, userId, formattedPriceStatus}) => {
+  const formActionPath = buildStripeCheckoutPath({
+    userId,
+    quantity: formattedPrice?.quantity,
+    productId: formattedPrice?.id,
+    bulk: false,
+    couponId: formattedPrice?.appliedMerchantCoupon?.id,
+    upgradeFromPurchaseId: formattedPrice?.upgradeFromPurchaseId,
+  })
 
   return (
     <div>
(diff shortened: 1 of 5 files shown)


packages/skill-lesson/team/buy-more-seats.tsx
```diff
@@ -41,10 +26,12 @@ const BuyMoreSeats = ({
       quantity: debouncedQuantity,
     })
 
-  const formActionPath = buildFormActionPath({
+  const formActionPath = buildStripeCheckoutPath({
     userId,
     quantity: debouncedQuantity,
     productId,
+    bulk: true,
+    couponId: formattedPrice?.appliedMerchantCoupon?.id,
```
jbranchaud, line 34: These are the values that were missing, that caused the coupon to not get propagated to Stripe

packages/skill-lesson/utils/build-stripe-checkout-path.ts
```diff
@@ -0,0 +1,34 @@
...
+  })
+type Params = z.input<typeof ParamsSchema>
+
+export const buildStripeCheckoutPath = (params: Params) => {
+  const result = ParamsSchema.safeParse(params)
+  if (result.success) {
+    const queryParams = omitBy(result.data, isNil)
+    const queryParamString = new URLSearchParams(queryParams).toString()
+    return `/api/skill/checkout/stripe?${queryParamString}`
+  }
```
jbranchaud, line 30: This is a helper that we can use across apps to build the skill-api path for a Stripe checkout. It provides types and uses `URLSearchParams` to build the query param string. This means each app doesn't have to do a bunch of manual string interpolation and can instead just pass in a type-checked object.

packages/skill-lesson/path-to-purchase/pricing.tsx
```diff
@@ -269,15 +270,16 @@ export const Pricing: React.FC<React.PropsWithChildren<PricingProps>> = ({
...
-                  action={`/api/skill/checkout/stripe?productId=${
-                    formattedPrice?.id
-                  }&couponId=${appliedMerchantCoupon?.id}&bulk=${
-                    isBuyingForTeam ? 'true' : 'false'
-                  }&quantity=${quantity}${userId ? `&userId=${userId}` : ``}${
-                    formattedPrice?.upgradeFromPurchaseId
-                      ? `&upgradeFromPurchaseId=${formattedPrice?.upgradeFromPurchaseId}`
-                      : ``
-                  }${cancelUrl ? `&cancelUrl=${cancelUrl}` : ``}`}
+                  action={buildStripeCheckoutPath({
```
jbranchaud, line 273: Updating one of the other places in TT where this stripe checkout path is built

apps/total-typescript/src/templates/purchased-product-template.tsx
```diff
@@ -195,14 +196,15 @@ const Upgrade: React.FC<{
...
-}> = ({
-  purchaseToUpgrade,
-  formattedPrice,
-  product,
-  userId,
-  formattedPriceStatus,
-}) => {
-  const appliedMerchantCoupon = formattedPrice?.appliedMerchantCoupon
+}> = ({formattedPrice, userId, formattedPriceStatus}) => {
+  const formActionPath = buildStripeCheckoutPath({
```
jbranchaud, line 200: Updating one of the other places in TT where this stripe checkout path is built

packages/skill-lesson/utils/build-stripe-checkout-path.ts
joelhooks: would love to see a test for this one :)

joelhooks (approved):

PR #1109, opened 2023-08-08
Add tests to stripe checkout path builder, use builder across other apps

jbranchaud:
- feat(skill-lesson): add tests for path builder
- feat: use stripe checkout path builder everywhere
- feat: add missing applied merchant coupons

path


joelhooks (approved):
🔥

PR #1110, opened 2023-08-08
feat(commerce): indicate bulk purchase in price

jbranchaud:
The FormattedPrice should indicate to consumers whether the purchase is
going to be a bulk purchase or not. This is because the price formatting
logic does a qualified seat count lookup and knows if the purchase is
individual or adding to an existing bulk purchase.

Existing logic might think a purchase of `quanity = 1` is an individual
purchase when in fact a single seat is being added to an existing bulk
purchase.

captain planet

### DIFF
--- a/apps/testingaccessibility/src/components/pricing.tsx
+++ b/apps/testingaccessibility/src/components/pricing.tsx
@@ -196,7 +196,7 @@ export const Pricing: React.FC<React.PropsWithChildren<PricingProps>> = ({
               userId,
               quantity,
               productId: formattedPrice?.id,
-              bulk: quantity > 1,
+              bulk: Boolean(formattedPrice?.bulk),
               couponId: appliedMerchantCoupon?.id,
               upgradeFromPurchaseId: formattedPrice?.upgradeFromPurchaseId,
             })}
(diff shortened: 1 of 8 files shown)


apps/testingaccessibility/src/components/pricing.tsx
```diff
@@ -196,7 +196,7 @@ export const Pricing: React.FC<React.PropsWithChildren<PricingProps>> = ({
               userId,
               quantity,
               productId: formattedPrice?.id,
-              bulk: quantity > 1,
+              bulk: Boolean(formattedPrice?.bulk),
```
jbranchaud, line 199: There are several places where we are doing this `quantity > 1` check which can be inaccurate.

packages/commerce-server/src/determine-coupon-to-apply.ts
```diff
@@ -297,6 +298,7 @@ const getBulkCouponDetails = async (params: GetBulkCouponDetailsParams) => {
     newPurchaseQuantity: quantity,
     prismaCtx,
   })
+  const consideredBulk = seatCount > 1
```
jbranchaud, line 301: Here we determine if the purchase will be considered bulk based on the qualified seat count

joelhooks (approved):

PR #1112, opened 2023-08-11
jbranchaud/tt 335 ppp purchaser gets unrestricted access when upgrading to

jbranchaud:
There is a bug in the post-purchase logic for an upgrade from Core to Bundle PPP purchase where we don't recognize that the purchase should be region restricted and end up marking it as Valid even though the user paid a PPP price.

This PR fixes it by including the PPP Coupon Stripe Identifier in the checkout metadata so that on the other end we are able to know post purchase whether PPP was applied or not. Typically this hasn't been an issue, but when we combine the PPP discount **with** the Fixed Discount for Upgrade, we lose track the PPP coupon ID.

- fix(tt): upgrades with PPP should be restricted
- feat(commerce): use Stripe ID instead of boolean

stripe

### DIFF
--- a/packages/commerce-server/src/record-new-purchase.ts
+++ b/packages/commerce-server/src/record-new-purchase.ts
@@ -139,6 +139,7 @@ export async function recordNewPurchase(checkoutSessionId: string): Promise<{
     quantity,
     bulk: metadata?.bulk === 'true',
     country: metadata?.country,
+    pppApplied: metadata?.pppApplied === 'true',
     checkoutSessionId,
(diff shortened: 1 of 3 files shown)


joelhooks (approved):
figure out test coverage on stripe-checkout would be a big win imo

jbranchaud:
> figure out test coverage on stripe-checkout would be a big win imo

I was thinking the same. I suspect it is going to look a lot like the tests for `formatPriceForProduct`.

PR #1117, opened 2023-08-15
jbranchaud/tt 334 ppp box should be checked for bundle upgrade w ppp pricing

jbranchaud:
This PR adds an option to `formatPricesForProduct` to control whether a qualifying PPP coupon should be auto-applied to the upgrade of an existing purchase.

- types(tt): clean up flow of types for PPP Pricing
- feat(commerce): get pricing w/o PPP auto-applied

Scenario: Let's say a user has purchased Core(PPP). When we are computing the pricing for the upgrade to the Bundle, we want to auto-apply the PPP discount to that upgrade as long as they are in the same region. This is a great default, but we need it to be configurable in case the user wants to toggle-off PPP pricing and make an unrestricted purchase.

Adding this option allows the UI to request pricing with and without the PPP discount being auto-applied.

### Before

image

### After

https://github.com/skillrecordings/products/assets/694063/ce65d476-7c13-4cb1-a515-72a6712147af

auto

### DIFF
--- a/packages/commerce-server/src/determine-coupon-to-apply.ts
+++ b/packages/commerce-server/src/determine-coupon-to-apply.ts
@@ -16,6 +16,7 @@ const DetermineCouponToApplyParamsSchema = z.object({
   userId: z.string().optional(),
   productId: z.string(),
   purchaseToBeUpgraded: PurchaseSchema.nullable(),
+  autoApplyPPP: z.boolean(),
 })
 
 type DetermineCouponToApplyParams = z.infer<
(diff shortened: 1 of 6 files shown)


joelhooks:
I think the box should just be removed entirely in this scenario.

jbranchaud:
> I think the box should just be removed entirely in this scenario.

we should allow them to upgrade to an All Region purchase in this scenario if they'd like, right? Without the checkbox, we are only making PPP an option.

joelhooks:
My assumption is that in 95% of cases folks want PPP and the extra option leads to more support requests than the inverse.

jbranchaud:
Closing in favor of #1142

PR #1122, opened 2023-08-18
jbranchaud/tt 340 fix pricing confusion for team seat upgrades 2

jbranchaud:
- fix(commerce): missing individual upgrade

> Because I was checking the `upgradeFrom` instead of the `upgradeTo` when
filtering out already-purchased products, we were missing the fixed
discount for an individual Core to Bundle upgrade.

- feat(commerce): no fixed discount for bulk

> We may at some point have a fixed discount for bulk pricing when that
part of the commerce flow is further developed, but for now the fixed
discount we are computing is for individual purchases and so this now
differentiates for that.

whoops

### DIFF
--- a/packages/commerce-server/src/format-prices-for-product.ts
+++ b/packages/commerce-server/src/format-prices-for-product.ts
@@ -31,7 +31,7 @@ type FormatPricesForProductOptions = {
   userId?: string
 }
 
-export async function getFixedDiscountForUpgrade({
+export async function getFixedDiscountForIndividualUpgrade({
   purchaseToBeUpgraded,
   productToBePurchased,
   purchaseWillBeRestricted,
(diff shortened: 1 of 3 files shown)


packages/skill-lesson/trpc/routers/pricing.ts
```diff
@@ -43,24 +43,30 @@ const checkForAnyAvailableUpgrades = async ({
...
+  // filter out potential upgrades that have already been purchased
+  const availableUpgrades = potentialUpgrades.filter<AvailableUpgrade>(
+    (
+      availableUpgrade: AvailableUpgrade,
+    ): availableUpgrade is AvailableUpgrade => {
+      return !productIdsAlreadyPurchased.includes(
+        availableUpgrade.upgradableTo.id,
+      )
+    },
+  )
```
jbranchaud, line 63: I originally (in yesterday's PR) got the filtering wrong -- we want to filter out an potential upgrade _if_ the `upgradableTo` has already been purchased.

packages/commerce-server/src/format-prices-for-product.ts
```diff
@@ -166,12 +166,22 @@ export async function formatPricesForProduct(
...
+    })
+  }
+
+  // Right now, we have fixed discounts to apply to upgrades for indvidual
+  // purchases. If it is a bulk purchase, a fixed discount shouldn't be
+  // applied. It's likely this will change in the future, so this allows us
+  // to handle both and distinguishes them as two different flows.
+  const fixedDiscountForUpgrade = result.bulk
+    ? 0
+    : await fireFixedDiscountForIndividualUpgrade()
```
jbranchaud, line 184: Separate out the fixed discount based on the individual and bulk path. For now, a fixed discount on a bulk purchase is always going to be 0. It will be easy enough to insert some different functionality when that changes.

As for the existing fixed discount function, that has been renamed to `getFixedDiscountForIndividualUpgrade` to emphasize that it is for **individual** purchases, and shouldn't be used for bulk.

joelhooks (approved):

PR #1142, opened 2023-08-28
PPP checkbox shouldnt show when bundle ppp discount is auto-applied

jbranchaud:
This PR takes advantage of the work done in #1117 to know whether PPP is being applied and if we should show or hide the checkbox.

This PR still includes the work done to make the Price Formatter configurable. The call site can request PPP to be not be auto-applied if it would like (test included), though we choose to always auto-apply for now.

Here is a screenshot of upgrading from Core(PPP) to Bundle(PPP). The PPP discount is auto-applied and the checkbox is hidden.

CleanShot 2023-08-28 at 11 07 30@2x

ronco

### DIFF
--- a/packages/skill-lesson/path-to-purchase/pricing.tsx
+++ b/packages/skill-lesson/path-to-purchase/pricing.tsx
@@ -436,6 +436,7 @@ export const Pricing: React.FC<React.PropsWithChildren<PricingProps>> = ({
               setMerchantCoupon={setMerchantCoupon}
               index={index}
               setAutoApplyPPP={setAutoApplyPPP}
+              purchaseToUpgradeExists={Boolean(purchaseToUpgrade)}
             />
           )}
           <div data-pricing-footer="">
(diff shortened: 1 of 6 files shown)


packages/skill-lesson/path-to-purchase/pricing.tsx
```diff
@@ -679,20 +701,25 @@ const RegionalPricingBox: React.FC<
...
-          checked={Boolean(merchantCoupon)}
-          onChange={() => {
-            merchantCoupon
-              ? setMerchantCoupon(undefined)
-              : setMerchantCoupon(pppCoupon as any)
-          }}
-        />
-        <span>Activate {percentOff}% off with regional pricing</span>
-      </label>
+      {!hideCheckbox && (
```
jbranchaud, line 706: Don't show the checkbox if this is a PPP discounted bundle upgrade.

joelhooks (approved):

PR #1179, opened 2023-09-18
fix(tjs): make text visible on post-redeem page

jbranchaud:
On the post-redeem page, I wasn't able to see the text. That is because it is white-on-white.

The highlighting (`selection-...`) was also behaving weirdly, so I removed those classes.

CleanShot 2023-09-18 at 10 12 24@2x

bojack


DrShpongle (approved):
🙌

PR #1192, opened 2023-10-03
TJS: tweak pricing component

DrShpongle:
The tiers that are lower than purchased one are gray-out and not available for purchase now.

screencapture-localhost-3018-buy-2023-10-04-00_04_05 (1)

Out Of Office Wine GIF by NETFLIX

### DIFF
--- a/apps/testing-javascript/src/components/pricing-section.tsx
+++ b/apps/testing-javascript/src/components/pricing-section.tsx
@@ -70,7 +70,7 @@ const PricingSection: React.FC<{
                   index={i}
                   couponId={couponId}
                   allowPurchase={allowPurchase}
-                  isLowerTier={productRank < highestPurchasedProductRank}
+                  unavailable={productRank < highestPurchasedProductRank}
                 />
               )
             })}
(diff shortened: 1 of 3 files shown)


apps/testing-javascript/src/components/pricing-section.tsx
```diff
@@ -1,13 +1,20 @@
...
 import Balancer from 'react-wrap-balancer'
 
 import {Pricing} from 'path-to-purchase/pricing'
 
+const PRODUCTS_RANKS = {
+  'kcd_da6ab36c-b091-4f6f-90aa-d7db2fc798ff': 1,
+  'kcd_fb976b99-0633-4329-bbfb-f5f76dc278b3': 2,
+  'kcd_4f0b26ee-d61d-4245-a204-26f5774355a5': 3,
+}
+
```
DrShpongle, line 17: Hardcoded tiers and related ranks

jbranchaud: Though there is probably some clever way to use the `UpgradableProducts` table to derive this ranking, I think this is probably a better and quicker approach to the same result 💯

jbranchaud (approved):
this looks great 🙌

PR #1197, opened 2023-10-07
feat(progress): add lessonSlug to epic-web progress API

kentcdodds:
lesson slug

This is necessary because on https://gist.githubusercontent.com/joelhooks/ca912713d61ef024da520a0d2fec135c/raw/eb4bd1a4759e8447581a13ee717ee567c1aedefd/module.json we have the `solution` which does not iniclude its ID. 

I thought about adding it here: https://github.com/skillrecordings/products/blob/19a3680302e45ff686de6c521717d69ef849641b/packages/skill-lesson/lib/modules.ts#L22

But that impacts more than just epic web. I'll make a PR for that anyway and you can merge whatever you like.

### DIFF
--- a/apps/epic-web/src/pages/api/progress.ts
+++ b/apps/epic-web/src/pages/api/progress.ts
@@ -18,6 +18,7 @@ const lesson = async (req: NextApiRequest, res: NextApiResponse) => {
         lessonProgress.map((progress) => {
           return {
             lessonId: progress.lessonId,
+            lessonSlug: progress.lessonSlug,
             completedAt: progress.completedAt,
           }
         }),


apps/epic-web/src/pages/api/progress.ts
```diff
@@ -18,6 +18,7 @@ const lesson = async (req: NextApiRequest, res: NextApiResponse) => {
         lessonProgress.map((progress) => {
           return {
             lessonId: progress.lessonId,
+            lessonSlug: progress.lessonSlug,
```
joelhooks, line 21: we specifically only store the lessonId (it's Sanity id) and don't use slugs for recording progress. we've found that it is more stable and gives learners a more consistent progress experience. lesson/module/section slugs are still in the data model for legacy (lazy) reasons, but aren't used.

Can you let me know what you're thinking here?

kentcdodds: That makes sense. The reason I thought to do this is because the API for getting modules includes a lesson "solution" which doesn't have an ID. If you merge #1198 then I think that should work.

PR #1204, opened 2023-10-10
chore(ew): return 401 or 403 for device-req

### LINKED ISSUE #1202: Return 401 or 403 for lesson-for-device-req
I have special logic in the Workshop App that will display the proper text to the user based on whether the API returns 401 (unauthenticated) or 403 (unauthorized). Right now I think it just returns a 404 if the user does not have access.

https://github.com/skillrecordings/products/blob/8225422c8e087c56f54dc8a7332a5c1091746324/apps/epic-web/src/server/lesson-for-device-req.ts#L34

I think we need an else case here: https://github.com/skillrecordings/products/blob/8225422c8e087c56f54dc8a7332a5c1091746324/apps/epic-web/src/lib/lessons.ts#L43C40-L43C40

vojtaholik:
fixes #1202

gif


apps/epic-web/src/server/lesson-for-device-req.ts
```diff
@@ -27,11 +28,16 @@ export async function lessonForDeviceReq({
...
       useSolution: isSolution,
     })
-
     if (lessonForDevice) {
       res.status(200).json(lessonForDevice)
     } else {
-      res.status(404).end()
+      if (user) {
+        // unauthorized
+        res.status(403).json(lessonForDevice)
```
joelhooks, line 36:
```suggestion
        res.status(403).end()
```

in the else the lessonForDevice will be `undefined`

joelhooks (approved):

PR #1207, opened 2023-10-10
feat(ew): dynamic sale share cards

vojtaholik:
if sanity product is active and default coupon is applied, generate a share card with dynamic discount percentage:

generated share card

gif


joelhooks (approved):

kentcdodds (commented):
This is rad

Issue #1208, opened 2023-10-10
Add epic-web endpoint to get user's purchases


kentcdodds:
As part of my "connect to discord" flow, my server makes an authenticated request to `/api/progress` to determine the validity of the user's device token. This seems brittle because that endpoint could reasonably be changed in the future to return success for non-purchasers.

It would be best if I had an endpoint that would return the user's purchases so I could determine whether they really should have access.

PR #1209, opened 2023-10-10
use the complete user

joelhooks:
user


kentcdodds (approved):
Looks good to me! For now I plan on only accessing the id, name, and email. If image ever gets support then I'll probably use that too.

And actually I think this solves #1208 right too. I probably could've used that instead of the progress API. The server I have only has the device token, but it looks like that's all I need to hit this API.

PR #1255, opened 2023-10-27
fix(build): pnpm bump to 8 in release.yml

johnlindquist:
https://eggheadio.slack.com/archives/G01CAT0P9TL/p1698260020696249

Apparently, my local env updating the lockfile using 8.9 is conflicting with GitHub actions running on 7. Do you see any issue with change our actions to us 8?

g version


vojtaholik (approved):
should be fine! 👍

PR #1258, opened 2023-10-31
feat(tt): loop through purchases to get bonuses

vojtaholik:
I haven't tested it properly but they now show up for the user in question.

image

gif


apps/epic-web/src/pages/products/index.tsx
```diff
@@ -100,21 +106,15 @@ export default ProductsIndex
 const ProductCard: React.FC<{
   product: SanityProduct
   purchase: PurchaseWithProduct | undefined
-}> = ({product, purchase}) => {
+  purchases: PurchaseWithProduct[] | undefined
```
joelhooks, line 109: purchases?: is more common syntax imo

joelhooks (approved):

PR #1266, opened 2023-11-06
jb/add interview pages

jbranchaud:
With the Interview data being recently added to Sanity, I wanted to get the `/interview/[slug]` page set up. However, Interviews are not lessons and the `VideoProvider` expects to be wrapped in a `LessonProvider`.

Maybe we need to just use an unwrapped instance of a Mux Player? Not sure if that has implications for out-of-the-box completion tracking and purchase/permission checking.

- fix: use `slug.current` for interview links
- WIP - add interview pages

interview


vojtaholik (approved):
LGTM 👍

PR #1267, opened 2023-11-06
feat: adding endpoints for support

nicollguarnizo:
can't do it without you


packages/skill-api/src/core/services/update-name.ts
```diff
@@ -0,0 +1,64 @@
...
+    const updatedUser = await prisma.user.update({
+      where: {
+        id: userId,
+      },
+      data: {
+        name: newName,
+      },
+    })
+
+    const merchantCustomerId = await prisma.merchantCustomer.findFirstOrThrow({
```
joelhooks, line 37: we probably don't want to throw here

joelhooks (approved):

PR #1285, opened 2023-11-23
feat: sanity product sync via webhook & inngest

vojtaholik:
screenshots

gif


joelhooks (approved):
You'll need to change any sanity webhooks you setup to use `/api/skill/webhook/sanity` instead of the other one for this to work!

PR #1288, opened 2023-11-28
Upgrade prisma to v5.0.0

jbranchaud:
I mostly pursued this because of some weird type display behavior that is fixed in newer versions of Prisma, but we also get a bunch of bug fixes, speed improvements, and new features.

- chore(deps): bump prisma to v5.0.0

I followed the upgrade guide to address any breaking changes. Very few
affected us except for needing to get the `Decimal` type from the
`Decimal.js` library.

https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-5

- feat(db): these are no longer preview features
- feat(db): this key is now relationMode

train


packages/database/src/client.ts
```diff
@@ -1,8 +1,9 @@
 import {PrismaClient} from '@prisma/client'
+import type {Decimal} from 'decimal.js'
```
jbranchaud, line 2: It doesn't seem like `Decimal` is directly exported as part of `Prisma` and since they are deprecating getting it off the `@prisma/client/runtime`, I decided it would be easiest to pull in the underlying `decimal.js` library and grab the type from that.

joelhooks (approved):

PR #1294, opened 2023-11-29
chore(deps): bump dep versions for react package

jbranchaud:
since these packages only exist in the monorepo and not on NPM, the versions need to match

elf


joelhooks:
I think I got this in my fix PR

PR #1299, opened 2023-11-30
refactor(api): forward all sanity events to inngest

vojtaholik:
screenshot

gif


packages/inngest/src/utils/sanity-client.ts
```diff
@@ -0,0 +1,8 @@
+import {createClient} from '@sanity/client'
+
+export const sanityClient = createClient({
```
vojtaholik, line 3: I wanted to import and use sanityClient from `skill-lesson` pkg but that'd be a cyclic dependency :|

packages/skill-api/src/core/services/process-sanity-webhooks.ts
```diff
@@ -1,13 +1,12 @@
...
-import {getSdk} from '@skillrecordings/database'
 import {Inngest} from 'inngest'
 import {SYNC_SANITY_PRODUCT} from '@skillrecordings/inngest'
 import {isValidSignature, SIGNATURE_HEADER_NAME} from '@sanity/webhook'
 
-const webhookSecret = process.env.SANITY_WEBHOOK_SECRET as string // TODO: rename?
+const webhookSecret = process.env.SANITY_WEBHOOK_SECRET as string
 
-export async function processSanityProductsWebhook({
+export async function processSanityWebhooks({
```
vojtaholik, line 9: process all sanity webhooks

packages/skill-api/src/core/services/process-sanity-webhooks.ts
```diff
@@ -31,23 +30,23 @@ export async function processSanityProductsWebhook({
...
-            product: req.body,
-          },
-        })
+        console.log('sending inngest event', data)
+        if (data._type === 'product') {
+          await inngest.send({
+            name: SYNC_SANITY_PRODUCT,
+            data,
+          })
+        }
```
vojtaholik, line 49: this is where we can choose which event to send based on `_type` or `event` coming through the webhook

packages/skill-api/src/core/services/process-sanity-webhooks.ts
```diff
@@ -31,23 +30,23 @@ export async function processSanityProductsWebhook({
...
-        await inngest.send({
-          name: SYNC_SANITY_PRODUCT,
-          data: {
-            product: req.body,
-          },
-        })
+        console.log('sending inngest event', data)
+        if (data._type === 'product') {
+          await inngest.send({
+            name: SYNC_SANITY_PRODUCT,
```
joelhooks, line 46: the way to do this is to forward an "event type" and not build a stack of specific events so this handler ends up being a simple relay similar to stripe or mux

this means on the sanity side there will be a `type` at the top of the event like "type": "sanity/" + _type + "." + operation 


so then you can just send that out instead of conditionals


as an aside with events like this is always "what has occurred" not "what you want to happen" with event names. So "product created" or "product updated" would be the event name

jbranchaud: this is a good example of forwarding the event along and letting inngest do the filtering/conditionals https://github.com/joelhooks/course-builder/blob/main/apps/course-builder-web/src/app/api/mux/webhook/route.ts

packages/inngest/src/utils/sanity-client.ts
jbranchaud: we have this exact file exporting `sanityClient` in every app and many/most packages, could we create a super-small `sanity-client` package that does this once and then each app can import it. I think each app's individual `process.env` would be applied.

packages/skill-api/src/core/services/process-sanity-webhooks.ts
joelhooks: the principle here is that we want to edit `skill-api` package "practically never"

packages/inngest/src/utils/sanity-client.ts
joelhooks: the inngest package should be nuked

apps/epic-web/src/inngest/functions/sanity/product.ts
```diff
@@ -0,0 +1,114 @@
...
+export const sanityProductUpdated = inngest.createFunction(
+  {id: `product-update`, name: 'Update Product in Database'},
+  {
+    event: SANITY_WEBHOOK_EVENT,
+    if: 'event.data.event == "product.update"',
+  },
+  async ({event, step}) => {
+    const productToUpdate = await step.run('get product', async () => {
+      return await getProduct(event.data._id)
+    })
```
joelhooks, line 51: `await step.run` is needed for inngest!

apps/epic-web/src/inngest/functions/sanity/product.ts
```diff
@@ -0,0 +1,114 @@
...
+    {id},
+  )
+  return product
+}
+
+export const sanityProductUpdated = inngest.createFunction(
+  {id: `product-update`, name: 'Update Product in Database'},
+  {
+    event: SANITY_WEBHOOK_EVENT,
+    if: 'event.data.event == "product.update"',
```
joelhooks, line 46: instead of individual events we can filter events like this, makes config boilerplate simpler overall

apps/epic-web/src/inngest/inngest-config.ts
```diff
@@ -0,0 +1,23 @@
+import {inngest} from 'inngest/inngest.server'
+import {processNewTip} from 'inngest/functions/tips/process-new-tip'
+import {addSrtToMuxAsset} from 'inngest/functions/tips/add-srt-to-mux-asset'
+import {stripeCheckoutCompleted} from 'inngest/functions/stripe/checkout-completed'
+import {lessonCompleted} from 'inngest/functions/progress/lesson-completed'
+import {writeAnEmail} from 'inngest/functions/ai-email/write-an-email'
+import {purchaseTransferComplete} from 'inngest/functions/purchase/purchase-transfer-complete'
+import {productFunctions} from 'inngest/functions/sanity/product'
+
+export const inngestConfig = {
```
joelhooks, line 10: this was stuffed in `/apps/api/inngest/route.ts` so I lifted it to here so ALL inngest edits happen in `/inngest`

apps/epic-web/src/lib/products.ts
```diff
@@ -57,3 +58,106 @@ export async function getProduct(productId: string): Promise<Product> {
...
+              identifier: product.merchantProduct.identifier,
+              merchantAccountId: product.merchantProduct.merchantAccountId,
+            },
+          },
+        },
+      }),
+    },
+  })
+  return newProduct
+}
```
joelhooks, line 163: not super stoked on this and think in particular we need to reconsider the `any` for `product`

any ideas @jbranchaud? this could be a followup

PR #1300, opened 2023-11-30
feat: add admin-protected tRPC routes

jbranchaud:
This adminProcedure can be used to define tRPC routes that require an
authenticed user with an admin role in order for the request to proceed,
otherwise it returns 401.

Resources and references:
- [Using Next-Auth with tRPC (t3 docs)](https://create.t3.gg/en/usage/next-auth#:~:text=When%20using%20NextAuth.,be%20accessed%20by%20authenticated%20users.&text=sets%20all%20of%20this%20up,session%20object%20within%20authenticated%20procedures.)
- [NextAuth Session Callback](https://next-auth.js.org/configuration/callbacks#session-callback)
- [tRPC Authorization Middleware](https://trpc.io/docs/server/middlewares#authorization)
- [tRPC Typing and Creating Context](https://trpc.io/docs/server/context)

rc


packages/skill-lesson/trpc/routers/coupons.ts
```diff
@@ -49,7 +49,7 @@ export const couponsRouter = router({
     })
     return coupons
   }),
-  delete: publicProcedure
+  delete: adminProcedure
```
jbranchaud, line 52: This allows us to make something like _delete a coupon_ an admin-only action, where tRPC checks in the middleware if the user is authorized to take the action.

packages/skill-lesson/trpc/trpc.server.ts
```diff
@@ -31,6 +31,40 @@ const t = initTRPC.context<TrpcContext>().create({
...
+
+  return next({
+    ctx: {
+      // infers the `session` as non-nullable
+      session: {...ctx.session, user: ctx.session?.user},
+    },
+  })
+})
+
+export const adminProcedure = t.procedure.use(isAdmin)
```
jbranchaud, line 51: Uses the base procedure and then layers in an `isAdmin` check

packages/skill-api/next-auth.d.ts
```diff
@@ -1,12 +1,20 @@
 import NextAuth, {DefaultSession} from 'next-auth'
 
+type Resolve<T> = T extends Function ? T : {[K in keyof T]: T[K]}
+
+type User = Resolve<
+  {
+    role?: string
+    purchases?: any[]
+  } & DefaultSession['user']
+>
```
jbranchaud, line 10: This expands the NextAuth `Session` / `User` type to allow `user` to include things like `role` and `purchases`.

packages/skill-lesson/trpc/trpc.server.ts
```diff
@@ -8,9 +8,29 @@
...
+
+declare module 'next-auth' {
+  /**
+   * Returned by `useSession`, `getSession` and received as a prop on the
+   * `SessionProvider` React Context
+   */
+  interface Session {
+    user?: User
+  }
+}
```
jbranchaud, line 33: I don't totally have my head wrapped around the issue, but it seems like when this `declare module ...` was in the `next-auth.d.ts` at the root-level, it wasn't getting picked up by anything during a lint or build. In contrast to the `skill-api` package where it is getting picked up, I'm guessing it has something to do with it being transpiled as an [internal package](https://turbo.build/repo/docs/handbook/sharing-code/internal-packages).

Even adding `next-auth.d.ts` to the `include` of the `skill-lesson`'s `tsconfig.json` wasn't getting it picked up.

However, putting it directly in `trpc.server.ts` does get it picked up during build/lint.

What is extra strange is that the types were getting properly picked up by during development by VS Code.

image

joelhooks (approved):

PR #1305, opened 2023-12-04
feat(ew): handle unfound exercise scenario

jbranchaud:
Because we look up the exercise based on a slug that comes from a URL
(user input), it is possible that the slug is bad and we don't find an
exercise. At the moment, the Zod parse fails the validation and the
server responds with a 500. Instead, we can handle the Record Not Found
scenario by returning `null` and having the various call sites react
appropriately. Often that means rendering a 404 response.

CleanShot 2023-12-04 at 16 07 01@2x

https://skill-recordings.sentry.io/issues/4545337140/?_allp=1&query=is%3Aunresolved&referrer=issue-stream&stream_index=6

not found


vojtaholik:
@jbranchaud this is great 🎉 do you think you could apply these changes to skill template as well? :)

jbranchaud:
> @jbranchaud this is great 🎉 do you think you could apply these changes to skill template as well? :)

@vojtaholik for sure! you mind if I do that in a separate PR to not make this one any bigger than it already is?

vojtaholik (approved):
🌮 🌮 🌮

PR #1311, opened 2023-12-08
fix(builds): await isValidSignature in sanity webhook

vojtaholik:
`isValidSignature` from `@sanity/webhook` returns type of `Promise<boolean>` thus has to be awaited.

gif


jbranchaud (approved):
good catch 🙌

PR #1313, opened 2023-12-12
chore: setup inngest for jjs

joelhooks:
setup


apps/testing-javascript/src/pages/api/inngest.ts
```diff
@@ -0,0 +1,8 @@
+import {serve} from 'inngest/next'
+import {inngestConfig} from '../../inngest/inngest.config'
+
+export const config = {
+  maxDuration: 300,
+}
```
jbranchaud, line 6: I wasn't sure "who" this exported config was for. TIL that Next.js supports per-route configuration for things like `maxDuration` and streaming the raw request body https://nextjs.org/docs/pages/building-your-application/routing/api-routes#custom-config

PR #1320, opened 2023-12-19
fix: NEXTAUTH_URL has to match app URL exactly

jbranchaud:
http -> https

auth


apps/testing-javascript/.env.production
```diff
@@ -34,7 +34,7 @@ NEXT_PUBLIC_SANITY_DATASET_ID=production
 NEXT_PUBLIC_SANITY_API_VERSION=2021-10-21
 
 # NextAuth
-NEXTAUTH_URL=http://testing-javascript.vercel.app
+NEXTAUTH_URL=https://testing-javascript.vercel.app
```
jbranchaud, line 37: this mismatch is why NextAuth was not able to decode the encrypted JWTs in production

vojtaholik (approved):

PR #1322, opened 2023-12-27
jb/fix undefined product access in stripe checkout

jbranchaud:
fix: throw error when product is incomplete

if we are unable to look up the entire product and its associations,
either because those records are missing or because they are not all
marked as active, then we should preemptively throw instead of getting
to a point in the checkout function where we try to do a property access
on an undefined value.

This is related to the error that was reported by a learner trying to purchase an epic-web workshop a la carte last week.

I have some ideas for better handling of exceptional cases while building the checkout session, but as I starting point I wanted to deliver a better error message than `"Cannot read properties of undefined (reading 'identifier')"`.

christmas story


packages/skill-api/src/core/services/stripe-checkout.ts
```diff
@@ -244,6 +263,30 @@ export async function stripeCheckout({
         },
       })
 
+      const result = LoadedProductSchema.safeParse(loadedProduct)
```
jbranchaud, line 266: check for these values toward the top of the stripe checkout builder so that we can bail early with a better error message if they are missing (missing meaning that some aspect of the `Product -> MerchantProduct -> MerchantPrice` data setup is invalid/inactive.

packages/skill-api/src/core/services/stripe-checkout.ts
```diff
@@ -313,9 +356,7 @@ export async function stripeCheckout({
             redeem_by: TWELVE_FOUR_HOURS_FROM_NOW,
             currency: 'USD',
             applies_to: {
-              products: [
-                loadedProduct.merchantProducts?.[0].identifier as string,
-              ],
+              products: [merchantProductIdentifier],
```
jbranchaud, line 359: No need to type cast to string here since we ensure it is a string when parsed earlier

packages/skill-api/src/core/services/stripe-checkout.ts
```diff
@@ -346,17 +387,6 @@ export async function stripeCheckout({
...
-      const price =
-        loadedProduct.merchantProducts?.[0].merchantPrices?.[0].identifier
-
-      if (!price) {
-        throw new CheckoutError(
-          'no-pricing-available',
-          loadedProduct.id,
-          couponId as string,
-        )
-      }
```
jbranchaud, line 358: We extract this value and throw `CheckoutError` as necessarily at the top of the function, no need to do it here

joelhooks (approved):
great use of zod

PR #1325, opened 2023-12-29
feat: improve error page on failed Stripe Checkout

jbranchaud:
As of right know when an error bubbles up while we are creating the Stripe Checkout Session, we have a broad try/catch that renders a default server 500 (pictured below). This error isn't actionable for the user, we lose the site branding, and we don't necessarily know it happened.

We can set up Sentry or other alerts so that we know when this is happening. And to improve the user experience, we can allow each app to opt-in to sending the user to a customized, branded error page with better copy and actionable next steps.

This PR is a spike of how we can do that by allowing apps to optionally pass in an `errorRedirectUrl` that our Stripe Checkout Setup will redirect to instead of rendering a 500. We've had learners recently run into 500 pages like this will checking out. We want to alert ourselves when this happens and also give the learner a chance to contact support and some assurance that we're looking into it.

Here is what you currently see when an error happens in Stripe Checkout setup:

CleanShot 2023-12-29 at 09 52 48@2x

Here is a spike of a customizable page we can show instead:

CleanShot 2023-12-29 at 10 19 12@2x

customer support


apps/testing-javascript/src/pages/error.tsx
```diff
@@ -0,0 +1,32 @@
...
+import {useRouter} from 'next/router'
+
+const ErrorPage = ({}) => {
+  const router = useRouter()
+  const {type} = router.query
+
+  let message = 'There was an unexpected error'
+
+  if (type === 'stripe-checkout') {
+    message = 'Something went wrong with the Stripe Checkout session'
```
jbranchaud, line 12: This page definitely needs copy work, I was just throwing together a spike.

packages/skill-api/src/core/services/stripe-checkout.ts
```diff
@@ -408,6 +412,13 @@ export async function stripeCheckout({
...
     } catch (err: any) {
+      if (errorRedirectUrl) {
+        return {
+          redirect: errorRedirectUrl,
+          status: 303,
+        }
+      }
+
       return {
         status: 500,
```
jbranchaud, line 423: This is where we can redirect to a custom error page for an app instead of just rendering a 500

PR #1330, opened 2024-01-07
fix: filter purchases to just those that are TJS

jbranchaud:
The logic for finding the most valued product wasn't account for purchases across other apps in the same database. Adding a line to filter down to just those products that are Testing JavaScript related is the most direct way to deal with this. There is probably more of refactoring of this logic that could be done, but I don't think it is worth the squeeze.

This fixes the error that @joelhooks was seeing on friday:

CleanShot 2024-01-07 at 15 49 18@2x

MVP


apps/testing-javascript/src/pages/index.tsx
```diff
@@ -22,6 +22,12 @@ import type {TestimonialProps, FaqProps, InterviewProps} from '@types'
 
 import LandingTemplate from 'templates/landing-template'
 
+const testingJavaScriptProductIds = [
```
joelhooks, line 25: imo this should be dynamic instead of hard coded in general but it's OK for here probanly

joelhooks (approved):

PR #1335, opened 2024-01-10
feat(ER): epic-react new app init

DrShpongle:
This is a fresh start of the `epic-react-next`.

The old app renamed to `epic-react-delete-me` for now and will be deleted after @jbranchaud will save/check the stuff he might need.

CleanShot 2024-01-10 at 19 47 00@2x

E Network Fresh Start GIF by E!


apps/epic-react-legacy/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "epic-react",
+  "name": "epic-react-legacy",
```
joelhooks, line 2: we can keep the name epic-react since there won't be a conflict 
```suggestion
  "name": "epic-react",
```

joelhooks (approved):

jbranchaud:
I think we will need to call the "old" directory something other than `epic-react` for the time being because otherwise it causes CI failures across all the apps. I just need time to pull over my data scripts and any other useful artifacts before we `rm` it.

CleanShot 2024-01-10 at 16 31 19@2x

DrShpongle:
> I think we will need to call the "old" directory something other than `epic-react` for the time being because otherwise it causes CI failures across all the apps. I just need time to pull over my data scripts and any other useful artifacts before we `rm` it. https://github.com/skillrecordings/products/pull/1335/commits/6855939b96902c1e76a1e8cae820aa5777e3a311
> 
> 
> 
> CleanShot 2024-01-10 at 16 31 19@2x
> 
> 

Yes, I mentioned it in the chat. But if I understand correctly problem is not with folder name, but `name` field in the `package.json`. So to avoid that namespace conflict we should rename it to something different from `epic-react` to be able to use `epic-react` name for the new app.

PR #1341, opened 2024-01-16
feat(ewd): polish event template & purchase flow

vojtaholik:
screenshot
image
screenshot

gif

### DIFF
--- a/apps/epic-web/src/lib/products.ts
+++ b/apps/epic-web/src/lib/products.ts
@@ -8,7 +8,7 @@ export const ProductSchema = z.object({
   _updatedAt: z.string(),
   _createdAt: z.string(),
   title: z.string(),
-  type: z.enum(['self-paced', 'live']).optional(),
+  type: z.enum(['self-paced', 'live']),
   slug: z.string(),
   image: z
     .object({
--- a/apps/epic-web/src/lib/products.ts
+++ b/apps/epic-web/src/lib/products.ts
@@ -8,6 +8,7 @@ export const ProductSchema = z.object({
   _updatedAt: z.string(),
   _createdAt: z.string(),
   title: z.string(),
+  type: z.enum(['self-paced', 'live']).optional(),
   slug: z.string(),
   image: z
     .object({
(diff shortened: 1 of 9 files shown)


packages/commerce-server/src/@types/index.ts
```diff
@@ -60,13 +60,14 @@ export type SanityProduct = {
...
   slug?: string
   image: {
     url: string
     alt: string
   }
-  state?: 'draft' | 'active' | 'unavailable'
+  state?: 'draft' | 'active' | 'unavailable' | 'archived'
   summary?: PortableTextBlock
   modules: SanityProductModule[]
   upgradableTo?: SanityProduct
```
vojtaholik, line 73: added new `type` type. we were also missing `archived` state.

apps/epic-web/src/lib/products.ts
```diff
@@ -8,6 +8,7 @@ export const ProductSchema = z.object({
   _updatedAt: z.string(),
   _createdAt: z.string(),
   title: z.string(),
+  type: z.enum(['self-paced', 'live']).optional(),
```
joelhooks, line 11:
```suggestion
  type: z.enum(['self-paced', 'live']),
```

not optional

apps/epic-web/src/templates/event-template.tsx
```diff
@@ -182,62 +216,87 @@ const Header: React.FC<HeaderProps> = ({
...
+  quantityAvailable: number
+  commerceProps: CommerceProps
+}> = ({product, quantityAvailable, commerceProps}) => {
+  const router = useRouter()
+  const couponFromCode = commerceProps?.couponFromCode
+  const {validCoupon} = useCoupon(commerceProps?.couponFromCode)
+  const couponId =
+    commerceProps?.couponIdFromCoupon ||
+    (validCoupon ? couponFromCode?.id : undefined)
+  console.log({couponFromCode, couponId, commerceProps})
```
joelhooks, line 262:
```suggestion
```

killed the console

packages/commerce-server/src/@types/index.ts
joelhooks: we've got a separate `Product` type that actually has a schema in the commerce package that I prefer generally because it has a backing schema where this one is pre-Zod

joelhooks (approved):

PR #1344, opened 2024-01-16
feat: fetch only used articles from sanity

jbranchaud:
We have static page generation of at least one of the TT articles failing consistently in TT.

I haven't tracked down the cause of this, but while taking an initial look I noticed that we are over-fetching article metadata from Sanity during `getStaticProps`. I figured it might help to only fetch the 6 articles we need, rather than all ~50 (and counting).

CleanShot 2024-01-16 at 14 47 11@2x

pick six


jbranchaud:
This didn't fix the build failure in CI, but it is still an improvement to the getStaticProps fetching for an article.

CleanShot 2024-01-16 at 16 22 01@2x

joelhooks (approved):
r

PR #1346, opened 2024-01-17
feat(ewd): custom welcome video & event details

vojtaholik:
Pixel Perfect Figma to Tailwind

we don’t have welcome video for this event (yet). for self-paced products I left fallback to the one we’re using.

Pasted Graphic 2

I think it makes sense for video resources to have customisable poster image, can come in handy for open graph images, etc. right now we default to first frame.

Pasted Graphic 3

welcome video is a reference to videoResource

gif

### DIFF
--- a/apps/epic-web/src/pages/events/[event].tsx
+++ b/apps/epic-web/src/pages/events/[event].tsx
@@ -73,7 +73,9 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
         mdx,
         availableBonuses,
         quantityAvailable:
-          productWithQuantityAvailable?.quantityAvailable || -1,
+          (productWithQuantityAvailable &&
+            productWithQuantityAvailable.quantityAvailable - purchaseCount) ||
+          -1,
         purchaseCount,
       },
     }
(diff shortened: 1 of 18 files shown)


apps/epic-web/src/pages/events/[event].tsx
```diff
@@ -111,7 +114,7 @@ export type EventPageProps = {
   event: Event
   quantityAvailable: number
   purchaseCount: number
-  product: Product
+  product: SanityProduct
```
Comment (vojtaholik) [inline apps/epic-web/src/pages/events/[event].tsx:117]: this is actually expecting SanityProduct down the road :/ there's quite a few dependencies on it, seems like it would be a bigger refactor

apps/epic-web/src/templates/purchased-event-template.tsx
```diff
@@ -147,11 +163,14 @@ const PurchasedEventTemplate = ({
                   <>
                     <p>
                       <Balancer>
-                        Welcome to {product.title}! We're thrilled to have you
-                        join us for this event. Below, you'll find everything
-                        you need to manage your purchase and access related
-                        information. If you have any questions or need
-                        assistance at any point, please don't hesitate to{' '}
+                        <strong>
+                          Welcome to {product.title}! We're thrilled to have you
```
vojtaholik, line 167: here we should probably let them know they’ll receive a zoom link to event via email

PR #1347, opened 2024-01-17
feat(ER-next): initial data and structure for the Learn page

DrShpongle:
This is very initial `/learn` page:
CleanShot 2024-01-18 at 17 31 32@2x

Still requires:
1. get rid of the `[section]` segment in url
2. pull workshops from purchased product instead of `getAllWorkshops` query
3. Output not only lessons under the sections, but  resources that includes the lessons like intro/outro


apps/epic-react/src/pages/learn.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <h2 className="text-center text-5xl">Learn Page</h2>
+        <ul className="space-y-6">
+          {workshops.map((workshop: any) => {
+            return (
+              <li key={workshop._id} className="flex space-x-6">
+                <Image src={workshop.image} alt="" width={200} height={200} />
+                <div className="space-y-3">
+                  <h3 className="text-2xl">{workshop.title}</h3>
+                  <ul>
+                    {workshop.sections.map((section: any) => {
```
DrShpongle, line 44: `resources` should be here

jbranchaud (approved):
🙌

PR #1361, opened 2024-01-22
feat: exclude coupon if not for productId

jbranchaud:
there are many pages where we show pricing for multiple products (e.g.
when there are different tiers) and the coupon coming from a coupon code
is only relevant to one of those products. Ensure that the price gets
calculated correctly for each by restricting the coupon to its product.

### Before

image

### After

CleanShot 2024-01-22 at 12 10 24@2x

relevant

### DIFF
--- a/packages/commerce-server/src/format-prices-for-product.test.ts
+++ b/packages/commerce-server/src/format-prices-for-product.test.ts
@@ -107,15 +107,6 @@ test('does not apply restricted coupon to other product', async () => {
   const ONE_OFF_COUPON_FROM_CODE = 'one-off-coupon-from-code-id'
   const ONE_OFF_MERCHANT_COUPON_ID = 'one-off-merchant-coupon-id'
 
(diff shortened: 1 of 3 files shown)


packages/commerce-server/src/format-prices-for-product.test.ts
```diff
@@ -102,6 +102,47 @@ for (const quantity of [69, 89, 99]) {
...
+      merchantCouponId: ONE_OFF_MERCHANT_COUPON_ID,
+      usedCouponId: ONE_OFF_COUPON_FROM_CODE,
+      ctx,
+    })
+
+  const expectedPrice = 100
+
+  expect(expectedPrice).toBe(calculatedPrice)
+  expect(appliedMerchantCoupon).toBe(undefined)
+})
```
joelhooks, line 144: imo this should be two tests, but love to see the tests

joelhooks (approved):
love it!

PR #1365, opened 2024-01-24
feat: handle un-sectioned lessons

jbranchaud:
- cleanup: collapse duplicate const from useLesson
- feat: variable renaming to prepare refactor
- feat: render un-sectioned lessons to list

Note: this is incomplete / proof-of-concept and needs some work. I think
the lessons and sections that are getting passed down through the
Collection component as a result of these changes aren't entirely
compatible. Some parts are regressing slightly.

@vojtaholik it looks like you are the primary author of this component, so I'd love to get your feedback and insights on how to accomplish this.

A big takeaway that I had from digging into the Collection component (https://github.com/skillrecordings/products/blob/main/packages/skill-lesson/video/collection/index.tsx#L117) is that it heavily relies on the `module` and the structure of `module.sections` nested within that (which come from the `LessonProvider` component).

My idea is that we can make this component work for modules that have both top-level lessons and lessons nested in sections (currently it assumed lessons are nested in sections). To work toward that, I tried shifting the component logic to `module.resources` (if the app opts-in to that via `module.useResourcesInsteadOfSections`) where a resource can be either a `lesson` or a `section`.

Another tangential thing I noticed is that `module.lessons` is always `undefined` because no app makes a groq query where `module.lessons` is defined and passed in to the `LessonProvider`. This doesn't seem to matter to much, but thought it was worth pointing out since there is a bit of logic in `Collection` that operates on that value.

CleanShot 2024-01-23 at 18 33 54@2x

dazed


vojtaholik:
abilities (locked state) is one of the regressed parts I've noticed

packages/skill-lesson/video/collection/index.tsx
```diff
@@ -305,22 +307,46 @@ const Sections = React.forwardRef<SectionsElement, SectionsProps>(
...
+                    // lessonsProps.className,
+                  )}
+                >
+                  <li key={resource._id}>
+                    <Lesson
+                      section={undefined}
+                      lesson={resource}
+                      index={resource._id}
+                    />
+                  </li>
```
vojtaholik, line 346: ideally this could be styled using children approach, similar to the rest. I'll look into it asap.

jbranchaud:
> this seems to be working pretty good @jbranchaud @DrShpongle

@vojtaholik thanks so much for digging into this! 🌮  I got sidetracked with a couple other issues, but will check this out soon.

jbranchaud:
Something that got a little messy is that there are a bunch of places in the app that pass around Lessons (explainers, exercises) and Solutions as if they are interchangeable. And they mostly are. Lessons and Solutions look mostly the same. But one way they are distinct is that Lessons are their own Sanity object so they have an `_id` field. Whereas Solutions are a nested resource without an `_id` field. So I ended up having to attach an `omit -> _id` to several places where a generic `Resource` was specifically acting as a Solution.

packages/skill-lesson/video/collection/index.tsx
```diff
@@ -792,6 +802,50 @@ const useScrollToActiveLesson = (
   }, [activeElementToScrollTo, scrollContainerRef])
 }
 
+const getLessonChildren = (children: React.ReactNode, props: LessonProps) => {
```
jbranchaud, line 805: It seems like `getLessonChildren` and `getSectionChildren` aren't needed anymore.

I did some refactoring because the rendering was getting weird, such that each section was rendering multiple times. I think that was because we had the root component rendering sections and lessons, but then the nested Section components were using data from context to render themselves additional times.

I shifted some logic down to the `Sections` component to deal with this. Which is why these two functions are no longer being used.

@vojtaholik I'd appreciate if you could advise on these changes since you have a lot more familiarity with this set of components.

image

vojtaholik: I haven't looked at the code yet, but purpose of the `getLessonChildren` and `getSectionChildren` was primarily so that we can apply separate styles from the lesson template level.

jbranchaud:
I just noticed while smoke testing this on Epic Web that there is a regression in the _Can View Content_ logic that is locking some lessons. Putting a `NO MERGE` label for now while I figure that out.

DrShpongle (approved):

PR #1371, opened 2024-01-27
feat: epic web edit routes in app router

joelhooks:
editor


apps/epic-web/src/server/auth.ts
```diff
@@ -0,0 +1,79 @@
...
+          ...session,
+          user: {
+            ...session.user,
+            id: dbUser?.id,
+            role,
+          },
+        }
+      },
+    },
+  })
```
joelhooks, line 79: I had to duplicate the next-auth config for the app router serverside session grabbing otherwise it was throwing deep cutting errors

packages/skill-lesson/markdown/mdx.tsx
```diff
@@ -1,4 +1,5 @@
-import * as React from 'react'
+'use client'
```
joelhooks, line 1: this could probably be moved to a "closer" component wrapper, but had to be used to get this to render

if we were using RSC for the actual presentation and not the just the "backend" this might affect seo

apps/epic-web/src/app/articles/[slug]/edit/page.tsx
```diff
@@ -0,0 +1,39 @@
...
+}: {
+  params: {slug: string}
+}) {
+  const session = await getServerAuthSession()
+  const ability = getCurrentAbility({user: session?.user})
+  const article = await getArticle(params?.slug as string)
+
+  if (!article || !ability.can('create', 'Content')) {
+    notFound()
+  }
```
Comment (joelhooks) [inline apps/epic-web/src/app/articles/[slug]/edit/page.tsx:21]: we can load content and check abilities

because this is an edit route we can technically make everything below this a `use client` since we don't care about SEO

vojtaholik (approved):
thanks for setting this up 🥇🥇🥇

PR #1373, opened 2024-01-29
fix: pass productIds to validateCoupon

jbranchaud:
The `validateCoupon` function now accepts a `productIds` array for validating coupons that are restricted to a certain product ID (https://github.com/skillrecordings/products/pull/1360). The `redeem-golden-ticket` flow was not passing along a list of product IDs to that function which meant that any coupon restricted to a product ID was coming out as invalid (`coupon-not-valid-for-product`).

This was also going to impact redemption of seats for team/bulk purchases.

More details on this in roam: https://roamresearch.com/#/app/egghead/page/muT11DK4o

redemption

### DIFF
--- a/apps/devrel-fyi/src/pages/index.tsx
+++ b/apps/devrel-fyi/src/pages/index.tsx
@@ -71,7 +71,7 @@ const Home: NextPage<{
 
   // TODO: should we move the coupon call down into `<Pricing>` so that it
   // can be called separate for each product?
-  const productId = products[0].productId
+  const productId = products[0]?.productId
 
   const {redeemableCoupon, RedeemDialogForCoupon, validCoupon} = useCoupon(
     commerceProps?.couponFromCode,
--- a/apps/devrel-fyi/src/pages/index.tsx
+++ b/apps/devrel-fyi/src/pages/index.tsx
@@ -69,9 +69,14 @@ const Home: NextPage<{
       productId: defaultProduct?.productId,
     })
 
+  // TODO: should we move the coupon call down into `<Pricing>` so that it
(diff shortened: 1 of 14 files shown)


apps/devrel-fyi/src/pages/index.tsx
```diff
@@ -69,9 +69,14 @@ const Home: NextPage<{
...
     })
 
+  // TODO: should we move the coupon call down into `<Pricing>` so that it
+  // can be called separate for each product?
+  const productId = products[0]?.productId
+
   const {redeemableCoupon, RedeemDialogForCoupon, validCoupon} = useCoupon(
     commerceProps?.couponFromCode,
     {
+      id: productId,
```
jbranchaud, line 79: I put in this temporary bandaid fix for epic-react, devrel-fyi, and pro-aws since those are all in pre-purchase mode. But as the comment says, might make sense to think about moving the useCoupon call somewhere within `products.map` so that the Redeem Purchase modal will work independently for each.

apps/epic-web/src/pages/index.tsx
```diff
@@ -69,6 +69,7 @@ const Index: NextPage<{
   const {redeemableCoupon, RedeemDialogForCoupon, validCoupon} = useCoupon(
     commerceProps?.couponFromCode,
     {
+      id: product.productId,
```
jbranchaud, line 72: The fix for epic-web is to pass in the `productId` wherever this is called which gets passed down through several components and an API call and eventually makes its way into a `productIds` array for the `validateCoupon` call.

joelhooks (approved):

PR #1374, opened 2024-01-30
fix: pass in restricted product ID to coupon modal

jbranchaud:
If the coupon is restricted to a specific product, then we need to pass down that product ID as part of the metadata so that the `validateCoupon` check can verify the match. 

Ultimately there is some refactoring that needs to happen in this coupon flow, but for now this is a fix that will re-enable coupons for TJS.

succession


joelhooks (approved):

jbranchaud:
This PR includes additional fixes related to https://github.com/skillrecordings/products/pull/1373, but for Testing JavaScript.

PR #1380, opened 2024-02-01
refactor: not how we lookup defaultCoupon anymore

jbranchaud:
No apps are using the `defaultCoupon` value coming from Commerce Props, they are all looking it up themselves with [Prisma `getDefaultCoupon`](https://github.com/skillrecordings/products/blob/7f3c88cad5c0b0cdb55b1fd6c7608b7d83cb9ef4/packages/database/src/prisma-api.ts#L730) or they are getting it from [`trpc.pricing.defaultCoupon`](https://github.com/skillrecordings/products/blob/7f3c88cad5c0b0cdb55b1fd6c7608b7d83cb9ef4/packages/skill-lesson/trpc/routers/pricing.ts#L212).

This is going to help simplify some coupon logic and clean up optional values in a couple types that are used all over the monorepo.

cleanup


joelhooks (approved):
🚮

PR #1381, opened 2024-02-02
cleanup: remove unused loadPrices API

jbranchaud:
We get pricing info from the API via the trpc `pricing.formatted`
endpoint, so there is no need for the `loadPrices` skill-api endpoint
anymore.

Since ejecting Testing Accessibility from the monorepo, this endpoint is no longer used. It can be safely removed.

Note: there is now only a single place where `formatPricesForProduct` is called and that is in the trpc `pricing` router.

CleanShot 2024-02-02 at 14 47 48@2x

succession


joelhooks (approved):
🧹

PR #1385, opened 2024-02-06
feat: add endpoint for requesting user magic link

jbranchaud:
This PR adds an internal-use (authenticated against `SKILL_SECRET`) endpoint for requesting a magic link on behalf of a user. It makes use of the existing Verification URL logic that hooks into NextAuth.

Example of calling this `api/skill/create-magic-link` endpoint via a REST client with an `email` param and the `x-skill-secret` header:

CleanShot 2024-02-06 at 12 05 55@2x

magic link

### DIFF
--- a/packages/skill-api/src/core/services/create-magic-link.ts
+++ b/packages/skill-api/src/core/services/create-magic-link.ts
@@ -1,0 +1,84 @@
+import {z} from 'zod'
+import {IncomingRequest, OutgoingResponse} from '../index'
+import {SkillRecordingsHandlerParams} from '../types'
+import {createVerificationUrl} from '../../server'
+
+class AuthorizationError extends Error {
+  constructor(message: string) {
+    super(message)
+    this.name = 'ValidationError'
+  }
+}
+
+const requireSkillSecret = (req: IncomingRequest): void => {
+  try {
+    const verifySkillSecretHeader = z
+      .object({skillSecretInHeader: z.string(), skillSecret: z.string()})
+      .transform(({skillSecretInHeader, skillSecret}, ctx) => {
+        if (skillSecret !== skillSecretInHeader)
+          ctx.addIssue({
+            code: z.ZodIssueCode.custom,
+            message: 'Unauthorized request to skill-api',
+          })
+      })
+
+    verifySkillSecretHeader.parse({
+      skillSecretInHeader: req.headers['x-skill-secret'],
+      skillSecret: process.env.SKILL_SECRET,
+    })
+  } catch (error: any) {
+    if (error instanceof z.ZodError) {
+      throw new AuthorizationError(error.message)
+    }
+
+    throw error
+  }
+}
+
+export async function createMagicLink({
+  params,
+}: {
+  params: SkillRecordingsHandlerParams
+}): Promise<OutgoingResponse> {
+  try {
+    const {req} = params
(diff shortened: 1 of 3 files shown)


packages/skill-api/src/core/services/create-magic-link.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+export async function createMagicLink({
+  params,
+}: {
+  params: SkillRecordingsHandlerParams
+}): Promise<OutgoingResponse> {
+  try {
+    const {req} = params
+    requireSkillSecret(req)
```
jbranchaud, line 45: Verify skill-secret before continuing

packages/skill-api/src/core/services/create-magic-link.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    const email = z.string().parse(req.query.email)
+
+    const {nextAuthOptions} = params.options
+
+    const verificationDetails = await createVerificationUrl({
+      email,
+      nextAuthOptions,
+    })
```
jbranchaud, line 54: create a magic link based on a incoming `email` and the next auth options

packages/skill-api/src/core/services/create-magic-link.ts
```diff
@@ -0,0 +1,84 @@
...
+      },
+    }
+  } catch (error: any) {
+    if (error instanceof AuthorizationError) {
+      return {
+        status: 401,
+        body: {
+          error: 'Unauthorized',
+        },
+      }
```
jbranchaud, line 76: 401 response if authentication fails

packages/skill-api/src/router.ts
```diff
@@ -95,6 +97,8 @@ export async function actionRouter({
         return await transferPurchase({params})
       case 'refund':
         return await stripeRefund({params})
+      case 'create-magic-link':
+        return await createMagicLink({params})
```
jbranchaud, line 101: add `create-magic-link` route to skill-api

packages/skill-api/src/server/send-server-email.ts
```diff
@@ -20,22 +20,16 @@ function hashToken(token: string, options: any) {
   )
 }
 
-export async function sendServerEmail({
+export async function createVerificationUrl({
```
jbranchaud, line 23: Extracted the logic for building the verification URL from `sendServerEmail` into a `createVerificationUrl` function. That function then gets used by `sendServerEmail`, but can also be imported and used in the `createMagicLink` route.

packages/skill-api/src/core/services/create-magic-link.ts
```diff
@@ -0,0 +1,84 @@
...
+      skillSecret: process.env.SKILL_SECRET,
+    })
+  } catch (error: any) {
+    if (error instanceof z.ZodError) {
+      throw new AuthorizationError(error.message)
+    }
+
+    throw error
+  }
+}
```
jbranchaud, line 36: I modeled this skill-secret check after the ones already used for endpoints like `process-refund`. I pulled it out into its own function to make it a little more robust and tighten the types. I like where I got to with this and think it could become a shared function used across these other endpoints:

CleanShot 2024-02-06 at 12 18 24@2x

joelhooks (approved):

PR #1392, opened 2024-02-09
feat(er): use custom path builder for lesson list

jbranchaud:
This adds an optional `lessonPathBuilder` function to the root `Collection` component so that individual apps can override how those paths are built. This is useful if the path structure of lessons for a specific app differs from the norm. In the case of Epic React v1 we don't use section slugs in the lesson paths. Instead, regardless of whether a lesson is nested in a section or not, we want to render the path as `/workshops/[module-slug]/[lesson-slug]`. This makes overriding the default easy.

This is implemented against https://github.com/skillrecordings/products/pull/1365, but I'm opening as a separate PR to make it easier to review.

overdrive


DrShpongle:
@jbranchaud imo idea of optional path builder function is great. Works well on my side (for the first sight at least).

vojtaholik (approved):

joelhooks (approved):

PR #1396, opened 2024-02-13
jb/make learn page dynamic for user

jbranchaud:
This makes it so that the `/learn` page can load modules based on the user's purchased level of access. It is still faking the user lookup part because we haven't layered in auth yet, but this should be good enough for now.

- feat(er): use getServerSideProps on learn page
- cleanup(er): remove unused imports
- feat(er): adjust query so sectioned resources show

dino


apps/epic-react/src/pages/learn.tsx
```diff
@@ -1,23 +1,25 @@
...
 import Header from '@/components/app/header'
 import {getOgImage} from '@/utils/get-og-image'
 import {WorkshopSchema} from '@/lib/workshops'
 import React from 'react'
 
-export const getStaticProps: GetStaticProps = async (context) => {
-  const workshops = await getAllWorkshops()
+export const getServerSideProps: GetServerSideProps = async ({req, query}) => {
+  // TODO: load the user's purchases and figure out what product they should have access to
+  const productId = 'kcd_2b4f4080-4ff1-45e7-b825-7d0fff266e38'
```
jbranchaud, line 14: Faking the product ID lookup. Eventually we will look up the auth'd user's purchases and figure out what productId to use.

DrShpongle (approved):
👍

PR #1397, opened 2024-02-13
feat(er): load bonuses based on user's productId

jbranchaud:
@DrShpongle this wires up Epic React bonuses so that they are also loaded based on the purchased `productId`

bonus


DrShpongle (approved):
👍

PR #1405, opened 2024-02-16
Add PaymentOptions with a Stripe provider to skill-api

jbranchaud:
Here is my initial attempt at moving toward a multi-provider payment options approach for skill-api. I'm guessing a little at the right abstraction for this since we only have the one provider (Stripe) at the moment. But I've used NextAuth for inspiration for the patterns used here.

Besides continuing to explore what exactly the `PaymentOptions` scheme for skill-api should look like, other questions that will need to be explored are:

- there are parts of each app outside the configured skill-api endpoints that import stripe to do things like retrieve an invoice or transfer a purchase/customer. how do we plug the `PaymentOptions` config into those places?
- in that same vein, how do we plug the `PaymentOptions` config into other packages that make Stripe calls, e.g. trpc endpoints?
- how do we create a common-denominator interface for payment provider activities to support multiple providers, e.g. if I'm retrieving an invoice for a user, I need to determine what provider they used to pay and then make a request to that provider. I'm imagining a `skill-payments` package with generic functions like `getInvoiceDetails` that will look up a user, their purchase and merchant charge records, and then make the appropriate request.
- I still need to look at the LemonSqueezy docs to see what their API and webhooks look like. Curious to see how cleanly they will map onto our data model.

---

- feat: add PaymentOptions with Stripe provider
- feat: use configured stripe for rest of skill-api

options


apps/testing-javascript/src/pages/api/skill/[...skillRecordings].ts
```diff
@@ -1,14 +1,24 @@
...
+  StripeProvider,
 } from '@skillrecordings/skill-api'
 import {nextAuthOptions} from '../auth/[...nextauth]'
 
+const paymentOptions = defaultPaymentOptions({
+  stripeProvider: StripeProvider({
+    stripeSecretKey: process.env.STRIPE_SECRET_TOKEN,
+    apiVersion: '2020-08-27',
+  }),
+})
```
Comment (jbranchaud) [inline apps/testing-javascript/src/pages/api/skill/[...skillRecordings].ts:13]: as an example, this is TJS's skill-api endpoint configuring `PaymentOptions` with a `StripeProvider` instance which will give skill-api access to a configured Stripe client

if the Stripe client is present in the `PaymentOptions`, then stripe-based endpoints will work, otherwise they will 500.

packages/skill-api/src/core/services/process-refund.ts
```diff
@@ -24,6 +29,12 @@ export async function stripeRefund({
       }
     }
 
+    const stripe = paymentOptions?.stripeCtx.stripe || defaultStripe
```
jbranchaud, line 32: these various skill-api endpoints are now defaulting to the configured stripe client from `PaymentOptions`, but since not all the apps are configuring that yet, I'm falling back to a manual import of the `stripe-sdk` stripe client

packages/skill-api/src/server/default-payment-options.ts
```diff
@@ -0,0 +1,53 @@
...
+}) => PaypalProvider
+
+type PaymentProviderOptions = StripeProvider | PaypalProvider
+
+export type PaymentOptions = {
+  providers: {
+    stripe?: StripeProvider
+    paypal?: PaypalProvider
+  }
+}
```
jbranchaud, line 24: this file is mostly exploratory "writing on a whiteboard" -- `PaypalProvider` will probably need to be more appropriately named `LemonSqueezyProvider` for example.

I'm not married to any of these types, just trying to get the big picture sketched out in a way where the code also works.

joelhooks: related in a few ways, but i like this example of creating types to describe adapters https://github.com/treehousedev/treehouse/blob/main/lib/backend/mod.ts

packages/skill-api/src/server/default-payment-options.ts
```diff
@@ -0,0 +1,53 @@
...
+// define providers here for now,
+// but eventually they can go in a `providers/` directory
+export const StripeProvider: StripeProviderFunction = (config) => {
+  const stripeClient = new Stripe(config.stripeSecretKey, {
+    apiVersion: config.apiVersion,
+  })
+
+  return {
+    name: 'stripe',
+    paymentClient: stripeClient,
```
joelhooks, line 35: thinking: instead of `paymentClient` pointing to stripe, it gives us an interface that is wrapping a specific client, but would remain the same regardless of the actual client being used under the hood

https://github.com/skillrecordings/products/blob/902c94bb1b7a21879fb2fd206c06146d8fa98935/packages/skill-api/src/core/services/transfer-purchase.ts#L113C1-L118C8

```
const updateStripeCostumer = await stripe.customers.update(
  updateMerchantCustomer.identifier,
  {
    email: targetUserEmail,
  },
)
```

becomes 

```
const updatedCustomer = await paymentClient.updateCustomer(
  updateMerchantCustomer.identifier,
  {
    email: targetUserEmail,
  },
)
```

this is goofy way to capture it, but here's all the current touch points i could find that hit `stripe.*`

2024-02-20 at 12 15 21@2x

2024-02-20 at 12 15 51@2x

2024-02-20 at 12 16 30@2x

2024-02-20 at 12 16 49@2x

2024-02-20 at 12 17 01@2x

PR #1416, opened 2024-02-22
fix(er-next): add completionToggle to the exercise type of the lessons

DrShpongle:
CleanShot 2024-02-22 at 18 05 20@2x

### DIFF
--- a/apps/epic-react/src/templates/lesson-template.tsx
+++ b/apps/epic-react/src/templates/lesson-template.tsx
@@ -84,6 +84,7 @@ const ExerciseTemplate: React.FC<{
   const displayLessonCompletionToggle =
     (lesson._type === 'solution' ||
       lesson._type === 'explainer' ||
+      lesson._type === 'exercise' ||
       lesson._type === 'lesson' ||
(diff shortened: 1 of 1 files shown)


apps/epic-react/src/templates/lesson-template.tsx
```diff
@@ -84,6 +84,7 @@ const ExerciseTemplate: React.FC<{
   const displayLessonCompletionToggle =
     (lesson._type === 'solution' ||
       lesson._type === 'explainer' ||
+      lesson._type === 'exercise' ||
```
DrShpongle, line 87: The lessons wrapped in sections have `exercise` type (excluding `intro`/`outro` that have `explainer` type) so I've added this type to condition and now completion toggle is shown.

jbranchaud: I think we want them to be able to "toggle complete" any lesson type. Does this cover all lesson types?

DrShpongle: I believe yes

jbranchaud (approved):

PR #1417, opened 2024-02-22
feat(commerce): make record new purchase idempotent

jbranchaud:
Check for existing MerchantCharge and Purchase records before trying to
create them and the other associated records. This will help avoid
unique key errors and duplicate records.

One caveat: depending on where the webhook processing errored out the initial time, it is possible that we trigger the `sendServerEmail` an additional time. I wonder if there is a way to move this call into inngest which can track whether the email was sent and then prevent duplicate email sends.

This PR is in response to noticing a few failing `checkout.session.completed` events in stripe:

image

charge


joelhooks (approved):
@jbranchaud inngest has idempotent config option that guarantees a single execution in a 24 hour period and would be perfect.

PR #1424, opened 2024-02-29
fix(TJS-61): testing javascript course completion certificate

DrShpongle:


apps/testing-javascript/src/pages/api/skill/[...skillRecordings].ts
```diff
@@ -1,24 +1,24 @@
 import SkillRecordings, {
   type SkillRecordingsOptions,
-  defaultPaymentOptions,
-  StripeProvider,
+  // defaultPaymentOptions,
+  // StripeProvider,
```
Comment (DrShpongle) [inline apps/testing-javascript/src/pages/api/skill/[...skillRecordings].ts:4]: @jbranchaud is it appropriate to remove these broken imports in here?

PR #1427, opened 2024-03-07
jbranchaud/tjs 62 Make EW stripe webhook handler deal with TJS stripe events

jbranchaud:
_Note: I've tested this locally against TT, EW and TJS. All looks good._

This PR makes some updates to Stripe webhook handling in the skill-api. The `epic-web` app receives stripe webhook events for all kcd-products. For the events that are for one of the other kcd-products (e.g. testing-javascript), they will be forwarded along to that app so that it can process them in the proper context (that context is important for things like sending emails).

We know what app a webhook is meant for because we include the `siteName` in the webhook metadata. At this point, this is mainly relevant for `customer.updated` and `checkout.session.created`.

- **feat: add app name to env var type**
- **feat: epic-web forwards stripe events to tjs**

handler


joelhooks (approved):
🙏🏻

PR #1432, opened 2024-03-08
feat: compare previous email on customer updated

jbranchaud:
Previously we were comparing two emails that were always identical
because the user tied to the Merchant Customer record had already been
updated. Because of this, an email was never getting sent to the new
user.

compare

### DIFF
--- a/packages/skill-api/src/core/services/process-stripe-webhook.ts
+++ b/packages/skill-api/src/core/services/process-stripe-webhook.ts
@@ -285,25 +285,40 @@ export const processStripeWebhook = async (
       },
     })
 
-    const user = merchantCustomer?.user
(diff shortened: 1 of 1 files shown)


packages/skill-api/src/core/services/process-stripe-webhook.ts
```diff
@@ -285,25 +285,40 @@ export const processStripeWebhook = async (
...
+        })
+        .parse(eventObject)
+
+      const {previous_attributes} = z
+        .object({
+          previous_attributes: z
+            .object({email: z.string().optional()})
+            .optional(),
+        })
+        .parse(event.data)
```
jbranchaud, line 304: We can pull the previous email out of `previous_attributes` that stripe includes in its webhook payload.

PR #1458, opened 2024-03-19
feat: script to add and remove SRT track from Mux

jbranchaud:
Adding a WEBVTT subtitles track to an exist Mux asset took my a little trial and error because I didn't have the format right to begin with. These two scripts made that process easier.

- [Create Asset Track](https://docs.mux.com/api-reference#video/operation/create-asset-track)
- [Delete Asset Track](https://docs.mux.com/api-reference#video/operation/delete-asset-track)

after party


apps/testing-javascript/src/data/add-srt-to-specific-video.ts
```diff
@@ -0,0 +1,50 @@
...
+import Mux from '@mux/mux-node'
+
+require('dotenv-flow').config({
+  default_node_env: 'development',
+})
+
+const assetId = undefined // Set me!
+const srtUrl = undefined // Set me!
+
+const addSrtToSpecificVideo = async (
```
joelhooks, line 22: personally I found it useful to check here if it already exists and remove it first

joelhooks (approved):

PR #1473, opened 2024-03-26
jbranchaud/sks 91 set up payment options for epic web

jbranchaud:
- **feat(ew): add PaymentOptions, remove stripeData**
- **feat(ew): remove last Stripe import**
- **feat(ew): use payment options in inngest funcs**

rain


apps/epic-web/src/inngest/functions/sanity/product/sanity-product-created.ts
```diff
@@ -15,6 +15,10 @@ export const sanityProductCreated = inngest.createFunction(
     if: 'event.data.event == "product.create"',
   },
   async ({event, step}) => {
+    if (!stripe) {
+      throw new Error('Payment provider (Stripe) is missing')
+    }
```
jbranchaud, line 20: @joelhooks does this make sense as a way to handle the "Stripe provider not defined" scenario? This is mostly to let TS know that it definitely exists below since it has to get defined in the API setup for the app to work.

Or is there a more idiomatic inngest way of bailing if the payment provider isn't configured?

joelhooks: it will keep retrying. there's a specific error if you want it to stop trying

joelhooks: middlware is also cool with inngest: https://github.com/badass-courses/course-builder/blob/852fb36dd687bfe228fc9c10209bc529aaf6f4dd/apps/course-builder-web/src/inngest/inngest.server.ts#L64-L73

jbranchaud: > it will keep retrying. there's a specific error if you want it to stop trying

nice, I updated it to use `NonRetriableError` 🙌 https://www.inngest.com/docs/guides/error-handling#preventing-retries-with-non-retriable-errors

PR #1477, opened 2024-03-28
jbranchaud/sks 90 replace remaining calls to stripedata with getpurchaseinfo

jbranchaud:
This removes remaining calls to `stripeData` (and removes `stripeData` itself) and moves the `purchaseType` logic up into the `getPurchaseInfo` function so that we don't have to do a separate function call that fires off excess network requests.

- **feat: extract stripe-specific function**
- **feat: pull stripe pieces up into outer method**
- **feat: include purchase type info in info**
- **feat: fully replace determinePurchaseTypeViaStripe**
- **feat: finally remove stripeData function**

replace


jbranchaud:
This is weird, all I did was push a commit that removes a comment. All checks passed before.

CleanShot 2024-03-29 at 10 35 28@2x

joelhooks (approved):

PR #1488, opened 2024-04-10
feat(er): Correctly navigate to interviews in bonus section

zacjones93:
The routes generated for the bonus interview section were `/workshop` where the `LessonList` component dynamically builds the links based on module type which point to `/bonuses`. 

This extracts bonuses into their own routes and correctly navigates to those pages on the `/learn` page.

Also adjusted the Module titles on the `/learn` page to navigate to the first lesson in that module because that's how Epic React does it.

g bonus unlocked


packages/skill-lesson/video/video-overlays.tsx
```diff
@@ -652,12 +653,14 @@ const BuyProduct: React.FC<{product?: SanityProduct}> = ({product}) => {
...
-                          height={60}
-                        />
+                        {(module.image?.url || module.image) && (
+                          <Image
+                            src={module.image?.url ?? module.image}
+                            alt={`${module.title} workshop`}
+                            width={60}
+                            height={60}
+                          />
+                        )}
```
zacjones93, line 663: I likely didn't have set up working correctly so data wasn't there but added a check for image url as a fail-safe. Was running into rendering errors getting started

apps/epic-react/README.md
```diff
@@ -58,7 +58,13 @@ Copy the template `.env.local.template` file to `.env.local` and `.env.template`
...
 
 ```shell
-DATABASE_URL="mysql://root@localhost:3309/epic-react"
+DATABASE_URL="mysql://root@localhost:3309/kcd-products"
+```
+
+To create the Prisma client run:
+
+```shell
+pnpm -w generate
```
jbranchaud, line 67: @zacjones93 thanks for capturing these setup updates!

packages/skill-lesson/video/video-overlays.tsx
```diff
@@ -499,18 +499,19 @@ const BlockedOverlay: React.FC<{
...
-      <Image
-        data-thumbnail=""
-        src={thumbnail}
-        fill
-        alt=""
-        aria-hidden="true"
-        priority
-      />
-      {module.moduleType === 'tutorial' ? (
+      {thumbnail && (
```
jbranchaud, line 504: looks like thumbnail should always be defined on 479

apps/epic-react/src/pages/learn.tsx
```diff
@@ -31,14 +31,18 @@ const ResourceLink: React.FC<{
   workshopSlug: string
   resourceSlug: string
   isCompleted: boolean
-}> = ({title, workshopSlug, resourceSlug, isCompleted}) => {
+  isBonusModule?: boolean
+}> = ({title, workshopSlug, resourceSlug, isCompleted, isBonusModule}) => {
   const [isHovered, setHovered] = React.useState<Boolean>(false)
+  const resourceLink = isBonusModule
+    ? `/bonuses/${workshopSlug}/${resourceSlug}`
+    : `/workshops/${workshopSlug}/${resourceSlug}`
```
DrShpongle, line 39: Both bonuses and workshops should be "/modules/" eventually to be consistent with original ER:
CleanShot 2024-04-10 at 23 58 01@2x

apps/epic-react/src/pages/workshops/[module]/[lesson]/index.tsx
```diff
@@ -6,21 +6,15 @@ import {VideoResourceProvider} from '@skillrecordings/skill-lesson/hooks/use-vid
...
-import {getAllBonuses, getBonus} from '@/lib/bonuses'
 import {serialize} from 'next-mdx-remote/serialize'
 
 export const getStaticProps: GetStaticProps = async (context) => {
   const {params} = context
   const lessonSlug = params?.lesson as string
   const sectionSlug = params?.section as string
   const moduleSlug = params?.module as string
-  const isBonusModule = moduleSlug === 'epic-react-expert-interviews'
-
```
Comment (DrShpongle) [inline apps/epic-react/src/pages/workshops/[module]/[lesson]/index.tsx:20]: @zacjones93 Both bonuses and workshops lessons are located by `/modules/[module-slug]/[lesson-slug]` address at original EpicReact. It is why I get both types of resources in the same template (Lesson template). Only the thing that needs to be done is replace `/workshops/` with `/modules/` to be consistent with original ER. So I don't think it is good idea to have separate dynamic route `bonuses/[bonus]/[interview]` because it's unclear how to keep the same url as workshops have. What do you think @jbranchaud ?

Comment (zacjones93) [reply apps/epic-react/src/pages/workshops/[module]/[lesson]/index.tsx:20]: Ah ok, that makes sense. I misunderstood the requirement there

Comment (jbranchaud) [reply apps/epic-react/src/pages/workshops/[module]/[lesson]/index.tsx:20]: @DrShpongle yeah, we want parity on the URLs for sure. We could always create framework-level redirects in the `next.config.js`, but probably easier to match URLs now with the file structure.

apps/epic-react/src/pages/modules/[module]/[lesson]/index.tsx
```diff
@@ -6,18 +6,17 @@ import {VideoResourceProvider} from '@skillrecordings/skill-lesson/hooks/use-vid
...
 import {getSection} from '@/lib/sections'
-import serializeMDX from '@skillrecordings/skill-lesson/markdown/serialize-mdx'
 import {getAllWorkshops, getWorkshop} from '@/lib/workshops'
-import {getAllBonuses, getBonus} from '@/lib/bonuses'
 import {serialize} from 'next-mdx-remote/serialize'
+import {getAllBonuses, getBonus} from '@/lib/bonuses'
 
 export const getStaticProps: GetStaticProps = async (context) => {
   const {params} = context
+  console.log({params})
```
Comment (zacjones93) [inline apps/epic-react/src/pages/modules/[module]/[lesson]/index.tsx:15]:
```suggestion
```

apps/epic-react/src/pages/modules/[module]/[lesson]/index.tsx
```diff
@@ -71,6 +70,7 @@ export const getStaticPaths: GetStaticPaths = async (context) => {
     )
   })
 
+  console.log({paths})
```
Comment (zacjones93) [inline apps/epic-react/src/pages/modules/[module]/[lesson]/index.tsx:73]:
```suggestion
```

apps/epic-react/src/pages/modules/[module]/[lesson]/index.tsx
```diff
@@ -83,6 +83,15 @@ const ExercisePage: React.FC<any> = ({
...
 }) => {
+  console.log({
+    lesson,
+    lessonBodySerialized,
+    lessonBodyPreviewSerialized,
+    module,
+    section,
+    transcript,
+    videoResourceId,
+  })
```
Comment (zacjones93) [inline apps/epic-react/src/pages/modules/[module]/[lesson]/index.tsx:94]:
```suggestion
```

apps/epic-react/src/templates/lesson-template.tsx
```diff
@@ -90,6 +90,8 @@ const ExerciseTemplate: React.FC<{
       lesson._type === 'interview') &&
     session
 
+  const epicReactModule = {...module, moduleType: 'module'}
```
zacjones93, line 93:
```suggestion
// epic react modules are always type 'module'
  const epicReactModule = {...module, moduleType: 'module'}
```

because the lesson list is data driven, this is the only way to change the url without either changing the source code in `Collection` that uses `lessonPathBuilder` to build the dynamic path. 

This seems cleaner to me than redefining the `lessonPathBuilder`

DrShpongle (approved):

PR #1493, opened 2024-04-16
add episodeId to dependency array to update podcast player url

zacjones93:
gif

## Correct behavior

https://github.com/skillrecordings/products/assets/6188161/ad569403-7be9-4491-8efe-c24bc31666b7

## Broken behavior

https://github.com/skillrecordings/products/assets/6188161/06072c43-2c72-4339-865d-b08146177725


apps/epic-react/src/components/podcasts/podcast-player.tsx
```diff
@@ -20,7 +21,7 @@ const PodcastPlayer = ({episodeId}: {episodeId: string}) => {
         setState({loading: false}),
       )
     }
-  }, [theme])
+  }, [theme, episodeId])
```
zacjones93, line 24: the only time we set the iframe src is in this useEffect (which is stored in a ref) so we need to include `episodeId` in the dependency array to reassign the src on navigation

DrShpongle (approved):
Good catch! 🌮

PR #1494, opened 2024-04-16
jbranchaud/er 34 only pro pricing tier should include bulk purchasing option

jbranchaud:
- **feat: make selling live in dev**
- **feat: allow config options to be passed in**
- **feat: each pricing gets its own price provider**

The Buy page pricing components now differentiate themselves from one another. This does two things for us:
1. the prices are independent so that toggling on bulk quantity for one does load bulk quantity for other prices
2. the basic and standard products don't even have a bulk purchase options (parity with existing ER site)

CleanShot 2024-04-16 at 17 44 07@2x

pricing


DrShpongle (commented):
Works as expected on my side. Since files used by other apps are involved in this PR I've tested it with TT and EW locally. Do we need more tests before merge this?

DrShpongle (approved):
🌮

PR #1496, opened 2024-04-17
feat: adds productType in schema to allow valid live event purchases to have PPP for a selfpaced purchase

nicollguarnizo:
separate


packages/database/prisma/schema.prisma
```diff
@@ -180,6 +180,7 @@ model Price {
 model Product {
   id               String               @id @default(uuid())
   name             String
+  productType      String
```
jbranchaud, line 183: This is going to have to start as `productType    String?` in order to be able to migrate against a database that already has records in the `Products` table. Without the `?` on `String?`, the column is `not null`. That non-nullable constraint will be violated by the migration and it will be forced to rollback.

A good way of thinking about how to apply this kinds of schema changes is [the Expand and Contract pattern](https://www.prisma.io/dataguide/types/relational/expand-and-contract-pattern#:~:text=various%20database%20transformations.-,What%20is%20the%20expand%20and%20contract%20pattern%3F,data%20structure%20without%20affecting%20uptime.).

Typically that means the migration will have to be multi-step:
1. Add new nullable field and deploy
2. Update any _write_ app code to apply some param or default value when creating `Product` records
3. Backfill existing records with whatever data should go in that field
4. Update schema to make the field non-nullable and deploy

Another way to approach this would be to keep it non-nullable and give it a default value of `self-paced`.

Did you run in to this non-nullable migration rollback with your development database while implementing this?

apps/colt-steele/seed_data/01.sql
```diff
@@ -160,6 +160,7 @@ CREATE TABLE `Price` (
 CREATE TABLE `Product` (
     `id` VARCHAR(191) NOT NULL,
     `name` VARCHAR(191) NOT NULL,
+    `productType` VARCHAR(191) NOT NULL,
```
jbranchaud, line 163: a few of these `apps/<app-name>/seed_data/01.sql` files have `NOT NULL` on the `productType` which doesn't match the prisma schema

apps/epic-web/src/inngest/functions/sanity/product/index.ts
```diff
@@ -43,6 +44,7 @@ export const BaseSanityProductSchema = z.object({
   title: z.string(),
   slug: z.string(),
   state: z.enum(['draft', 'active', 'unavailable']).default('draft'),
+  type: z.enum(['self-paced', 'live']).default('self-paced'),
```
jbranchaud, line 47: this zod schema enum has `live`, but the test mock has `live-event` -- which value should it be? Might be worth having a type or constant that these can import for consistency.

packages/commerce-server/src/format-prices-for-product.test.ts
```diff
@@ -484,34 +484,161 @@ test('PPP can be forced to not auto-apply for upgrade', async () => {
...
     productId: DEFAULT_PRODUCT_ID,
-    userId: 'default-user',
+    userId: 'user-1',
+    quantity: 1,
     country: 'IN',
     ctx,
   })
 
-  expect(product.availableCoupons.length).toBe(0)
+  expect(availableCoupons.length).toBe(1)
```
jbranchaud, line 641: this assertion could be strengthened to something like:

```suggestion
  expect(availableCoupons[0].type).toBe('ppp')
```

jbranchaud (approved):

PR #1504, opened 2024-04-23
feat(ER): create module completion certificate on module finish

zacjones93:
This PR sets up the 'complete and continue' button that is currently in Epic React production.

When the module is completed, users are redirected to `[module]/completed` route where they can download their certificate.

https://www.loom.com/share/4b280a60a9fd4bd2955d0300aa093908

g check


DrShpongle (approved):
💥

PR #1507, opened 2024-04-23
(Epic Web) Adds more structured data to workshops

Creeland:
## Problem

Pricing information is not being shown in our search results. 

Workshops are also not being considered "courses" by Google. So they aren't being displayed in the courses carousel.

image
image

## Solution

Add additional structured data to the `CourseJsonLd` components in workshop-template and buy

molecule


apps/epic-web/src/templates/workshop-template.tsx
```diff
@@ -432,21 +432,34 @@ const Header: React.FC<{module: Module; canView: boolean}> = ({
...
-}: {
-  title: string
-  description?: string | null | undefined
-}) => (
+const CourseMeta = ({product}: {product: SanityProduct}) => (
   <CourseJsonLd
-    courseName={title}
-    description={description || ''}
+    courseName={product.title || ''}
+    description={product.description || ''}
```
zacjones93, line 438: we should beef these up. Crawlers expect certain char lengths for title and description, this goes for where ever normal SEO is being set as well (could be a different PR?) 

image

apps/epic-web/src/pages/buy.tsx
```diff
@@ -86,6 +89,55 @@ const BuyPage: React.FC<
...
+        <CourseJsonLd
+          courseName="Full Stack Vol 1"
+          description="The most comprehensive guide to professional web development by Kent C. Dodds."
+          provider={{
+            name: 'Kent C. Dodds',
+            type: 'Person',
+            url: isBrowser()
+              ? document.location.href
+              : process.env.NEXT_PUBLIC_URL,
+          }}
```
zacjones93, line 101: `sameAs` is being passed by default which we can add kents socials here and personal site. This could be 

```json
{
  sameAs: ['https://kentcdodds.com', 'https://twitter.com/kentcdodds', 'https://github.com/kentcdodds'] 
}
```

https://schema.org/sameAs

image

Creeland: Unfortunately, the provider prop doesn't support sameAs. I think that's the intended use of the URL prop. But, you can't pass an array to it. 

I'll set Kent's personal site as that URL.

zacjones93 (approved):
looks great

PR #1509, opened 2024-04-24
[epic-web] Pluralize "spot" string for `availability?.quantityAvailable`

brandonmcconnell:
attn @kentcdodds
—

Fix plural typo when only 1 quantity is present

I couldn't manage to spin up the repo locally even after installing all the deps, but here is a proof of concept:


brandonmcconnell:
I'm unable to view the error reports, but if someone tells me what is failing, I can investigate it if needed.

vojtaholik (approved):
thanks @brandonmcconnell 🙌

PR #1528, opened 2024-05-10
feat(tt): look up section slug for module/lesson

jbranchaud:
there is no `sectionSlug` in the URL path params, so we were always getting `null` when trying to look up the `section`.

this adds a reverse lookup of the `section` based on the lesson's slug and the list of `resources` on the `module`

lookup


DrShpongle (approved):
Thanks, Josh 🙌

PR #1535, opened 2024-05-20
feat: add types and organize logic flow

jbranchaud:
there were a couple paths through this `tagSubscriber` function where
things were missing, but functions were still executed with the missing
data.

e.g. without a purchased ID we don't have a subscriber, so why not exit
early.

without a subscriber, we cannot tag the subscriber or add fields

field


packages/convertkit-sdk/src/convertkit.ts
```diff
@@ -101,13 +111,17 @@ export async function tagSubscriber(email: string, tagId: string) {
...
   })
     .then((res) => res.json())
-    .then(({subscription}: any) => {
-      return subscription.subscriber
+    .then((jsonRes: any) => {
+      const result = TagSubscriberResponseSchema.safeParse(jsonRes)
+      if (!result.success) {
+        return undefined
+      }
+      return result.data.subscription.subscriber
```
jbranchaud, line 119: use Zod to validate the exact data we want from this function. the types flow where this gets used which made it easier to notice a logic inconsistency.

packages/skill-api/src/core/services/convertkit/convertkit-tag-purchase.ts
```diff
@@ -46,57 +44,62 @@ async function getSanityProduct(productId: string) {
 
 export async function convertkitTagPurchase(email: string, purchase: Purchase) {
   // skip CK tagging in development
-  if (process.env.SKIP_CK_TAGGING === 'true') return
+  if (process.env.SKIP_CK_TAGGING === 'true')
+    return {tagged: false, reason: 'SKIP_CK_TAGGING is true'}
```
jbranchaud, line 48: there were a mix of boolean and `{ status: 200 }` return types from this function, trying to unifying into something more consistent.

(though it doesn't look like the call site is using the return type)

packages/skill-api/src/core/services/convertkit/convertkit-tag-purchase.ts
```diff
@@ -46,57 +44,62 @@ async function getSanityProduct(productId: string) {
...
+        })
+      }
 
-    if (process.env.CONVERTKIT_PURCHASED_TAG_ID) {
-      subscriber = await tagSubscriber(email, convertkitPurchasedTagId)
+      // no tagging is going to happen (unless we want to lookup by email), so return early
+      return {
+        tagged: false,
+        reason: 'convertkitPurchasedTagId is not set',
+      }
```
jbranchaud, line 71: if we don't have a purchased tag id, then post to slack (in production) and then return early.

unless we want to add new logic to look up the subscriber by `email` and still try to do the field tagging with that?

packages/skill-api/src/core/services/convertkit/convertkit-tag-purchase.ts
```diff
@@ -46,57 +44,62 @@ async function getSanityProduct(productId: string) {
...
 
-    if (!convertkitPurchasedTagId && process.env.NODE_ENV === 'production') {
-      await postToSlack({
-        webClient: new WebClient(process.env.SLACK_TOKEN),
-        channel: process.env.SLACK_ANNOUNCE_CHANNEL_ID,
-        text: `process.env.CONVERTKIT_PURCHASED_TAG_ID is not set for ${purchase.productId}`,
-        attachments: [],
-      })
-    }
+    let subscriber = await tagSubscriber(email, convertkitPurchasedTagId)
```
jbranchaud, line 74: tag the subscriber could be collapsed to this single line with the guard/early-return above

packages/skill-api/src/core/services/convertkit/convertkit-tag-purchase.ts
```diff
@@ -46,57 +44,62 @@ async function getSanityProduct(productId: string) {
...
+      const purchasedOnFieldName = sanityProduct
+        ? `purchased_${sanityProduct.slug.current.replace(/-/gi, '_')}_on`
+        : process.env.CONVERTKIT_PURCHASED_ON_FIELD_NAME || 'purchased_on'
 
-    await setConvertkitSubscriberFields(subscriber, {
-      [purchasedOnFieldName]: format(new Date(), 'yyyy-mm-dd HH:MM'),
-    })
+      await setConvertkitSubscriberFields(subscriber, {
+        [purchasedOnFieldName]: format(new Date(), 'yyyy-mm-dd HH:MM'),
+      })
```
jbranchaud, line 83: only try to set subscriber fields if we have a subscriber object

joelhooks (approved):

PR #1556, opened 2024-06-07
Added automatic type acquisition to shiki plugin

mattpocock:
This PR:

- Removes the dependency on shiki-service
- Moves all shiki-twoslash rendering to the build script/rebuild script

If this works, we can close shiki-service for good - and good riddance.

The reason we had shiki-service in the first place was to solve a problem: by default, twoslash looks at the file system to acquire its types. So, `typescript`, `@types/react` etc all need to be _on the file system_ when twoslash transforms the code. This was not a problem on initial build, but on subsequent builds (due to a Sanity update) the incremental rebuild didn't have the entire FS loaded.

shiki-service was built as a persistent server so that this file system would always be in place.

This PR makes it so that twoslash looks to a CDN for all of its types, including TS. This means it can run on incremental rebuilds.

It should also be significantly faster, and not rely on a memory-hogging external server we need to pay for.

### DIFF
--- a/packages/skill-lesson/markdown/serialize-mdx.ts
+++ b/packages/skill-lesson/markdown/serialize-mdx.ts
@@ -5,8 +5,8 @@ import {serialize} from 'next-mdx-remote/serialize'
 import rehypeRaw from 'rehype-raw'
 import {
   ShikiRemotePluginOptions,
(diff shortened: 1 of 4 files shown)


vojtaholik (approved):
⭐

PR #1571, opened 2024-06-21
Attempted to fix a twoslash issue

mattpocock:
Exercise 2: Optional Property Types on Chapter 4 is buggy - twoslash is erroring out for some reason.

image

This PR is an attempt to fix it.


mattpocock:
Closed, fixed via a content change instead.

PR #1574, opened 2024-06-21
feat(tt): Track bookmarks with database through tRPC

zacjones93:
There is talk about turning bookmarks into a pro feature for total typescript and this PR would set us up for doing that if we wanted to.

This swaps the Total TypeScript Essentials bookmarking feature from using `IndexedDB` to our actual database. Instead of using a hook and react-query this uses tRPC to manage interactions with the database. Behind the scenes we have helper functions running CRUD through Prisma. 

If these changes look good I will need to open up merge requests on all the products in planetscale.

gif


apps/total-typescript/src/lib/bookmarks.ts
```diff
@@ -0,0 +1,164 @@
...
+export type BookmarkInput = z.infer<typeof BookmarkInputSchema>
+
+export async function getBookmarksForUser(userId: string) {
+  try {
+    const bookmarks = await prisma.bookmark.findMany({
+      select: {
+        id: true,
+        module: true,
+        section: true,
+        resource: true,
```
zacjones93, line 40: we don't need to return `userId` or when bookmark was created

apps/total-typescript/src/lib/bookmarks.ts
```diff
@@ -0,0 +1,164 @@
...
+      },
+      where: {
+        userId,
+        resource: {
+          path: '$.id',
+          equals: resourceId,
+        },
```
zacjones93, line 73: filtering on JSON values is a little weird but this is what prisma docs recommend https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types/working-with-json-fields#filter-on-a-json-field-advanced

apps/total-typescript/src/pages/books/[book]/[chapter]/index.tsx
```diff
@@ -170,20 +176,14 @@ const BookChapterRoute: React.FC<{
...
-      .then(() => {
-        toast.success('Bookmark added')
-      })
+    addBookmarkMutation.mutate({
+      module: book.slug.current,
+      section: {
+        title: chapter.title,
+        slug: chapter.slug,
+      },
+      resource: heading,
```
Comment (zacjones93) [inline apps/total-typescript/src/pages/books/[book]/[chapter]/index.tsx:185]: prisma handles `createdOn` attribute (now `createdAt`)

apps/total-typescript/src/trpc/routers/bookmarks.ts
```diff
@@ -0,0 +1,84 @@
...
+      if (!token) {
+        throw new Error('Unauthorized')
+      }
+
+      const bookmark = await getBookmarkByResourceId({
+        userId: token.id as string,
+        resourceId: input.id,
+      })
+
+      return ReturnedBookmarkSchema.parse(bookmark)
```
zacjones93, line 28: parse the returned bookmark with zod for friendlier types. Prisma returns a Prisma.JsonValue for section and resource which isn't great to work with

vojtaholik (approved):
LGTM

PR #1577, opened 2024-06-24
jbranchaud/er 90 basic price tier can be purchased if higher bundle is

jbranchaud:
- **fix(er): wrap all prices with Price Check Provider**
- **cleanup: made isDowngrade logic clearer**

The main thing this PR does is to make lower-tier options show up as _Unavailable_ if you have already purchased one of the higher tiers.

CleanShot 2024-06-24 at 11 16 43@2x

tier

### DIFF
--- a/packages/skill-lesson/path-to-purchase/product-tiers.tsx
+++ b/packages/skill-lesson/path-to-purchase/product-tiers.tsx
@@ -28,17 +28,15 @@ export const PricingTiers: React.FC<
     couponIdFromCoupon || (validCoupon ? couponFromCode?.id : undefined)
 
   const purchasedProductIds = purchases.map((purchase) => purchase.productId)
+
   return (
     <>
       {redeemableCoupon ? <RedeemDialogForCoupon /> : null}
       <div data-pricing-container="">
-        {products?.map((productWithOptions, i) => {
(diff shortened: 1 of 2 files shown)


packages/skill-lesson/path-to-purchase/product-tiers.tsx
```diff
@@ -28,17 +28,15 @@ export const PricingTiers: React.FC<
...
       <div data-pricing-container="">
-        {products?.map((productWithOptions, i) => {
-          const {options, ...product} = productWithOptions
-          return (
-            <PriceCheckProvider
-              key={productWithOptions.slug}
-              purchasedProductIds={purchasedProductIds}
-            >
+        <PriceCheckProvider purchasedProductIds={purchasedProductIds}>
+          {products?.map((productWithOptions, i) => {
```
jbranchaud, line 37: the Price Check provider needs to be _outside_ the `products.map` so that all 3 of the `<Pricing>` components get wrapped in a single provider. Otherwise we have three separate `PriceCheckProvider` instances each of which only knows about the price it is wrapping and is not able to identifier downgrades.

DrShpongle (approved):

joelhooks (commented):
this should be ported over to course builder as well most likely

PR #1581, opened 2024-06-25
feat: replace deprecated deskTool with structure

jbranchaud:
[Desk is now Structure](https://www.sanity.io/help/desk-is-now-structure)

desk


vojtaholik (approved):
that's it?!?! 😂 awesome

jbranchaud:
> that's it?!?! 😂 awesome

this did not fix the Vision tool issue unfortunately. I just noticed that _Desk_ was deprecated and the upgrade path was easy.

PR #1582, opened 2024-06-25
fix(ew): EW-136 Solution Embed Fails with API error

joelhooks:
fail

### DIFF
--- a/apps/epic-web/src/templates/embed-template.tsx
+++ b/apps/epic-web/src/templates/embed-template.tsx
@@ -40,7 +40,7 @@ export type VideoEmbedPageProps = {
   convertkitSubscriber?: Subscriber
   abilityRules: any
   isSolution?: boolean
-  solution?: Lesson
+  solution?: any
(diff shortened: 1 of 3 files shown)


apps/epic-web/src/templates/embed-template.tsx
```diff
@@ -77,7 +81,11 @@ const EmbedTemplate: React.FC<VideoEmbedPageProps> = ({
         }}
       />
       <ModuleProgressProvider moduleSlug={module.slug.current}>
-        <LessonProvider lesson={lesson} module={module} section={section}>
+        <LessonProvider
+          lesson={isSolution ? {...solution, slug: lesson.slug} : lesson}
```
joelhooks, line 85: this is what we do on the non-embed page

the problem is a solution isn't a "real" lesson so the query and zod schema fails which makes the CASL abilities lookup fail

vojtaholik (approved):

PR #1584, opened 2024-06-26
feat(tt): Track bookmarks with database through tRPC 

zacjones93:
This is updated to have a more generic (and useful) bookmark structure that can be used for more than the total typescript book use-case.

The Bookmark models new structure looks this below. `resourceId` for the tt book use-case is the section heading (h2) on a given page (chapter). 

We query for a set of `resourceId`s based on the module we are on (book) and pass that array of resources to our bookmarks query to filter for all the users bookmarks for that current module. 

Fields lets us get specific to this use-case without hampering ourselves if we want to use bookmarks elsewhere so we can have nice display titles stored on the object.

```js
{
    userId: 'abc',
    resourceId: 'sanity-id',
    type: 'book',
    fields: {
        chapterSlug: 'kickstart-your-typescript-setup',
        chapterTitle: 'Kickstart Your TypeScript Setup',
        resourceTitle: "TypeScript's Beginnings",
        resourceSlug: 'typescript-s-beginnings',
    }
}
```

TODO: I will need to update the merge requests on planetscale for the products this affects.

gif


apps/total-typescript/src/lib/bookmarks.ts
```diff
@@ -0,0 +1,158 @@
...
+export const ReturnedBookmarkForBookSchema = z.object({
+  id: z.string(),
+  resourceId: z.string(),
+  type: z.string(),
+  fields: z.object({
+    chapterSlug: z.string(),
+    chapterTitle: z.string(),
+    resourceTitle: z.string(),
+  }),
+})
```
zacjones93, line 27: when we know we are working with a Books bookmark we can get more specific on the type being returned

apps/total-typescript/src/lib/bookmarks.ts
```diff
@@ -0,0 +1,158 @@
...
+      },
+      where: {
+        type,
+        userId,
+      },
+      orderBy: {
+        createdAt: 'desc',
+      },
+    })
+    return latestBookmarkResources
```
zacjones93, line 103: we can't know at the database level if the latest bookmark is for the current module so we return all of the users bookmarks filtered on the type and let the frontend do further filtering to grab the latest for the relevant module

apps/total-typescript/src/pages/books/[book]/[chapter]/index.tsx
```diff
@@ -162,6 +160,18 @@ const BookChapterRoute: React.FC<{
...
   const [isMenuOpen, setIsMenuOpen] = React.useState(false)
   const articleRef = React.useRef<HTMLDivElement>(null)
+  const addBookmarkMutation = trpc.bookmarks.addBookmark.useMutation({
+    onSuccess: (data) => {
+      toast.success('Bookmark added')
+    },
+    onError: (error) => {
+      if (error?.data?.httpStatus === 401) {
+        toast.error('Please log in to save bookmarks')
+      } else {
```
Comment (zacjones93) [inline apps/total-typescript/src/pages/books/[book]/[chapter]/index.tsx:170]: sending useful notification if user isn't signed in instead of silently erroring out.

apps/total-typescript/src/pages/books/[book]/index.tsx
```diff
@@ -66,8 +61,17 @@ const BookRoute: React.FC<{
...
+    trpc.bookmarks.lastBookmarkedResource.useQuery({
+      type: 'book',
+    })
+
+  const chaptersSlugs = book.chapters.map((chapter) => chapter.slug)
+  const {data: bookmarks} = trpc.bookmarks.getBookmarksForUser.useQuery()
+
+  const bookmarksForBook = bookmarks?.filter((bookmark) => {
+    return chaptersSlugs.includes(bookmark?.fields?.chapterSlug ?? '')
+  })
```
Comment (zacjones93) [inline apps/total-typescript/src/pages/books/[book]/index.tsx:74]: get all of users bookmarks and then filter on the chapter slugs that are present in the current module. 

We can't do this check at the `resourceId` level because of how data is stored in sanity, `resourceId` for these bookmarks are `h2`s in the body of a chapter resource.

apps/total-typescript/src/trpc/routers/bookmarks.ts
```diff
@@ -0,0 +1,116 @@
...
+  getBookmark: publicProcedure
+    .input(z.object({id: z.string()}))
+    .query(async ({ctx, input}) => {
+      const token = await getToken({req: ctx.req})
+      if (!token) {
+        throw new TRPCError({
+          code: 'UNAUTHORIZED',
+          message: 'User not logged in',
+        })
+      }
```
zacjones93, line 27: throwing a more useful error when not logged in

apps/total-typescript/src/trpc/routers/bookmarks.ts
```diff
@@ -0,0 +1,116 @@
...
+
+      const bookmark = await getBookmarkByResourceId({
+        userId: token.id as string,
+        resourceId: input.id,
+      })
+
+      return ReturnedBookmarkSchema.nullable().parse(bookmark)
```
zacjones93, line 34: parsing bookmarks strips `Prisma.JsonValue` types out and gives us objects that we expect on the front end. This works well but not sure how one generally deals with the `Prisma.JsonValue` type otherwise.

apps/total-typescript/src/trpc/routers/bookmarks.ts
```diff
@@ -0,0 +1,116 @@
...
+        const bookmarkForBooks = bookmarks.map((bookmark) =>
+          ReturnedBookmarkForBookSchema.parse(bookmark),
+        )
+        return bookmarkForBooks
+          ?.filter((bookmark) => {
+            return sectionSlugs.includes(bookmark?.fields?.chapterSlug ?? '')
+          })
+          .shift()
+      } else {
+        return ReturnedBookmarkSchema.parse(bookmarks.shift())
```
zacjones93, line 77: if we are working with a `book` type then we know more about the structure of the data. At the end of the day we are shifting the first item off of the array to return the latest bookmark

packages/database/prisma/schema.prisma
```diff
@@ -326,3 +327,14 @@ model PurchaseUserTransfer {
...
+
+model Bookmark {
+  id            String                    @id @default(uuid())
+  userId        String 
+  user          User                      @relation(fields: [userId], references: [id], onDelete: Cascade)
+  resourceId    String
+  type          String
+  fields        Json                      @default("{}")
+  createdAt     DateTime                  @default(now())
+}
```
zacjones93, line 340: prisma schema model now more generic

apps/total-typescript/src/lib/bookmarks.ts
joelhooks: you probably can with sql but this is OK too

apps/total-typescript/src/pages/books/[book]/index.tsx
Comment (joelhooks) [reply apps/total-typescript/src/pages/books/[book]/index.tsx:74]: imo from a data perspective we should be treating sections as "lessons" and have them stand as Content Resources and not formatted text in a document so they can be chunked/referenced/indexed at that level

packages/database/prisma/schema.prisma
joelhooks: should always have an updated and deleted at column imo

joelhooks (commented):
i'd assumed that the sections where individual resources. how massive would that be to change? silly? 🤪

zacjones93:
> i'd assumed that the sections where individual resources. how massive would that be to change? silly? 🤪

that would be a pretty large refactor on the sanity front going from the 16 chapter bodies to converting to ~318 sections attached to those chapters

apps/total-typescript/src/lib/bookmarks.ts
```diff
@@ -0,0 +1,167 @@
...
+  id: z.string(),
+  resourceId: z.string(),
+  type: z.string(),
+  fields: z.object({
+    chapterSlug: z.string(),
+    chapterTitle: z.string(),
+    resourceTitle: z.string(),
+    resourceSlug: z.string(),
+  }),
+})
```
zacjones93, line 28: we need `chapterSlug`, `chapterTitle`, `resourceTitle`, and `resourceSlug` to display and link to bookmarks

apps/total-typescript/src/lib/bookmarks.ts
```diff
@@ -0,0 +1,167 @@
...
+        userId,
+        ...(resourceIds
+          ? {
+              resourceId: {
+                in: resourceIds,
+              },
+            }
+          : {}),
+      },
+    })
```
zacjones93, line 56: when getting user bookmarks we pass `resourceIds` as an array and filter for only resources that are in that array. For prisma we have to do this spread syntax because we don't want to set a value if no `resourceIds` are set

apps/total-typescript/src/pages/books/[book]/index.tsx
```diff
@@ -66,8 +61,16 @@ const BookRoute: React.FC<{
...
+  const {data: lastBookmarkedResource} =
+    trpc.bookmarks.lastBookmarkedResource.useQuery({
+      type: 'book',
+    })
+  const resourceIds = book.chapters.flatMap((chapter) => {
+    return chapter.resources?.map((resource) => resource._id) ?? []
+  })
+  const {data: bookmarks} = trpc.bookmarks.getBookmarksForUser.useQuery({
+    resourceIds,
+  })
```
Comment (zacjones93) [inline apps/total-typescript/src/pages/books/[book]/index.tsx:73]: flatmap resources in chapters and pass them to `betBookmarksForUser` query

apps/total-typescript/src/templates/book-chapter-template.tsx
```diff
@@ -154,21 +172,24 @@ const BookChapterTemplate = ({
                     )
                   },
                   h2: (props: any) => {
+                    const resourceId = resources?.find(
+                      (r) => r.slug === props.id,
+                    )?._id
```
zacjones93, line 177: we find the Sanity `resourceId` on the h2 by checking the resources array in the current chapter and doing a find on slug

apps/total-typescript/src/templates/book-chapter-template.tsx
```diff
@@ -154,21 +172,24 @@ const BookChapterTemplate = ({
...
+                        resourceId={resourceId}
                         {...props}
                       />
                     )
                   },
                   h3: (props: any) => {
                     return (
                       <BookmarkableMarkdownHeading
-                        onAddBookmark={handleAddBookmark}
                         appendValueForRepeatedIds={`-for-${slugify(
```
zacjones93, line 193: theres no way currently (that I can think of) h3s and h4s to set the parent h2 `resourceId` for h3s and h4s so we don't pass the `onAddBookmark` prop so it won't render the bookmark button

packages/database/prisma/schema.prisma
```diff
@@ -326,3 +327,17 @@ model PurchaseUserTransfer {
...
+  id            String                    @id @default(uuid())
+  userId        String 
+  user          User                      @relation(fields: [userId], references: [id], onDelete: Cascade)
+  resourceId    String
+  type          String
+  fields        Json                      @default("{}")
+  createdAt     DateTime                  @default(now())
+  updatedAt    DateTime                   @updatedAt() @default(now())
+
+  @@index([userId, resourceId])
```
zacjones93, line 342: added an index on `userId` and `resourceId` as that will be the primary way we query this table

vojtaholik (approved):

PR #1602, opened 2024-07-03
fix(er-next): fix podcasts rendering

DrShpongle:
<details>
  <summary>Problem</summary>
  
</details>

- moved node utilities `fs` and `path` from /[podcast].tsx page file to separate function
- improved `fs` call to exclude `.DS_Store` and other hidden files/folders
- applied webpack settings in `next.config.js`

P.S. Thanks to Claude

giphy-ezgif com-webp-to-gif-converter


apps/epic-react/src/utils/get-podcasts.ts
```diff
@@ -0,0 +1,51 @@
...
+import matter from 'gray-matter'
+import {PodcastFrontMatter} from '@/@types/mdx-podcast'
+import serializeMDX from '@skillrecordings/skill-lesson/markdown/serialize-mdx'
+
+const isPodcastDirectory = (dir: string) => {
+  return (
+    fs
+      .statSync(path.join(process.cwd(), 'src/content/podcast', dir))
+      .isDirectory() && !dir.startsWith('.')
+  ) // This will exclude .DS_Store and other hidden files/folders
```
vojtaholik, line 12: `.DS_Store` is in gitignore and shouldn't be committed to git as a general rule, so I don't think this is really necessary but probably fine

vojtaholik (approved):

PR #1603, opened 2024-07-03
feat(tt): discount cannot exceed cost of product

jbranchaud:
this addresses an issue where some stacked discounts for upgrading from
multiple previous purchases were resulting in a negative display price
when really it should just display as $0. this fixes that.

### Before

CleanShot 2024-07-03 at 12 22 54@2x

### After

CleanShot 2024-07-03 at 12 54 16@2x

### Note that other more standard discounts still work as expected

CleanShot 2024-07-03 at 12 56 49@2x

disco


vojtaholik (approved):
⭐

PR #1605, opened 2024-07-03
fix(ER): Flatten module lesson list and simplify 'complete and continue' logic

zacjones93:
Previously we were trying to use `moduleProgress.nextLesson.slug` to navigate to the next lesson after completion. That data is [meant for use on the module page](https://eggheadio.slack.com/archives/C053T20CUCE/p1720014884326439?thread_ts=1719983025.515299&cid=C053T20CUCE) as it gives the first uncompleted lesson in the module back.

This simplifies the logic here by flattening the lessons into a single array and navigating to the next lesson in the array. If at any point the user hits `moduleComplete` they will be navigated to the modules `/completed` page for a certificate and continue to the next module.

gif


DrShpongle (approved):
👍👍

packages/skill-lesson/trpc/routers/progress.ts
```diff
@@ -100,17 +102,24 @@ export const progressRouter = router({
...
+        if (token && token?.sub) {
           const progress = await toggleLessonProgressForUser({
             userId: token.id as string,
             lessonId: lesson._id as string,
+            moduleId: input.moduleId,
+          })
+
+          const moduleProgress = await getModuleProgress({
+            moduleId: progress.moduleId ?? '',
+            userId: token.sub,
```
zacjones93, line 122: query for module progress when lesson is completed to also return to frontend

apps/epic-react/src/components/lesson-completion-toggle.tsx
```diff
@@ -7,10 +7,14 @@ import {trpcSkillLessons} from '@skillrecordings/skill-lesson/utils/trpc-skill-l
 import toast from 'react-hot-toast'
 
 const LessonCompleteToggle = () => {
-  const {module} = useLesson()
+  const {module, lesson} = useLesson()
+  const flattenedLessons = module.sections?.flatMap(
+    (section) => section.lessons,
+  )
+  const currentLessonIndex =
+    flattenedLessons?.findIndex((l) => l?.slug === lesson.slug) ?? 0
```
jbranchaud, line 15: There are top-level lessons that aren't within a section, so I'm thinking `module.sections.flatMap` will miss those lessons.

`module.lessons` might be all of them for a given module

zacjones93 (commented):
this fixes ER-103

apps/epic-react/src/components/lesson-completion-toggle.tsx
zacjones93: ah that's what this was for 

https://github.com/skillrecordings/products/pull/1605#discussion_r1664788713

jbranchaud: my only hesitance with this is that there might be some skill-stack internals that expects `module.sections` to only contain sections, I'm not certain tho. I just remember having to tread carefully when I was getting the Collection List display to work with ER shaped data.

PR #1610, opened 2024-07-08
article(EpicReact): Create what-is-react.mdx

kentcdodds:
I'm not sure what happens if I don't have an image or socialImage...


kentcdodds:
Looks like it fails the build if you don't include those 😅

vojtaholik (approved):

PR #1613, opened 2024-07-09
feat(er): add redeem-dialog styles

jbranchaud:
This adds the styles missing for displaying the Redeem Team Seat / Redeem Coupon dialog.

I copied these over from TJS and they aren't quite right because it is being displayed behind the rocket image. @DrShpongle could I get an assist with this?

CleanShot 2024-07-09 at 19 43 12@2x

Plus an additional fix to the way `PricingTiers` component in `skill-lesson` calls `useCoupon` so that a product is getting passed in based on the coupon. This makes it actually work (important!) and also enhances the information in the Redeem dialog.

CleanShot 2024-07-09 at 13 47 47@2x

rocket


packages/skill-lesson/path-to-purchase/product-tiers.tsx
```diff
@@ -21,8 +21,28 @@ export const PricingTiers: React.FC<
...
+          ...restrictedToProduct.image,
+          width: 132,
+          height: 112,
+        },
+      }
+    : undefined
+
+  const {redeemableCoupon, RedeemDialogForCoupon, validCoupon} = useCoupon(
+    couponFromCode,
+    productMetadata,
```
jbranchaud, line 44: We need to pass in a `product` to `useCoupon` or we invalidate any product-restricted coupons, namely team invites for a specific product tier.

packages/skill-lesson/path-to-purchase/product-tiers.tsx
```diff
@@ -21,8 +21,28 @@ export const PricingTiers: React.FC<
...
+    ? {
+        ...restrictedToProduct,
+        id: restrictedToProduct.productId,
+        image: {
+          ...restrictedToProduct.image,
+          width: 132,
+          height: 112,
+        },
+      }
+    : undefined
```
jbranchaud, line 40: This is how we look up the product metadata needed to pass in to `useCoupon` for the Redeem dialog to work. It looks up the product that matches the coupon restriction.

DrShpongle (approved):

PR #1614, opened 2024-07-09
fix(ER): add module id for lessonProgress updates

zacjones93:
gif

### DIFF
--- a/packages/database/src/prisma-api.ts
+++ b/packages/database/src/prisma-api.ts
@@ -304,6 +304,7 @@ export function getSdk(
             data: {
               completedAt: null,
               updatedAt: now,
+              moduleId,
             },
           })
         } else {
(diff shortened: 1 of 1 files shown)


packages/skill-lesson/trpc/routers/progress.ts
```diff
@@ -118,7 +118,7 @@ export const progressRouter = router({
           })
 
           const moduleProgress = await getModuleProgress({
-            moduleId: progress.moduleId ?? '',
+            moduleId: input.moduleId ?? '',
             userId: token.sub,
```
zacjones93, line 122: using the input that is passed in will be more resilient so that if for some reason progress is undefined, we can still retrieve module progress

vojtaholik (approved):

PR #1628, opened 2024-07-12
fix(er): allow teams to see redemption codes

joelhooks:
* there is no team page
* there is no display of team link on invoice page
* some purchases are old and charge isn't in KCD stripe causing crash and fallback to invoice list
* purchases were sorted `asc` so `[0]` grabbed oldest bulk purchase

We need to properly display team links and give people the opportunity to redeem their team seats as well as add more seats etc

We need to properly account for the alternative stripe, or at least fail with grace. I'd guess people don't need to see invoices from years ago so failing gracefully is fine going forward.

redeem

### DIFF
--- a/apps/epic-react/src/pages/invoices/[merchantChargeId].tsx
+++ b/apps/epic-react/src/pages/invoices/[merchantChargeId].tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import {DownloadIcon} from '@heroicons/react/outline'
+import {DownloadIcon, UserGroupIcon} from '@heroicons/react/outline'
 import {convertToSerializeForNextResponse} from '@skillrecordings/commerce-server'
 import {useLocalStorage} from 'react-use'
 import {GetServerSideProps} from 'next'
@@ -13,6 +13,9 @@ import {Transfer} from '../../purchase-transfer/purchase-transfer'
 import {MailIcon} from '@heroicons/react/solid'
 import {z} from 'zod'
 import {useRouter} from 'next/router'
+import InviteTeam from '@skillrecordings/skill-lesson/team'
+import Card from '@skillrecordings/skill-lesson/team/card'
+import {useSession} from 'next-auth/react'
 
 export const getServerSideProps: GetServerSideProps = async ({query}) => {
   const {merchantChargeId} = z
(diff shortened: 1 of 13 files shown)


apps/epic-react/src/pages/invoices/[merchantChargeId].tsx
```diff
@@ -90,6 +103,21 @@ const Invoice: React.FC<
...
+            }
+          >
+            <InviteTeam
+              session={session}
+              purchase={purchaseDetails.purchase}
+              existingPurchase={purchaseDetails.existingPurchase}
+              setPersonalPurchase={() => {}}
+            />
+          </Card>
+        )}
```
Comment (joelhooks) [inline apps/epic-react/src/pages/invoices/[merchantChargeId].tsx:120]: the current link on the home page in the WelcomeBanner links directly to an invoice so i plopped this in here

apps/epic-react/src/pages/team/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+  }
+  existingPurchase: {
+    id: string
+    product: {id: string; name: string}
+  }
+  availableUpgrades: {upgradableTo: {id: string; name: string}}[]
+  userId: string
+}
+
+const TeamPage: React.FC<React.PropsWithChildren<TeamPageProps>> = ({
```
joelhooks, line 84: the team page was missing entirely from this site

packages/database/src/prisma-api.ts
```diff
@@ -404,7 +404,7 @@ export function getSdk(
       const purchases = userId
         ? await ctx.prisma.purchase.findMany({
             orderBy: {
-              createdAt: 'asc',
+              createdAt: 'desc',
```
joelhooks, line 407: @jbranchaud this might have effects in other places but the problem was that with Epic React specifically we have some charges that aren't in this strip account, they are in another stripe account and we often have multiple team purchases because of age of the site and we grab `[0]` which fails

this sort the opposite way and lets us grab the first fresh team purchase

packages/skill-lesson/trpc/routers/invoices.ts
```diff
@@ -18,47 +18,58 @@ export const invoicesRouter = router({
...
+            }
           }
         }
+      } catch (error) {
+        console.error(error)
+        return {
+          state: 'FAILED' as const,
+          error:
+            'Unable to lookup the charge and related entities' +
+            (error as any).message,
```
joelhooks, line 71: stripe call fails when charge doesn't exist which happens because the stripe charge is in egghead stripe 😭

jbranchaud (approved):

PR #1642, opened 2024-07-19
fix: ew-144 this will allow you to redeem any product from the root if a coupon is present

joelhooks:
this is a draft that will allow people to redeem a product from the root

we lean heavily on the idea that there is a single product that people will want to redeem, so we want to make sure that we don't break that. Additionally we lean on the idea that there is a single creator.

several potential problems with this approach. 

* it's only at the top level index not product pages etc
* it's pretty naive about arrays of products
* feels brittle?


apps/epic-web/src/pages/index.tsx
```diff
@@ -63,8 +63,16 @@ const Index: NextPage<{
   const {data: commerceProps, status: commercePropsStatus} =
     trpc.pricing.propsForCommerce.useQuery({
       ...router.query,
-      productId: product.productId,
+      productId:
+        router.query.code || router.query.coupon
+          ? undefined
+          : product.productId,
```
joelhooks, line 69: `product.productId` is the hard coded default site product, but if there is a coupon or code in the url we don't want to use that

potential problem: coupon/code is general discount and not restricted to a product. what happens?

apps/epic-web/src/pages/index.tsx
```diff
@@ -63,8 +63,16 @@ const Index: NextPage<{
...
     })
+
+  const {data: commercePropsProduct} = trpc.products.getProductById.useQuery({
+    productId: commerceProps?.products[0]?.productId,
+  })
```
joelhooks, line 74: commerce props returns a product id array, so we get the first one

problem: will it return multiple products sometimes?

apps/epic-web/src/pages/index.tsx
```diff
@@ -74,8 +82,8 @@ const Index: NextPage<{
         width: 132,
         height: 112,
       },
-      title: product.title as string,
-      description: product?.description,
+      title: commercePropsProduct?.name as string,
+      description: commercePropsProduct?.description,
```
joelhooks, line 86: for the redeem dialog we want to use the restricted to product not the default product for the site

packages/skill-lesson/lib/products.ts
```diff
@@ -24,6 +24,32 @@ const productsQuery = groq`*[_type == "pricing"][0] {
...
+  },
+  features[]{
+    value
+  }
+  }
+}`
+
+export const getAllActiveProducts = async () =>
+  (await sanityClient.fetch(allProductsQuery)) || []
+
```
joelhooks, line 52: was thinking about the "active" products, but didn't get very far

packages/skill-lesson/path-to-purchase/redeem-dialog.tsx
```diff
@@ -74,11 +74,7 @@ const RedeemDialog = ({open = false, couponId, product}: RedeemDialogProps) => {
             )}
             {title ? (
               <div className="text-lg pt-5 font-medium">
-                <Balancer>
-                  Coupon for {title} by{' '}
-                  {process.env.NEXT_PUBLIC_PARTNER_FIRST_NAME}{' '}
-                  {process.env.NEXT_PUBLIC_PARTNER_LAST_NAME}
-                </Balancer>
```
joelhooks, line 81: this needs to be more intelligent per product

packages/skill-lesson/trpc/routers/pricing.ts
```diff
@@ -150,12 +155,32 @@ export const pricing = router({
...
+            },
+            {
+              id: input.code || input.coupon,
+            },
+          ],
+        },
+        select: {
+          restrictedToProductId: true,
+        },
+      })
```
joelhooks, line 172: if there's a coupon, grab it with the restricted to product id

apps/epic-web/src/pages/index.tsx
```diff
@@ -63,19 +63,27 @@ const Index: NextPage<{
...
+
+  const {data: commercePropsProduct} = trpc.products.getProductById.useQuery({
+    productId: commerceProps?.products[0]?.productId,
+  })
+
   const {redeemableCoupon, RedeemDialogForCoupon, validCoupon} = useCoupon(
     commerceProps?.couponFromCode,
     {
-      id: product.productId,
+      id: commerceProps?.products[0]?.productId || product.productId,
```
joelhooks, line 79: use the id from props

problem: only uses the first id

packages/skill-lesson/lib/products.ts
```diff
@@ -24,6 +24,32 @@ const productsQuery = groq`*[_type == "pricing"][0] {
   }
 }`
 
+const allProductsQuery = groq`*[_type == "pricing"] {
```
vojtaholik, line 27: shouldn't this be `_type == 'product'` instead of `pricing` ?

packages/skill-lesson/path-to-purchase/redeem-dialog.tsx
jbranchaud: I think we could pass these details into `useCoupon` and then that could pass them along to the RedeemDialog that gets rendered.

packages/skill-lesson/lib/products.ts
nicollguarnizo: `pricing` is the component we use to display the tiers in the home page, not necessarily all the active products. We already took care of `inactive` products, for the home page we could do a query for active `products` instead of `pricing`

packages/skill-lesson/lib/products.ts
joelhooks: nope

apps/epic-web/src/components/app/layout.tsx
```diff
@@ -68,6 +71,7 @@ const Layout: FunctionComponent<React.PropsWithChildren<LayoutProps>> = ({
         noindex={noIndex}
       />
       <GlobalSearchBar />
+      <RedeemDialogForCoupon />
```
joelhooks, line 74: no need to validate/conditional since it is done behind the scenes already

uses layout, so any page using layout "just works"

apps/epic-web/src/hooks/use-golden-ticket.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const router = useRouter()
+
+  const {data: couponData} = trpc.coupons.getForCodeOrCoupon.useQuery({
+    code: router.query.code ? (router.query.code as string) : undefined,
+    coupon: router.query.coupon ? (router.query.coupon as string) : undefined,
+  })
+
+  const {data: product} = trpc.products.getProductById.useQuery({
+    productId: couponData?.restrictedToProductId as string,
+  })
```
joelhooks, line 36: loads the coupon and then the product it's restricted to

apps/epic-web/src/hooks/use-golden-ticket.tsx
```diff
@@ -0,0 +1,66 @@
...
+      id: product?.id,
+      image: {
+        url: 'https://res.cloudinary.com/epic-web/image/upload/v1695972887/coupon_2x.png',
+        width: 132,
+        height: 112,
+      },
+      title: product?.name as string,
+      description: product?.description,
+    },
+  )
```
joelhooks, line 52: uses existing coupon validation stuff

apps/epic-web/src/hooks/use-golden-ticket.tsx
jbranchaud: I was starting to leave a comment about potentially weird behavior with a golden ticket not having a `restrictedToProductId`, but I see you handled that by returning `null` from `getProductById`. It still leaves me with the question of whether it is valid for a golden ticket to be unrestricted to a product and if the provider should have some logic for short-circuiting in that case.

apps/epic-web/src/components/app/layout.tsx
jbranchaud, line 74: So much nicer having this in the layout now instead of needing to embed it on each page where we might want to display the dialog.

apps/epic-web/src/pages/buy.tsx
```diff
@@ -53,19 +53,16 @@ const BuyPage: React.FC<
...
-    {
-      id: products[0].productId,
-      image: {
-        url: 'https://res.cloudinary.com/epic-web/image/upload/v1695972887/coupon_2x.png',
-        width: 132,
-        height: 112,
-      },
-      title: products[0].title as string,
-      description: products[0]?.description,
+  const {validCoupon} = useCoupon(couponFromCode, {
```
jbranchaud, line 56: I imagine this could be taken a step further -- a `CouponProvider` that handles both golden ticket logic and making other kinds of non-golden validated coupon details available to downstream consumers like where `couponId` gets passed in to `<Pricing ...>`

jbranchaud (approved):

PR #1645, opened 2024-07-19
Creeland/zac/fix lesson complete

zacjones93:
Because Epic React has bespoke urls (using /modules) our default `Complete and Continue` buttons lead to 404s. This creates `DefaultOverlay` and `FinsihedSectionOverlay` components to pass to `Video` to use the `Complete and Continue` button built for Epic React

## DefaultOverlay
DefaultOverlay

## DefaultOverlay completed desktop 
DefaultOverlay completed desktop

##  DefaultOverlay completed mobile
DefaultOverlay completed mobile

## FinishedSectionOverlay
FinishedSectionOverlay

gif


apps/epic-react/src/components/video-overlays/components/progress.tsx
```diff
@@ -0,0 +1,28 @@
...
+    data-progress-bar=""
+    className={cn(
+      'relative h-4 w-full overflow-hidden rounded-full bg-secondary',
+      className,
+    )}
+    {...props}
+  >
+    <ProgressPrimitive.Indicator
+      data-indicator=""
+      className="h-full w-full flex-1 bg-blue-500 transition-all"
```
zacjones93, line 23: `bg-primary` makes the progress bar invisible because so copied component and added `bg-blue-500`

apps/epic-react/src/components/video-overlays/finished-section-overlay.tsx
```diff
@@ -0,0 +1,109 @@
...
+              className="h-2 bg-blue-500"
+            />
+            <div data-lessons-completed="">
+              {completedLessonCount} / {moduleProgress?.lessonCount} lessons
+              completed
+            </div>
+          </div>
+        )}
+        <div>
+          <LessonCompleteToggle className="text-md my-4 bg-white text-white hover:bg-er-gray-300 disabled:pointer-events-none disabled:opacity-50 dark:bg-er-gray-200 dark:hover:bg-er-gray-300" />
```
zacjones93, line 69: use epic reacts `LessonCompleteToggle` this also records `moduleId` for progress which ER needs to check module progress that we aren't doing elsewhere in products

packages/skill-lesson/video/video.tsx
```diff
@@ -36,6 +37,7 @@ export const Video: React.FC<
       exerciseOverlayRenderer = () => <DefaultOverlay />,
       defaultOverlayRenderer = () => <DefaultOverlay />,
       blockedOverlayRenderer = () => <BlockedOverlay product={product} />,
+      finishedSectionOverlayRenderer = () => <FinishedSectionOverlay />,
```
zacjones93, line 40: needed to be able to override `FinishedSectionOverlay` so added as optional prop to `Video`

joelhooks (commented):
one thing we will probably want to do is label this as "legacy" or similar since new work for this site will use standard approach

joelhooks (approved):

PR #1664, opened 2024-07-29
fix(er) fix broken complete and continue button for tutorials

zacjones93:
We had hard-coded values in the `Complete and Continue` button that navigated to the next lesson when you are in a `module` lesson.

This needs to be dynamic now that we have tutorials on epic react (and in the future when we implement /workshops though I'd advocate to ditch this button in favor of how skill-stack currently handles things)

When we remove sections from url we will need to update this functionality again as `section.slug` is hardcoded in the url that is navigated to. 

https://www.loom.com/share/ec43f53ce0584854bb4e92d36483a6e9

Also fixed icon styling on the complete and continue overlay

## before
before icon style

## after
after icon style

gif

### DIFF
--- a/apps/epic-react/src/components/lesson-completion-toggle.tsx
+++ b/apps/epic-react/src/components/lesson-completion-toggle.tsx
@@ -10,7 +10,7 @@ import cx from 'classnames'
 import Spinner from '@/components/spinner'
 
 const LessonCompleteToggle = ({className}: {className?: string}) => {
-  const {module, lesson} = useLesson()
+  const {module, section, lesson} = useLesson()
   const flattenedLessons = module.sections?.flatMap(
     (section) => section.lessons,
   )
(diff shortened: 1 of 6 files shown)


apps/epic-react/src/components/lesson-completion-toggle.tsx
```diff
@@ -84,15 +84,19 @@ const LessonCompleteToggle = ({className}: {className?: string}) => {
...
+                  section?.slug
+                }/${flattenedLessons[currentLessonIndex + 1]?.slug}`,
+              )
+            } else {
+              router.push(
+                `/modules/${module.slug.current}/${
+                  flattenedLessons[currentLessonIndex + 1]?.slug
+                }`,
+              )
+            }
```
zacjones93, line 99: check if ` moduleType` is tutorial and if it is, use the tutorial url.

apps/epic-react/src/components/lesson-completion-toggle.tsx
```diff
@@ -84,15 +84,19 @@ const LessonCompleteToggle = ({className}: {className?: string}) => {
             flattenedLessons &&
             flattenedLessons[currentLessonIndex + 1]
           ) {
-            reward()
```
zacjones93, line 87: getting rid of these reward animations as we are triggering this optimistically and this double renders the animation

joelhooks (approved):
i think there is a good argument for having "legacy" components for v1 so the alternative url scheme isn't polluting the project forever

PR #1653, opened 2024-07-23
chore: Update workshop welcome email to use post-purchase template

nicollguarnizo:
Screenshot 2024-07-31 at 7 17 47 PM

emails

PR #1671, opened 2024-08-01
fix(ew): fixing the type for product

nicollguarnizo:
@joelhooks You are right about the array type for `product`, this fixes that - https://github.com/skillrecordings/products/pull/1653#discussion_r1700723844

buttons


joelhooks (approved):
nice!

PR #1672, opened 2024-08-02
fix(er): claim seat for a team 

nicollguarnizo:
team

Screenshot 2024-08-02 at 1 59 46 PM


apps/epic-react/src/templates/product-template.tsx
```diff
@@ -20,17 +20,45 @@ import {ProductPageProps} from '@/pages/products/[slug]'
...
+      }
+    : undefined
+
+  const {redeemableCoupon, RedeemDialogForCoupon, validCoupon} = useCoupon(
+    couponFromCode,
+    productMetadata,
+  )
+
+  const couponId =
+    couponIdFromCoupon || (validCoupon ? couponFromCode?.id : undefined)
```
joelhooks, line 60: should we consider implementing golden ticket provider in this app instead of "whack-a-mole" approach? https://github.com/skillrecordings/products/blob/ddcb80827ce97a2bf2c85325ab13bc949476d462/apps/epic-web/src/hooks/use-golden-ticket.tsx

joelhooks (approved):
approving but think we should use GoldenTicketProvider in this (and all) apps https://github.com/skillrecordings/products/blob/ddcb80827ce97a2bf2c85325ab13bc949476d462/apps/epic-web/src/hooks/use-golden-ticket.tsx

apps/epic-react/src/templates/product-template.tsx
nicollguarnizo: totally agree with you! will create a linear issue for this

PR #1677, opened 2024-08-07
feat(er): add new moduleType === 'workshop-v2' 

nicollguarnizo:
new


joelhooks:
went another route with this

PR #1691, opened 2024-08-21
feat: make skill lesson hooks configurable for ignoring [section] in url

vojtaholik:
this fixes navigating to next lesson from overlays and collection component. I added new `ignoreSections` prop for `Collection` component making it work with url structure without `[section]`. same prop is used for `handleContinue` util.

gif

### DIFF
--- a/apps/epic-react/src/pages/workshops/[module]/[lesson]/index.tsx
+++ b/apps/epic-react/src/pages/workshops/[module]/[lesson]/index.tsx
@@ -22,7 +22,11 @@ export const getStaticProps: GetStaticProps = async (context) => {
   const sectionSlug = params?.section as string
 
   const module = await getWorkshop(params?.module as string)
-  const section = await getSection(sectionSlug)
+  // if sectionSlug does not exist in url but is still present in data structure, we need to get current lesson by filtering through all sections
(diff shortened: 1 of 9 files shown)


apps/epic-react/src/pages/workshops/[module]/[lesson]/index.tsx
```diff
@@ -22,7 +22,11 @@ export const getStaticProps: GetStaticProps = async (context) => {
   const sectionSlug = params?.section as string
 
   const module = await getWorkshop(params?.module as string)
-  const section = await getSection(sectionSlug)
+  // if sectionSlug does not exist in url but is still present in data structure, we need to get current lesson by filtering through all sections
+  const currentLessonSection = module.sections.find((section) => {
+    return section.lessons.find((lesson) => lesson.slug === lessonSlug)
+  })
+  const section = await getSection(sectionSlug || currentLessonSection?.slug)
```
Comment (vojtaholik) [inline apps/epic-react/src/pages/workshops/[module]/[lesson]/index.tsx:29]: one of the tricks to make it all work was to still load the current section because it helps with existing functionality to find next resource, etc. and then have `ignoreSections` prop control how paths are build in `Collection` and `handleContinue` util.

PR #1711, opened 2024-09-06
feat(commerce): update bulk discount logic

nicollguarnizo:
sale

### DIFF
--- a/packages/commerce-server/src/format-prices-for-product.test.ts
+++ b/packages/commerce-server/src/format-prices-for-product.test.ts
@@ -65,7 +65,7 @@ for (const quantity of [7, 13, 26, 42]) {
     ])
 
     mockCtx.prisma.merchantCoupon.findFirst.mockResolvedValue(
-      MOCK_SITE_SALE_COUPON,
+      MOCK_LARGE_SITE_SALE_COUPON,
     )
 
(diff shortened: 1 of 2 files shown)


packages/commerce-server/src/format-prices-for-product.test.ts
```diff
@@ -65,7 +65,7 @@ for (const quantity of [7, 13, 26, 42]) {
     ])
 
     mockCtx.prisma.merchantCoupon.findFirst.mockResolvedValue(
-      MOCK_SITE_SALE_COUPON,
```
nicollguarnizo, line 68: this discount is lower than the quantities that are being tested. That's why we need it a higher discount.

packages/commerce-server/src/format-prices-for-product.test.ts
```diff
@@ -145,23 +145,23 @@ test('does not apply restricted coupon to other product', async () => {
 
 test('product with quantity 5 calculatedPrice to have discount applied', async () => {
   mockCtx.prisma.merchantCoupon.findMany.mockResolvedValue([
-    getMockCoupon('bulk', 0.05),
+    getMockCoupon('bulk', 0.15),
```
nicollguarnizo, line 148: new discounts

joelhooks (approved):

PR #1714, opened 2024-09-09
feat(kcd): resources in workshops/modules api response

vojtaholik:
gif


packages/skill-lesson/lib/modules.ts
```diff
@@ -254,6 +254,47 @@ export const getModuleWithResources = async (slug: string) =>
             }
           }
         },
+         "sections": resources[@->._type == 'section']->{
```
vojtaholik, line 257: leaving this in so it won't break the workshop app. Kent mentioned that because of the auto-update feature of the workshop app, we should be able to remove the old fields as well in a few months to reduce dead weight.

PR #1718, opened 2024-09-12
feat(skill-lesson): make first lesson of first section freely visible

vojtaholik:
gif


joelhooks (changes requested):
i think this should absolutely 100% be at data level and not a zero index operation

PR #1740, opened 2024-10-04
fix: re-enable purchase transfer

joelhooks:


packages/skill-lesson/trpc/routers/invoices.ts
```diff
@@ -20,6 +21,12 @@ export const invoicesRouter = router({
 
       try {
         if (merchantCharge && merchantCharge.identifier) {
+          const {stripe} =
+            merchantCharge.merchantAccount.label === 'egghead-stripe' &&
+            eggheadReadOnlyContext
+              ? eggheadReadOnlyContext
+              : defaultStripeContext
```
joelhooks, line 28: logntime issue that prevented legacy customers from accessing old invoices that I fixed while in here

packages/skill-api/src/core/services/process-stripe-webhook.ts
```diff
@@ -406,12 +406,17 @@ export const processStripeWebhook = async (
         ? targetEmail.toLowerCase() !== previousEmail.toLowerCase()
         : false
 
+      const {user: fromUser} = previousEmail
+        ? await findOrCreateUser(previousEmail)
+        : {user: null}
       const {user: updateUser} = await findOrCreateUser(targetEmail, name)
 
-      await transferPurchasesToNewUser({
-        merchantCustomerId: merchantCustomer.id,
```
joelhooks, line 412: this was the root cause because the incoming merchantCustomer row was shared across 20,557 charges and purchases for reasons

now there is no shared merchant customer and this is also more specific

packages/skill-api/src/core/services/transfer-purchase.ts
```diff
@@ -163,92 +166,100 @@ export async function transferPurchase({
     // creates a transfer record
     // update merchant records if the purchase involved stripe
 
-    if (purchase?.merchantCharge?.merchantCustomer) {
+    const updates = []
```
joelhooks, line 169: everything goes in a transaction this way

packages/skill-api/src/core/services/transfer-purchase.ts
```diff
@@ -163,92 +166,100 @@ export async function transferPurchase({
...
+
+    if (purchase?.merchantCharge) {
```
joelhooks, line 171: updates separately from the customer below

packages/skill-api/src/core/services/transfer-purchase.ts
```diff
@@ -163,92 +166,100 @@ export async function transferPurchase({
...
-      //   },
-      //   data: {
-      //     userId: targetUser.id,
-      //   },
-      // })
+      const {identifier} = purchase.merchantCharge.merchantCustomer
+      const isCustomer =
+        purchase.merchantCharge.merchantAccount.label !== 'egghead-stripe' &&
+        identifier &&
+        !identifier.startsWith('no_stripe_customer_id')
```
joelhooks, line 198: better if these rows didn't exist at all, but they do so...

packages/skill-api/src/core/services/transfer-purchase.ts
```diff
@@ -163,92 +166,100 @@ export async function transferPurchase({
...
+          )) as Stripe.Response<Stripe.Customer>
+
+          await stripe.customers.update(identifier, {
+            email: targetUser.email,
+            name: targetUser.name || existingCustomer.name || targetUser.email,
+          })
+        } catch (error) {
+          console.error('Error updating stripe customer', error)
+        }
+      }
```
joelhooks, line 213: this was the culprit since the inbound transfer had newer products but was also associated freshly with that shared MerchantCustomer row. This triggered the stripe webhook and BLAMMMMMO 💥

packages/database/src/prisma-api.ts
```diff
@@ -871,21 +877,21 @@ export function getSdk(
       }
     },
     async transferPurchasesToNewUser({
-      merchantCustomerId,
+      fromUserId,
```
joelhooks, line 880: the rows were removed but this is more constrained regardless

PR #1741, opened 2024-10-07
feat(ew): add 2024 and 2025 conf pages with redirects

Creeland:
direct person


zacjones93 (approved):

apps/epic-web/next.config.js
```diff
@@ -60,6 +60,21 @@ const nextConfig = {
         destination: '/contributors/:slug*',
         permanent: true,
       },
+      {
+        source: '/conf',
+        destination: '/conf/2025',
+        permanent: true,
```
joelhooks, line 66:
```suggestion
        permanent: false,
```

I don't think it is since we will change it again later probably

joelhooks (approved):

apps/epic-web/next.config.js
Creeland: good point

PR #1758, opened 2024-10-15
nicollg/update calculations for slack daily and monthly report 

nicollguarnizo:
donut


apps/epic-web/src/components/calculations/calculate-splits.ts
```diff
@@ -0,0 +1,167 @@
+type ProductGroup = {
+  productName: string
+  productId: string
+  count: number
+  amount: number
+  net: number
+  fee: number
+  refunded: number
+}
```
joelhooks, line 9: we should probably only declare these types once if possible

joelhooks (approved):

PR #1763, opened 2024-10-23
feat(tt): survey page for year end survey

joelhooks:
adds a survey page at /survey

chicken


mattpocock:
Ordering is off here

image

vojtaholik:
> Ordering is off here
> 
> image
> 
> In fact, most of the 'ordered' ones are in the wrong order, like beginner -> senior

got this fixed 👍

vojtaholik (approved):
saving individual responses as they are entered, instead of waiting until the last one is submitted might be helpful in case users drop off. NUQS for search params state management would also be a nice improvement.

PR #1782, opened 2024-12-02
feat(ewd): megabundle global banner

vojtaholik:
🚨 they show up with product state set to `active`

screenshot-2

gif

### DIFF
--- a/apps/epic-web/src/components/epic-principles.mdx
+++ b/apps/epic-web/src/components/epic-principles.mdx
@@ -1,0 +1,461 @@
+# Epic Programming Principles v1
+
+Principles guide the way you work together and make decisions. They are bigger than languages, frameworks, libraries, and tools. They are the foundation of your craft. As the field evolves, the principles guide you in making decisions in the new landscape. While principles do evolve as well, they are more stable than a technology stack.
+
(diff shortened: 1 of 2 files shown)


apps/epic-web/src/components/cta/resource-cta.tsx
```diff
@@ -294,9 +294,16 @@ const ActivePromotionCTA = ({
     return null
   }
 
+  // TODO: This is a temporary fix to check if the product is a megabundle,
+  // it would be better to have a flag in the product itself!
+  const IS_MEGABUNDLE =
+    currentActivePromotion.product.slug.includes('megabundle')
+
```
vojtaholik, line 301: we don't handle multiple instructors, etc. yet

image

apps/epic-web/src/trpc/routers/cta.ts
```diff
@@ -312,6 +314,58 @@ export const ctaRouter = router({
             CURRENT_ACTIVE_PROMOTION = activePromotion.data
           }
         }
+      } else {
+        const nonDefaultCoupons = await prisma.coupon.findMany({
```
vojtaholik, line 318: lookup for non default coupons. default ones still take precedence.

PR #1807, opened 2025-04-23
feat: add sanity-powered login wall for articles

zacjones93:
Matt mentioned putting up a 'email wall' for the cursor rules article for the sale that we will be running. 

This adds a `withEmailWall` property to articles in sanity that we can flip on if we want to selectively show this CTA to people. Makes sense to me to turn this login wall off after the sale is over.

Cursor Rules for Better AI Development  Total TypeScript

g stop

### DIFF
--- a/apps/total-typescript/schemas/documents/article.tsx
+++ b/apps/total-typescript/schemas/documents/article.tsx
@@ -88,6 +88,12 @@ export default {
       type: 'text',
       validation: (Rule) => Rule.max(160),
     },
+    {
+      name: 'showLoginWall',
+      title: 'Show Login Wall',
+      type: 'boolean',
+      initialValue: false,
+    },
   ],
   preview: {
     select: {
--- a/apps/total-typescript/schemas/documents/article.tsx
+++ b/apps/total-typescript/schemas/documents/article.tsx
@@ -89,8 +89,8 @@ export default {
       validation: (Rule) => Rule.max(160),
     },
     {
-      name: 'showLoginWall',
-      title: 'Show Login Wall',
+      name: 'withEmailWall',
+      title: 'Show Email Wall',
       type: 'boolean',
       initialValue: false,
     },
--- a/apps/total-typescript/src/pages/[article].tsx
+++ b/apps/total-typescript/src/pages/[article].tsx
@@ -41,15 +41,20 @@ export const getStaticProps: GetStaticProps = async ({params}) => {
       },
     }))
 
-  const shortenedArticleBody =
-    article.body &&
-    (await serializeMDX(take(article.body?.split('\n'), 6).join('\n'), {
-      useShikiTwoslash: true,
-      syntaxHighlighterOptions: {
-        authorization: process.env.SHIKI_AUTH_TOKEN,
-        endpoint: process.env.SHIKI_ENDPOINT,
-      },
-    }))
+  let shortenedArticleBody
+  if (article.withEmailWall) {
+    shortenedArticleBody =
+      article.body &&
+      (await serializeMDX(take(article.body?.split('\n'), 6).join('\n'), {
+        useShikiTwoslash: true,
+        syntaxHighlighterOptions: {
+          authorization: process.env.SHIKI_AUTH_TOKEN,
+          endpoint: process.env.SHIKI_ENDPOINT,
+        },
+      }))
+  } else {
+    shortenedArticleBody = null
+  }
 
   // Fetch first 6 articles only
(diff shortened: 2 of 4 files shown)


apps/total-typescript/src/pages/[article].tsx
```diff
@@ -40,6 +41,16 @@ export const getStaticProps: GetStaticProps = async ({params}) => {
...
 
+  const shortenedArticleBody =
+    articleBody &&
+    (await serializeMDX(take(article.body?.split('\n'), 6).join('\n'), {
+      useShikiTwoslash: true,
+      syntaxHighlighterOptions: {
+        authorization: process.env.SHIKI_AUTH_TOKEN,
+        endpoint: process.env.SHIKI_ENDPOINT,
+      },
+    }))
```
Comment (zacjones93) [inline apps/total-typescript/src/pages/[article].tsx:52]: because we serializeMDX on the server, it was going to be over complicated to try and shorten the article length on the client side so this adds a new `shortenedArticleBody` prop that will display if `showLoginWall` is true and user is not logged in

apps/total-typescript/src/templates/article-template.tsx
```diff
@@ -193,6 +200,25 @@ const ArticleTemplate: React.FC<ArticleTemplateProps> = ({
...
+                    Log in to read the full article
+                  </h3>
+                  <p className="mb-4 text-sm text-gray-400">
+                    Unlock this article and more by logging into your account.
+                  </p>
+                  <Link
+                    href={`/login?callbackUrl=${encodeURIComponent(
+                      router.asPath,
+                    )}`}
+                  >
```
zacjones93, line 216: redirect back to the article when logged in

apps/total-typescript/schemas/documents/article.tsx
```diff
@@ -88,6 +88,12 @@ export default {
       type: 'text',
       validation: (Rule) => Rule.max(160),
     },
+    {
+      name: 'showLoginWall',
```
vojtaholik, line 92: suggestion: `withEmailWall`

apps/total-typescript/src/pages/[article].tsx
```diff
@@ -40,6 +41,16 @@ export const getStaticProps: GetStaticProps = async ({params}) => {
       },
     }))
 
+  const shortenedArticleBody =
```
Comment (vojtaholik) [inline apps/total-typescript/src/pages/[article].tsx:44]:
```tsx
let articleBody
if (article.withEmailWall) {
  articleBody = shortenedArticle 
} else {
```

Comment (vojtaholik) [reply apps/total-typescript/src/pages/[article].tsx:44]: that should make it stop throwing, but if you want to make sure that it's never gonna break, we gotta remove all custom components from markdown in order to be able to slice it. here's example rehypePlugin that removes all code blocks: https://github.com/skillrecordings/products/blob/main/apps/epic-react/src/utils/mdx.tsx#L34

apps/total-typescript/src/templates/article-template.tsx
```diff
@@ -63,6 +66,10 @@ const ArticleTemplate: React.FC<ArticleTemplateProps> = ({
     threshold: 0.5,
   })
 
+  const isLoggedIn = session?.user
+  const articleBodyDisplay =
+    article.showLoginWall && !isLoggedIn ? shortenedArticleBody : articleBody
```
vojtaholik, line 71: this needs to happen on server side

apps/total-typescript/src/templates/article-template.tsx
```diff
@@ -193,6 +200,25 @@ const ArticleTemplate: React.FC<ArticleTemplateProps> = ({
                 }}
               />
             )}
+            {!isLoggedIn && (
+              <div className="absolute bottom-0 left-0 z-20 flex h-64 w-full flex-col items-center justify-end bg-gradient-to-t from-background via-background to-transparent p-5 text-center">
+                <div className="rounded-lg border bg-background p-5 shadow-lg">
+                  <h3 className="mb-2 text-lg font-semibold text-white">
```
vojtaholik, line 206: too many `p-5`s

apps/total-typescript/src/templates/article-template.tsx
```diff
@@ -193,6 +200,25 @@ const ArticleTemplate: React.FC<ArticleTemplateProps> = ({
...
+                    Log in to read the full article
+                  </h3>
+                  <p className="mb-4 text-sm text-gray-400">
+                    Unlock this article and more by logging into your account.
```
vojtaholik, line 210: it's an email wall, same as on tutorials and subscribing should unlock the article instantly (instead of redirecting).

PR #1824, opened 2025-09-18
fix(testing):  for coupon management in testing page

nicollguarnizo:
Screenshot 2025-09-18 at 3 40 18 PM


apps/epic-web/src/pages/testing/index.tsx
```diff
@@ -75,23 +75,25 @@ const Index: NextPage<{
   const ALLOW_PURCHASE =
     router.query.allowPurchase === 'true' || product.state === 'active'
   const {subscriber, loadingSubscriber} = useConvertkit()
+  const pricingProduct = products?.[0] // The actual product being sold
```
joelhooks, line 78: using array index access on products has bitten us many many times so it's something to be careful about and it will probably bite you later on 😅

joelhooks (approved):

